PR #88, opened 2016-10-07
GN build support

chrme:
some fixes for GN support


Eloston:
Oh neat! Thanks for doing this. Have you tested them to see if it works?

Also, what are your future plans? Are you planning to update the rest of the patches that use GYP? It'll be nice to know so we don't have multiple people working on this.

BTW, you'll need to modify `patch_order` if you add patches.

chrme:
> Have you tested them to see if it works?

Yes, it works.
> what are your future plans?

I'm only compile the original chromium with a custom config (i have fully automatic script that also checking for a new version) because i don't have much free time.
> Are you planning to update the rest of the patches that use GYP?

But the GYP don't have compilation problems.

Eloston:
> I'm only compile the original chromium with a custom config

So you've converted all of the GYP flags from `resources/common/gyp_flags` to GN args I'm guessing?

> But the GYP don't have compilation problems.

Hmm yeah you're making a Windows 64-bit build so you don't need `wow_helper`, and I'm guessing you're using Google's build of `yasm` since you're not running both source cleaner and domain substitution. You're also not building Debian or Mac OS packages, so you didn't need to port over those. I'm kind of surprised that you didn't have to fix `fix-building-without-safebrowsing.patch` though.

Anyways, don't feel obligated to fix more patches for GN. I can do them over time if no one else has by then. You've already saved me some time by doing this, so thanks again.

chrme:
> So you've converted all of the GYP flags from resources/common/gyp_flags to GN args I'm guessing?

some args are gone in GN, but i'm use most of it

> you're using Google's build of yasm

unfortunately yes
> since you're not running both source cleaner and domain substitution.

i'm running both, but on 32-bit windows the wow_helper is cleaned and i compile it by hand (can't figure it out how to send _'msvs_target_platform': 'x64'_ to BUILD.gn because it use x86 and show the error)

Eloston:
Okay, thanks for the input and the effort you've put in.

Issue #78, opened 2016-10-03
Hardcoded GCC version in macOS 


scottydelta:
The build fails due to hard coded GCC versions in the code. eg: 
`brew install gcc49` installs binary `x86_64-apple-darwin16.0.0-c++-4.9` on MacOS Sierra but the code has `x86_64-apple-darwin15.4.0-c++-4.9` and is hard coded in [resources/macos/patches/ungoogled-macos/fix-libcxx-archive-build-script.patch](https://github.com/Eloston/ungoogled-chromium/blob/master/resources/macos/patches/ungoogled-macos/fix-libcxx-archive-build-script.patch#L10) and [buildlib/macos.py](https://github.com/Eloston/ungoogled-chromium/blob/master/buildlib/macos.py#L72)

The versions can be put into a config file and can be parsed using ConfigParser but not sure how to include that version in the patch file(resources/macos/patches/ungoogled-macos/fix-libcxx-archive-build-script.patch)

Eloston:
Yeah someone reported that earlier and they closed it themselves for reasons I'm not sure...

Instead of hardcoding the path in a config file, I want to implemenet a function in `MacOSBuilder` that can auto-detect the path to Homebrew's GCC. However, I don't know enough about Homebrew to figure out the best way to detect the GCC path.

I am planning on using #69 to solve the problem of a version-dependent path to the compiler.

Eloston:
Adding `help wanted` label so someone can help out with Homebrew.

dangom:
I'm no expert, but this does the trick under Python 3.5 on MacOS Sierra:

```
@static_method
def get_homebrew_gcc_bin():
    brew_list_gcc = subprocess.Popen(['brew', 'list', 'gcc'],
                                     stdout=subprocess.PIPE)
    grep_gcc_path = subprocess.Popen(['grep', 'bin/x86_64'],
                                stdin=brew_list_gcc.stdout,
                                stdout=subprocess.PIPE,
                                stderr=subprocess.PIPE)
    out, err = grep_gcc_path.communicate()
    result = out.split(b'\n')[0]
    if not result:
        raise OSError("Is gcc properly installed? Please check the output of \"brew list gcc\"")
    return result
```

```
In [34]: get_homebrew_gcc_bin()
Out[34]: 
b'/usr/local/Cellar/gcc/6.2.0/bin/x86_64-apple-darwin15.6.0-c++-6'
```

Eloston:
[...]

PR #101, opened 2016-10-11
proper detection and usage of c++-4.9 on MacOS fixing #78

takinbo:


Eloston:
Hi, sorry it took me a little while to respond to your pull request as opposed to other issues here. I have a busy schedule with some variations, so I can only respond to issues on and off.

Anyway, does `c++-4.9` exist after `brew link gcc` is invoked, and does it always refer to the GCC C++ compiler? I was thinking that `g++-4.9` would be better to use since it generally implies GCC's version.

Eloston:
@dangom @9Morello Sorry to bug you two, but do you know the answer to the question above? Here it is again for convenience:
> Anyway, does `c++-4.9` exist after `brew link gcc` is invoked, and does it always refer to the GCC C++ compiler? I was thinking that `g++-4.9` would be better to use since it generally implies GCC's version.

9Morello:
`c++-4.9` points by default to `/usr/local/bin/c++-4.9`.

Eloston:
So what is `/usr/local/bin/c++-4.9` then? The GCC compiler? I'm worried that `c++` is ambiguous and may instead refer to clang if the versions happen to align. Does `g++-4.9` exist?

9Morello:
Sorry for not responding clearly before. Yes, it points to the GCC C++ compiler. (`/usr/local/Cellar/gcc49/4.9.3/bin/c++-4.9`). `g++-4.9` exists and its a shortcut to `/usr/local/Cellar/gcc49/4.9.3/bin/g++-4.9`.

Eloston:
Cool, that's what I needed to know. Thanks

PR #108, opened 2016-10-19
Make build.py executable

starfys:
Running ```./build.py``` in bash or zsh in GNU/Linux will result in ```bash: ./build.py: Permission denied``` or ```zsh: permission denied: ./build.py``` respectively. Adding the executable flag to build.py fixes this.


Eloston:
Oh, ok.

PR #127, opened 2016-11-27
Fix HTTPS by disabling CT enforcing.

VittGam:
Hey,

You can use this patch to fix the HTTPS mess until you find time to switch to Chromium 54.

Cheers,
Vittorio


Eloston:
I didn't push some changes I made because I didn't want people to think that 54 was ready yet. I think I'll just keep pushing to let people know things are being done.

Regarding your PR: it's a good patch, but it's a bit late now. Thanks anyway

PR #144, opened 2016-12-11
Fix building with Widevine support on MacOS

9Morello:
Fixes building with the `enable_widevine=true` flag on OS X. Widevine still doesn't work out of the box, but the user can copy the plugin files from Google Chrome and get it to work now.

### DIFF
--- a/resources/macos/gn_flags
+++ b/resources/macos/gn_flags
@@ -1,3 +1,3 @@
 is_clang=true
 clang_use_chrome_plugins=false
-enable_widevine=false
+enable_widevine=true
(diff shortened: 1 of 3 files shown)


pjv:
nice. @9Morello will you be updating your binary that is linked from the releases page here?

9Morello:
@pjv Yes, but getting the 16.04 packages is priority now. 
You will need to replace the `Libraries` folder inside `Chromium.app/Contents/Version/55.xxxx` with the same folder from Google Chrome's package.

Issue #137, opened 2016-12-08
Problem building on Windows: GN bootstrapping problem


Eloston:
[I'm having troubles trying to boostrap GN via the `bootstrap.py` script.](https://gist.github.com/Eloston/2647cff8f48322ace02e6ae93c255568)

Here's what I did to get to this point:
* Pull the current state of the master branch
* Setup the build dependencies according to BUILDING.md
  * Used `builder.path_overrides` in `build.py` for `python`, `bison`, `gperf`, `patch`, and `ninja`
* In `bootstrap.py` within the `check_call()` definition, add `kwargs["shell"] = True` right before the `subprocess.check_call` function call.
* EDIT: Open a VS 2015 Developer Console and invoke `build.py` with it (this sets environment variables for the bootstrap script)

Then the build fails with that error linked above. Does anyone have any ideas?

nopjmp:
It looks like it's a bug upstream that they fixed recently. I'm able to replicate it on my system. I'll see if I can figure something out.

nopjmp:
I got it fixed. The bootstrap.py file needed to be updated. I'm fixing up the other parts of the Windows build now as well and will submit a pull request with my changes once I'm done.

waqas:
Any progress on Windows build? I am checking this every few hours since last couple of weeks :)

nopjmp:
Not much progress except it's at least building to linking. There was a wierd patch problem because git converted the LFs to CRLFs on the line endings. I got a linker error yesterday probably because I kept stopping the build.

It takes several hours on my machine to compile and can be quite annoying to deal with. I'd check back next week or this weekend instead of what you are doing.

Eloston:
@waqas In the future, you can just click the "subscribe" button on the side to listen to notifications.

@nopjmp GNU patch from MSYS2 (which I tell users to use) seems to be fine with different line endings. What are you using to apply patches?

nopjmp:
@Eloston I forgot about the subscribe button. I'm using GNU patch from MSYS2 the problem is the patches were never applied due to how Python was working. Switching to -i and providing the full path resolved the issue. I'm going to try to build again tonight and see if I can fix my problems.

Sent from my android device.

-----Original Message-----

Eloston:
@nopjmp The subscribe button message wasn't for you. Anyways, it's puzzling that you needed to invoke patch that way, because it worked fine before by passing the patch data in via stdin (and it still did when I attempted to build and got that error)

[...]

Issue #118, opened 2016-11-14
Unable to compile 64bit swapimport on Windows


DeadSix27:
I've used git master:

```
[main]
chromium_version = 53.0.2785.143
release_revision = 1

```
I've added: 
`builder.target_arch = CPUArch.x64`
at
```
def main():
	try:
		builder = buildlib.get_builder()
                builder.target_arch = CPUArch.x64
```
in build.py

And changed:
`build_output = pathlib.Path("out", "Release")`
to
`build_output = pathlib.Path("out", "Release_x64")`
in:
common.py
(without the addition of _x64, it complains that I'm using the wrong arch)

However, I get this error (see bottom)

Any idea how to solve this?

```
[7395/20450] LINK_EMBED ..\..\third_party\syzygy\binaries\exe\swapimport.exe
FAILED: ../../third_party/syzygy/binaries/exe/swapimport.exe
C:\_chromium\depot_tools\python276_bin\python.exe gyp-win-tool link-with-manifests environment.x64 True ..\..\third_party\syzygy\binaries\exe\swapimport.exe "C:\_chromium\depot_tools\python276_bin\python.exe gyp-win-tool link-wrapper environment.x64 False link.exe /nologo /OUT:..\..\third_party\syzygy\binaries\exe\swapimport.exe @..\..\third_party\syzygy\binaries\exe\swapimport.exe.rsp" 1 mt.exe rc.exe "obj\third_party\syzygy\syzygy\swapimport\swapimport.swapimport.exe.intermediate.manifest" obj\third_party\syzygy\syzygy\swapimport\swapimport.swapimport.exe.generated.manifest ..\..\build\win\compatibility.manifest
LibDef: Total time = 0.000s
pe_lib.lib(pe_lib.pe_utils.obj) : error LNK2001: unresolved external symbol "unsigned __int64 const pe::kDosStubSize" (?kDosStubSize@pe@@3_KB)
pe_lib.lib(pe_lib.pe_utils.obj) : error LNK2001: unresolved external symbol "unsigned char const * const pe::kDosStub" (?kDosStub@pe@@3QBEB)
..\..\third_party\syzygy\binaries\exe\swapimport.exe : fatal error LNK1120: 2 unresolved externals
Final: Total time = 1.593s
Traceback (most recent call last):
  File "gyp-win-tool", line 323, in <module>
    sys.exit(main(sys.argv[1:]))
  File "gyp-win-tool", line 29, in main
[...]

PR #147, opened 2016-12-15
Fix building on Windows.

nopjmp:
This fixes building on Windows. We disable reordering and fix iridium to link correctly on Chromium 55. We still download syzygy for now until we decide if we want to keep it. icu_use_data_file has to be enabled on Windows for now to get it to at least build. Patches are not read into memory and piped over stdin anymore.

We can just kill syzygy building/downloading all together if you like, but I think it may be worth looking into at some point if performance or linking problems come up.

Should fix both #137 #118


nopjmp:
I do apologize for the terribly formatted git message. I was in a rush to get this pull request created before I went to bed to avoid another 24 hour wait.

I'll check 64bit build tomorrow.

Eloston:
The commit message is fine. Good implementation of the changes. We should probably disable downloading syzygy for now by commenting out the lines in `extra_deps.ini` so there won't be extra downloading.

Also what is the problem with the original code for patching? What kind of errors (or any output) do you get? The GNU patch utility I instructed users to use in BUILDING works on Windows 10.

nopjmp:
The patch function outputs patching but it doesn't patch anything. I can look further into it.

I'll disable the download when I test 64bit
Sent from my android device.

Eloston:
> The patch function outputs patching but it doesn't patch anything. I can look further into it.

That's really weird. I have no idea what can be causing that if the patch utility hasn't changed and buildlib's invocation of it hasn't changed. Maybe it's some weird Windows silent failure?

nopjmp:
I think it might be due to how python is interacting with the shell, but stdin might have been acting strangely in my setup. I'm not using ActivePython, but the official release. It might work if I updated python 3 I just noticed I'm on a REALLY old version.

```
C:\Users\nopjmp>C:\Python34\python.exe
Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:44:40) [MSC v.1600 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>>
```

Eloston:
@nopjmp Ah yeah Python 3.5 changed some stuff with the subprocess module. You should upgrade since I haven't tested the behavior with Python 3.4 on Windows (someone has on Debian Jessie though, but it's different from Windows).

nopjmp:
@Eloston 64bit builds, but I need to fix GN bootstrap.py to compile for 64bit. Otherwise, it works just fine. I'll try to clean it up today if I have time; otherwise, it will be done tomorrow. I'll test Python3.5 also.

Eloston:
@nopjmp Alright sounds good. Glad to hear that we're almost there.

nopjmp:
Windows 64bit is fixed and I reverted the hacked up patch function.

Eloston:
Perfect. Even all of the patches and code are consistent in style with everything else. Thanks for the great work!

PR #173, opened 2017-01-22
fix linux build

wimh:
I'm not a python expert, but I got an error when trying to build on linux. This fixed it.

```
  File "./build.py", line 58, in <module>
    exit(main())
  File "./build.py", line 44, in main
    builder.generate_package()
  File "/ungoogled-chromium/buildlib/linux.py", line 75, in generate_package
    for arcname, real_path in file_list_generator():
  File "/ungoogled-chromium/buildlib/linux.py", line 70, in file_list_generator
    for target_rel_path, input_rel_path in self._extra_packaging_files:
ValueError: too many values to unpack (expected 2)
```


Eloston:
Whoops. Thanks for catching that.

Issue #176, opened 2017-01-26
Update to version 57


Eloston:
Here's what needs to be done (in approximate order):
* [x] ~Finish #125 (@Eloston is working on this)~
* [x] ~Include new patches from Debian (@Eloston is working on this)~
* [x] ~Include new patches from Inox (@Eloston is working on this)~
* [x] ~Include (or upgrade) patches from Iridium (@Eloston is working on this)~
* [x] ~Update ungoogled-chromium's own patches (@Eloston is working on this)~
* [x] ~Get it to build (on Debian Stretch first) (@Eloston is working on this)~
* [x] ~Refactor the build system again. Implement domain substitution in Debian build files generator (@Eloston is working on this)~
* [ ] Get builds on other platforms to work
  * [x] ~macOS support (Binaries will be published soon)~
  * [ ] Windows support (See #215)
* [ ] Work on other issues in milestone (Pending)
* [x] ~~Merge from `develop` to `master` (Pending)~~
* [x] ~~Finish upgrading downloads system (Pending)~~
* [ ] Publish binaries (Pending)

**NOTE**: From now on, development will occur on `develop` and then be merged into `master` once it is ready. `master` should be considered "somewhat stable", i.e. features aren't completely broken.

The other issues associated to the 56.x.x.x milestone that are not related to any of the above will be resolved afterwards. EDIT: New issues can be created for specific parts of the tasks as necessary (i.e. don't create an issue unless it ends up involving quite a bit of work or requires discussion)

lenormf:
Will you experiment with community builds for this release? I have a `docker` container dedicated to building the project, so I can help on that front if needs be.

Eloston:
@lenormf Thanks for offering. I'm not sure how the new system we're designing will work out in practice, so keep that on standby in case we need it.

gcarq:
I'm currently compiling version 56.0.2924.76, some patches needed an update. The progress is being tracked here: https://github.com/gcarq/inox-patchset/issues/54

Eloston:
@gcarq Alright, thanks for the heads up.

triceratops1:
@eloston 
could you create "regional" issues
1 issue per issue

Eloston:
@triceratops1 No, I will not create "regional issues" unless they are necessary. Please see the edit in the original post.

I appreciate your status updates related to the tasks. As a result, I have included additional status information for all of the tasks that I will try to keep up-to-date. This should remove the need for unnecessary notifications.

[...]

PR #187, opened 2017-02-10
Update Inox's patches to version 56.0.2924.87

9Morello:
Update Inox patches to the latest Chromium stable release.


Eloston:
I actually want development to occur on `develop` from now on (which has the new file structure). I've updated #176 to reflect this change. Also, I usually update Debian patches before updating Inox patches, so I don't know if there's a conflict yet.

Sorry for the inconvenience. In the future, please let me know via #176 what tasks you want to work on and we can discuss it on there or some other place.

PR #222, opened 2017-04-26
Add note to macOS widevine FAQ

lmcmenemie:
Note that there is no slash following the first path in the copy command. If a slash is included - which many shell autocompletions will do automatically - then the command will copy only the contents of the directory instead of the directory itself.

I made this mistake once; this note might save somebody 10 minutes of wondering.


Eloston:
Looks good.

PR #253, opened 2017-08-21
Fix macos/widevine patch

tectiv3:


Eloston:
Looks good.

BTW, are you not using `quilt` to refresh the patch? I noticed the filenames are different in the patch header, which I think `quilt` removes if you use it to manage patches. There are some usage examples in `DEVELOPING.md`

tectiv3:
no, just `diff`.
I didn't touch the filenames though, I think it's just github patch syntax highlighting.

PR #255, opened 2017-08-22
Fix gn-bootstrap for macos build

tectiv3:
Add a new patch for that.


matthewbauer:
Hi,

Have you tried to upstream this to Chromium? It is still needed for me.

PR #258, opened 2017-09-05
simplify mac build instructions

pjv:
turn 3 lines into 1 that does the same thing


Eloston:
Huh, neat.

pjv:
Pretty sure you can do the same `mkdir` on linux.

PR #273, opened 2017-10-17
update FAQ widevine

solinium:


Eloston:
I modified your instructions for it to apply more generally.

PR #303, opened 2017-12-07
63.0.3239.108 on openSUSE

LeFroid:
This set of changes will update ungoogled chromium to version 63.0.3239.84 as well as fix some minor issues with the OpenSUSE build configuration

Feel free to ignore the changes to utilikit/export_resources.py and utilikit/_build_files_generators/linux_simple.py , they were made in order to have utilikit work with the version of Python on my system.

One other change I'd like to mention is the addition to the file resources/packaging/linux_simple/build.sh.in. This was modified to pull in NodeJS because the source chromium now requires it in order to compile, however they did not bundle NodeJS with the rest of the code. I could not find any build flags to disable this, the flag use_vulkanize=false may have done this in the past, but that flag (and a few others) were removed with this update.


resources/configs/common/patch_order
```diff
@@ -17,14 +17,12 @@ inox-patchset/0018-disable-first-run-behaviour.patch
...
 inox-patchset/0021-disable-rlz.patch
 inox-patchset/9000-disable-metrics.patch
-inox-patchset/9001-disable-profiler.patch
 
 debian/disable/promo.patch
 debian/disable/fuzzers.patch
 debian/disable/google-api-warning.patch
 debian/disable/device-notifications.patch
 debian/fixes/mojo.patch
-debian/fixes/crc32.patch
```
Eloston: I'm guessing these patches no longer applied cleanly (or at all)?

resources/configs/opensuse/gn_flags
```diff
@@ -1,7 +1,14 @@
...
 is_component_ffmpeg=true
+use_pulseaudio=true
 link_pulseaudio=true
-use_allocator="tmalloc"
+use_allocator="tcmalloc"
 use_aura=true
+use_gconf=true
 use_gtk3=true
 use_kerberos=true
+use_gold=true
```
Eloston: At this rate, perhaps it would help to make `opensuse` inherit from `linux_rooted` instead of `common`? Or are there changes in `linux_rooted` you don't want?

resources/patches/opensuse/chromium-dma-buf.patch
```diff
@@ -0,0 +1,20 @@
+--- a/ui/gfx/linux/client_native_pixmap_dmabuf.cc
```
Eloston: Would you mind adding a short description for this patch? I can't tell what this does from the changes alone.

resources/configs/common/patch_order
LeFroid: That's correct
The file crc32c.h essentially upstreamed the changes from debian/fixes/crc32.patch in v63.
As for inox-patchset/9001-disable-profiler.patch, I removed that because it was removed in the v63 branch of the inox-patchset repo.
Most of the other inox-patchset changes in this PR come from my merging of the branch https://github.com/gcarq/inox-patchset/tree/feature/63 into the ungoogled chromium version as well

resources/configs/opensuse/gn_flags
LeFroid: They do seem to have much in common. I'll try chaning the base config to linux_rooted on the next minor version update and see how it goes.

resources/patches/opensuse/chromium-dma-buf.patch
LeFroid: No problem

resources/patches/ungoogled-chromium/use-local-devtools-files.patch
```diff
@@ -41,7 +41,7 @@
 +        { "name": "product_registry_impl", "condition": "!v8only" },
          { "name": "profiler" },
          { "name": "resources", "condition": "!v8only" },
-         { "name": "audits", "condition": "!v8only" },
+         { "name": "audits", "condition": "!v8only", "type": "remote" },
```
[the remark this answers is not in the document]
LeFroid: I don't recall making any changes to this patch, it's probably that the upstream version added or modified the item and I did not notice it

resources/patches/opensuse/chromium-dma-buf.patch
Eloston: Ah, I was more interested in knowing what the change does so I could determine if I needed to include it in `linux_rooted` or not (like the user namespace sandbox patch). But since the source doesn't specify, I'll probably leave it in `opensuse` for now.

resources/configs/common/patch_order
Eloston: Good to know. Thanks!

resources/patches/ungoogled-chromium/use-local-devtools-files.patch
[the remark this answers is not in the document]
Eloston: No problem. I'll change it after I merge.

resources/patches/ungoogled-chromium/disable-google-host-detection.patch
```diff
@@ -7,18 +7,18 @@
  
  bool HasGoogleHost(const GURL& url) {
 -  static const char* kGoogleHostSuffixes[] = {
--      ".google.com",
```
Eloston: Looks like domain substitution tripped you up too huh

LeFroid: Yeah, I fixed this patch manually in the last update and forgot to consider the domain substitution! Just uploaded a fix

Eloston:
I decided that it'll be easier to update based on your changes than fix conflicts later. So, I'm merging now. Thanks for the work!

PR #323, opened 2018-01-07
Fix bootstrap patch

tectiv3:


Eloston:
That looks about right. Thanks :+1:

PR #327, opened 2018-01-14
Reverting some minor changes

dubvulture:
Explicit  `--parallel` for debianhelper 9 and removed redundant `-j` argument.


Eloston:
Would you be able to clarify why `-j` is redundant? The patch that adds this flag is `resources/patches/debian/gn/parallel.patch`

dubvulture:
`-j` options is already included while defining `njobs` [here](https://github.com/Eloston/ungoogled-chromium/blob/develop/resources/packaging/debian/standard/rules.in#L42).
Plus, `njobs` might not be defined and `bootstrap.py` will complain about the dangling `-j`.

Eloston:
Ah, missed that.

LGTM :+1:

PR #328, opened 2018-01-16
Resolved fuzz in minimal.patch

dubvulture:
See [comment](https://github.com/Eloston/ungoogled-chromium/issues/37#issuecomment-357717417)
Now it should work even without `--ignore-whitespace` option.

### DIFF
--- a/resources/packaging/debian/minimal.patch
+++ b/resources/packaging/debian/minimal.patch
@@ -65,7 +65,7 @@
   libharfbuzz-dev,
 --- a/rules.in
 +++ b/rules.in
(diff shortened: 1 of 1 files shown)


T-vK:
I tried it on gitlab and it still failed: https://gitlab.com/T-vK/ungoogled-chromium-dev-test/-/jobs/48375840

Then I added the suggested flag and it also failed: https://gitlab.com/T-vK/ungoogled-chromium-dev-test/-/jobs/48376514

dubvulture:
That error is not related to that patch or any of the files it modifies. (still, you should report that on the main issue page)

Eloston:
LGTM. Thanks.

PR #333, opened 2018-02-18
Add enable-potentially-annoying-security-features flag to chrome://flag for all OS types

kfur:


Eloston:
LGTM, but it's not referenced by any patch order config file. I'll add that in for you.

BTW, what does this flag do exactly? There's not much information about it online.

kfur:
@Eloston i don't know a lot about it flag. But at least it block all mixed content in pages. When by default chromium block only active mixed contend.

Eloston:
Hmm, alright. I guess I'll have to dig around the code if I want to find out.

PR #339, opened 2018-02-26
Patch for disabling crashpad handler on macOS

Artur96:


Artur96:
Didn't know about that, will do.

PR #341, opened 2018-02-27
linux_simple packaging fixes

squalus:
These are some fixes for linux_simple packaging problems that I ran into while trying to build on a Gentoo system

- Script resources (build.sh, packaging.sh, list_build_outputs.py) were not executable after being copied
- linux_simple.py failed because the ungoogled_packaging/scripts directory did not exist
- package.sh failed with `tar: Invalid transform expression`
- package.sh failed with `tar: You must specify one of the '-Acdtrux', '--delete' or '--test-label' options`

Perhaps some of this is related to version differences in the various utilities, I don't know. I'm on an updated Gentoo `~amd64` system

### DIFF
--- a/buildkit/packaging/linux_simple.py
+++ b/buildkit/packaging/linux_simple.py
@@ -47,6 +47,7 @@ def generate_packaging(config_bundle, output_dir, build_output=DEFAULT_BUILD_OUT
 
     ensure_empty_dir(output_dir) # Raises FileNotFoundError, FileExistsError
     (output_dir / 'scripts').mkdir()
+    (output_dir / 'archive_include').mkdir()
 
     # Build and packaging scripts
     _copy_from_resources('build.sh.in', output_dir)
--- a/resources/packaging/linux_simple/package.sh.in
+++ b/resources/packaging/linux_simple/package.sh.in
@@ -6,14 +6,17 @@ TARPREFIX=ungoogled-chromium_$ungoog{version_string}_linux
 CURRENTDIR=$(dirname $(readlink -f $0))
 # Assume buildspace tree is outside this script's directory
 BUILDSPACE_TREE=$(dirname "$CURRENTDIR")
-ARCHIVE_OUTPUT=$(dirname "$BUILDSPACE_TREE")/$TARPREFIX.tar.xz
+ARCHIVE_OUTPUT=$(dirname "$BUILDSPACE_TREE")/$TARPREFIX.tar
 
 # Include main build outputs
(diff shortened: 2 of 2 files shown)


Eloston:
Thanks for finding these bugs. Could you please test 204465832ce82335e89122eb0feec655c9c377a9?

Although I appreciate the PR, I wanted to make the changes differently for the following reasons:

* `output_dir.mkdir(parents=True, exists_ok=True)` is dangerous because it could potentially make a long chain of unwanted directories or create files in an unexpected location if the path was specified incorrectly.
 (Generally, I prefer to fail more aggressively so human error won't cause unexpected behavior.)
* `shutil.copy` does the same thing as `copyfile` and `copymode` combined
* `tar` can create compressed tar archives directly, but I forgot to specify [the required compression type](https://www.gnu.org/software/tar/manual/html_node/gzip.html).
* Using forward slashes in the transform expression can cause problems if `$TARPREFIX` is multiple directories deep. [The tar documentation used commas as a delimiter in this scenario](http://www.gnu.org/software/tar/manual/html_section/tar_51.html), but I forgot to specify the delimiter following the replacement string in the transform expression.

squalus:
With your version I get these errors:

```
2018-02-27 22:17:57,313 - ERROR: Parent directories do not exist for path: [Errno 2] No such file or directory: 'buildspace/tree/ungoogled_packaging/archive_include/README'
```

Looks like archive_include directory is not created. Adding `(output_dir / 'archive_include').mkdir()` before the README copy fixes it for me

Then running the package.sh script, adding the README file to the archive fails. This was the reason I did the compression in an extra step - I couldn't find a way to add a file to the compressed archive in-place.
```
+ tar --transform 's,^,ungoogled-chromium_64.0.3282.186-1_linux,' -r -J -v -f /src/ungoogled-chromium/buildspace/ungoogled-chromium_64.0.3282.186-1_linux.tar.xz --null -T -
tar: Cannot update compressed archives
Try 'tar --help' or 'tar --usage' for more information.
```

squalus:
Also, the archive created results in files that look like this: `ungoogled-chromium_64.0.3282.186-1_linuxchrome-wrapper`. I'm assuming the first part of this should be a directory instead, like `ungoogled-chromium_64.0.3282.186-1_linux/chrome-wrapper`

squalus:
Updated with a new proposed patch that's working for me

Eloston:
> This was the reason I did the compression in an extra step - I couldn't find a way to add a file to the compressed archive in-place.

Ah, I didn't realize this. Another option would be to combine the tar commands into one, but it probably isn't worth the effort.

PR #349, opened 2018-03-10
Update chromium to 65.0.3325.146

squalus:
- update debian patches from 65.0.3325.85-1 branch
- update inox patches from 65.0.3325.146 branch
- remove obsolete fix-icu-tangut patch (the bad symbols don't appear to be there anymore)
- fix http status symbol build error in disable-webrtc-log-uploading patch
- refresh other patches referenced from the config bundles, excluding windows
- add extra binary to the pruning exclusion list due to build error

Tested a linux_simple build on gentoo ~amd64 and it seems ok to me using it for a few hours. Patches for macOS and other Linux configs were refreshed but untested. Windows patches weren't touched at all.


Eloston:
Thanks! I'll merge this after I merge xsmile's PR for Arch Linux (which should be after the build is confirmed working).

resources/patches/debian/system/gtk2.patch
```diff
@@ -1,13 +0,0 @@
-description: add missing include needed to build with gtk2
```
Eloston: Was the GTK2 patch removed in Debian?

developer_utilities/update_lists.py
```diff
@@ -34,6 +34,7 @@
 # pathlib.Path.match() paths to exclude from binary pruning
 PRUNING_EXCLUDE_PATTERNS = [
     'chrome/common/win/eventlog_messages.mc', # TODO: False positive textfile
+    'chrome/browser/resources/media/mei_preload/preloaded_data.pb',
```
Eloston: Do you know what this file is used for?

resources/patches/debian/system/gtk2.patch
squalus: Yeah. It seems they fixed the gtk2 build error upstream

developer_utilities/update_lists.py
squalus: Not really. MEI = media engagement index, which is about collecting analytics in order to determine whether to allow videos to auto-play

It does break the build if not included. It's only 14 bytes and doesn't contain anything interesting if printed with `protoc --decode_raw`. 

```
media/base/media_switches.cc
// Allows Media Engagement to use preloaded data to decide whether an origin has
// a high media engagement.

// - PreloadMediaEngagementData: enables a list of origins to be considered as
//   having a high MEI until there is enough local data to determine the user's
//   preferred behaviour.

```

Maybe in an official branded build they preload a larger file with real data in it?

squalus: The proto is at `chrome/browser/media/media_engagement_preload.proto`

```
message PreloadedData {
  // The data in DAFSA format.
  optional bytes dafsa = 1;
}
```

Some more references:
`chrome/browser/media/media_engagement_preloaded_list.cc`
`chrome/browser/media/media_engagement_session.cc`
 `net/tools/dafsa`

squalus:
macOS is now tested and working for me

developer_utilities/update_lists.py
xsmile: Alternatively the mei_preload files can be excluded with this patch:
```
--- a/chrome/BUILD.gn
+++ b/chrome/BUILD.gn
@@ -304,7 +304,6 @@ if (!is_android && !is_mac) {
       }
 
       data_deps += [
-        "//chrome/browser/resources/media/mei_preload:component",
         "//third_party/widevine/cdm:widevinecdmadapter",
       ]
 
@@ -1239,7 +1238,6 @@ if (is_win) {
       ":widevine_cdm_library",
       "//build/config:exe_and_shlib_deps",
       "//chrome/app/nibs:chrome_xibs",
-      "//chrome/browser/resources/media/mei_preload:component_bundle",
     ]
 
     if (is_chrome_branded) {
```
I didn't notice anything unusual with such a build yet.

resources/config_bundles/linux_rooted/patch_order.list
```diff
@@ -1,17 +1,14 @@
-inox-patchset/chromium-vaapi-move.patch
-inox-patchset/chromium-vaapi-init.patch
-inox-patchset/chromium-vaapi-rgbx.patch
+inox-patchset/chromium-vaapi-init-r16.patch
 inox-patchset/chromium-vaapi-r16.patch
 
 debian/master-preferences.patch
 debian/system/icu.patch
 debian/system/vpx.patch
```
xsmile: The vpx and lcms2 patches can also be removed for Chromium 65, see the [Debian series file](https://sources.debian.org/src/chromium-browser/65.0.3325.146-1/debian/patches/series/).

squalus: You're right, I'll remove them

developer_utilities/update_lists.py
squalus: I prefer excluding these files too

Eloston: Yeah, I don't see much benefit of including them in this case.

@squalus Could you include @xsmile's change as a new patch in the `common` base bundle (and also refresh any subsequent patches that also touch this file)?

Eloston: Nevermind I can merge now and do it.

Issue #345, opened 2018-03-03
Update Windows support (continuation)


Eloston:
For some reason, #215 causes GitHub's Desktop webpage to crash with Server Error 500, but not their Mobile version. Since this is a bit inconvenient for everyone, this issue is a continuation of the discussion in #215.

Just in case #215 breaks completely, an archive of the Mobile version is here: https://gist.github.com/Eloston/24dab695470deba4c20a0b5683e160b4

mikerockett:
Out of curiosity, does the decision to move to Clang help this issue in any way?

https://mspoweruser.com/googles-chrome-ditches-microsofts-compiler-clang-windows/

Eloston:
@mikerockett The current state is described here: https://github.com/Eloston/ungoogled-chromium/issues/332#issuecomment-366890227. Basically, the Windows build *almost* works, except for some Windows-specific code that broke due to GN configuration (which I suspected in the linked comment to be caused by disabling Safe Browsing).

The move to Clang actually slowed me down a bit, because I didn't want to have to setup yet another tool to build Chromium (for purists, this potentially means yet another tool to build from source). In the end, I gave in because some non-trivial problems occurred with the Microsoft compiler.

If anyone's interested, BUILDING.md contains the steps and some notes I used to build Chromium up to the point of failure. I think all it takes to make a working build is to patch out some more obscure references to Safe Browsing-related components. But, there may also be something wrong with the changes I made to the GN files. Faster said than done for me at least.

(A bit unrelated, but I found it interesting that the Chrome Cleaner (downloads and runs the Chrome Cleanup Tool) is implemented within Safe Browsing. Safe Browsing's been more than just a website blacklisting service for a while now, but this is pushing the boundaries even further.)

mikerockett:
I see. I’m quite unversed when it comes to much of this, but a bit of insight is always good, so thanks for that. 👍

Eloston:
@mercerius (This is a response to https://github.com/Eloston/ungoogled-chromium/issues/235#issuecomment-370989331)

I forgot to mention that I disabled the 260 character limit for file paths. I'll add a note about that to the BUILDING instructions.

Eloston:
Status update for version 65: With #351, the Safe Browsing patch needs to be refreshed.

Eloston:
@squalus With #356, do you still get the same problem that I got?

squalus:
[...]

PR #359, opened 2018-03-19
windows: Fix build

squalus:
* disable multi-dll build to avoid linker errors with iridum trk
  patches and profile-related code
* add new patch to fix non-multi-dll linker errors
* in windows safebrowsing patch, hack out a few lines of code causing
  crash on startup when creating a new profile


squalus:
I used this config to get a successful build that withstood a few minutes of testing. I used LLVM 6.0.0

Eloston:
Awesome!

Would you be able to post a binary to #345 for others to test? Thanks.

squalus:
I'd rather not post binaries as the security of my Windows system is in doubt. If I can put together a better Windows environment then I may be able to do this in the future though

Eloston:
I see. I'm sure we'll figure out some way to get a clean build soon enough.

squalus:
Here are the outstanding issues I encountered. Thanks for the Windows building notes, those were helpful and some of these are noted there.

- build.bat script did not copy the args.gn file. Don't know why yet. Windows batch files are awful!
- quilt needed the series file to be in unix rather than windows (CRLF) format. Though this may have been because I was using WSL instead of MSYS2. I'll need to check this
- build.bat needs to have gperf/bison env vars set
- I needed to run the "x64 Native Tools" command prompt rather than use the entries in build.bat. Not sure why
- LLVM needs to be copied into tree (I have a patch coming to do this in getsrc)
- No packaging or installer at the end. I tried using `ninja mini-installer` but the installer did not start up. Haven't looked into it yet.

squalus:
Also, I used the 64 bit LLVM 6.0.0 official binaries from llvm.org

PR #360, opened 2018-03-19
buildkit: Add support for copying LLVM on Windows

squalus:
* On Windows, buildkit getsrc will try to locate the installed LLVM
  release and copy it into the buildspace tree
* Added --windows-llvm-path option to getsrc to allow users to use a
  custom version of LLVM instead

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -97,7 +97,8 @@ When installing the SDK, the "Debugging Tools for Windows" feature must be enabl
 1. Setup the following:
 
     * [LLVM](https://llvm.org/) 6.0.0 or newer.
-        * To use LLVM's pre-built binary, download the `Clang for Windows (64-bit)` version and unpack it with 7-zip into third_party/llvm-build/Release+Asserts in the buildspace tree.
+        * To use LLVM's pre-built binary, download and install `Clang for Windows (64-bit)` from [https://releases.llvm.org/download.html](https://releases.llvm.org/download.html)
+        * To use a custom LLVM release, use the `--windows-llvm-path` option on the `getsrc` step
         * *Developer note*: If the current stable version of LLVM is causing problems with the build, try matching Google's LLVM version (defined by the `CLANG_REVISION` variable in by downloading a snapshot build at the version specified by `CLANG_REVISION` and `VERSION` constants in `tools/clang/scripts/update.py`. For example, revision 123456 of LLVM 9.8.7 64-bit Windows would be: `http://prereleases.llvm.org/win-snapshots/LLVM-9.8.7-r123456-win64.exe` (link derived from [LLVM Snapshot Builds](http://llvm.org/builds/))
     * Python 2 for scripts in Chromium
     * Python 3 for buildkit
--- a/buildkit/cli.py
+++ b/buildkit/cli.py
@@ -138,7 +138,7 @@ def _add_getsrc(subparsers):
         try:
             source_retrieval.retrieve_and_extract(
                 args.bundle, args.downloads, args.tree, prune_binaries=args.prune_binaries,
-                show_progress=args.show_progress)
+                show_progress=args.show_progress, windows_llvm_path=args.windows_llvm_path)
         except FileExistsError as exc:
             get_logger().error('Directory is not empty: %s', exc)
             raise _CLIError()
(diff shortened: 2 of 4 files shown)


squalus:
I tried to make this as automatic as possible.

The first thought is to use the same method as macOS to download and extract official LLVM binaries. But the Windows LLVM binaries only come in the form of an installer (as far as I can tell). And I couldn't find any way to extract their self-extracting installer on the command-line. 7-Zip GUI does it, but the standalone command-line 7za.exe does not. And anyway, that would be another new binary to have to bootstrap. Nor does Python3.6 lzma, unzip, xz, etc. work to do this.

So instead I provided two options:

1) User installs system-wide LLVM and buildkit automatically locates it from the Windows Registry
2) User manually extracts LLVM and provides the --windows-llvm-path option to getsrc

These only work if the platform is detected as Windows, so a Windows buildspace tree does need to be generated on a Windows system.

squalus:
A simpler implementation, but a little more work for the user, would be to allow the user to provide a path (instead of tar.xz file) in extra deps, then the user points to the already extracted LLVM path. To do that they would either edit an existing dummy extra_deps.ini in the windows bundle or create a new bundle with the extra_deps listed.

Eloston:
It seems that the LLVM installer is a NSIS installer, so I believe you may be able to use `/S` and `/D` options as explained here: http://nsis.sourceforge.net/Docs/Chapter3.html

If it works, we could extend extra dependencies to invoke this automatically. But, I'd rather try to find a way to extract the file instead of invoking it if possible.

squalus:
This still "installs" the software though right? Uninstall info would get recorded in the registry. It's weird to have this tool automatically perform a system-wide installation into a temporary path. I agree extracting is better

Eloston:
Well it seems that 7za can't extract all formats. You'll have to use 7z.exe.

To get 7z.exe, we can use the `.msi` format and use `msiexec` to extract the files into a directory, then invoke it to extract the llvm installer.

As much as I like your code, I don't see much of a reason to copy a pre-installed LLVM toolchain installed to the system; in the future, we may need to switch back to a specific snapshot version. Copying from an existing path is nicer, but you might as well extract it directly to the destination path at that point.

There's not really a mechanism to have extra dependencies depend on each other, and I'm kind of disappointed in the performance of my hacky tar extraction script, so here's what I'm thinking:

* Have the user obtain 7-zip somehow (either a portable version or install it via the installer)
* For the CLI, add an extra optional flag `--7zip-cli` to specify a path to `7z.exe`. If it isn't specified, we default to using the registry to find it (on Linux, `which 7z` is used). Additionally, `--gnu-tar`, which specifies the path to GNU tar (default is `which tar`).
* Extend extra dependencies mechanism to support an `extractor` field. Use `tar` to extract with a tar-capable extractor (first GNU tar, then 7zip, then the internal extractor), `7zip` to force extraction with 7zip, `gnu_tar` to force GNU tar, and `internal_tar` to force the internal extractor.
* If necessary, add an `arguments` or `type` field to specify additional arguments to the utilities needed for extraction.

Thoughts?

squalus:
Thanks, I didn't realize the 7z vs 7za issue. I do like the idea of bootstrapping as many dependencies as we can for Windows since it reduces work for the user, increases repeatability (we know we all have the same version of tools), and increases security since we can validate the hashes on each tool.

Your proposal is better than what I have in this PR. How about this slightly modified proposal that includes 7-zip downloading?

- Create `extractor` field in extra_deps.ini. Supported values:
  - `7zip`: Extracts a 7-zip file
  - `msi`: Extracts MSI file using `msiexec`
  - `tar`/`gnu_tar`/`internal_tar`: Extracts tar file using logic you described
- Add `provides` field in extra_deps.ini
  - The 7zip dependency will be listed with `provides = 7zip`
  - The 7zip extractor will use this map to look up the output path of 7zip, e.g. `provided_deps['7zip'].output_path` or something
  - This allows one extractor to depend on a previously downloaded artifact as long as the order is correct
- Download 7-zip with `msi` extractor in the first extra_deps entry
- Download LLVM with `7zip` extractor in the second extra_deps entry

squalus:
Never mind about my comment above. If we want to do this we can do it as a second pass. I have a PR incoming implementing your suggestions

PR #361, opened 2018-03-20
buildkit: Add support for downloading LLVM on Windows

squalus:
* Add support for multiple extractors in buildkit
* Add support for native tarfile extraction using tar and 7z utilities
* Add extra dependency for LLVM windows binaries

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -96,9 +96,6 @@ When installing the SDK, the "Debugging Tools for Windows" feature must be enabl
 
 1. Setup the following:
 
-    * [LLVM](https://llvm.org/) 6.0.0 or newer.
-        * To use LLVM's pre-built binary, download the `Clang for Windows (64-bit)` version and unpack it with 7-zip into third_party/llvm-build/Release+Asserts in the buildspace tree.
-        * *Developer note*: If the current stable version of LLVM is causing problems with the build, try matching Google's LLVM version (defined by the `CLANG_REVISION` variable in by downloading a snapshot build at the version specified by `CLANG_REVISION` and `VERSION` constants in `tools/clang/scripts/update.py`. For example, revision 123456 of LLVM 9.8.7 64-bit Windows would be: `http://prereleases.llvm.org/win-snapshots/LLVM-9.8.7-r123456-win64.exe` (link derived from [LLVM Snapshot Builds](http://llvm.org/builds/))
     * Python 2 for scripts in Chromium
     * Python 3 for buildkit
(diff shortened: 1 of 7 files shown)


squalus:
This implements the suggestions from #360.

The only supported extractors are `7z` and `tar`. Untar strategies are chosen automatically, there's no support yet for explicit `internal_tar`, `gnu_tar`, etc as extractor fields.

The tar extractor is compatible with gnutar and bsdtar (for macos)

7z.exe is used to extract tars on Windows, with a fallback to Python. Auto-location is not yet implemented for Windows, so the `--7z-path` argument to `getsrc` is now required for a Windows build.

Eloston:
I've refactored your code a bit in 2cd39da2cbd584247fcb6c35487a19f71ff4571f, but it's missing the code to read from the registry. Could you add that in?

squalus:
Looks good!

I added the 7-zip registry lookup support and a small fix for the extra_deps schema.

I will also note that 7z doesn't need to be artifically limited to Windows-only. In its current state, it will work on linux/macos but only if you pass `--7z-path`

Eloston:
I made some more changes. If everything works I'll merge.

Eloston:
I confirmed that GNU tar decompression works, but I don't have a Windows machine on me. Would you be able toconfirm my 7-zip registry-related changes work?

squalus:
Added a small fix for the following error on Windows:

```
  File "D:\src\ungoogled-chromium\buildkit\cli.py", line 441, in main
    args.callback(args=args)
  File "D:\src\ungoogled-chromium\buildkit\cli.py", line 146, in _callback
    show_progress=args.show_progress, extractors=extractors)
  File "D:\src\ungoogled-chromium\buildkit\source_retrieval.py", line 213, in retrieve_and_extract
    pruning_set=remaining_files, extractors=extractors)
  File "D:\src\ungoogled-chromium\buildkit\source_retrieval.py", line 131, in _setup_chromium_source
    extractors=extractors)
  File "D:\src\ungoogled-chromium\buildkit\extractors.py", line 209, in extract_tar_file
    sevenzip_cmd = str(_find_7z_by_registry())
  File "D:\src\ungoogled-chromium\buildkit\extractors.py", line 40, in _find_7z_by_registry
    sevenzip_path = Path(sevenzipfm_dir / '7z.exe')
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

Otherwise it looks good.

Eloston:
Awesome :+1:

Issue #366, opened 2018-03-23
Remove build dependency on quilt


squalus:
Quilt is useful for patch maintainers, but not necessary for normal users. It can be replaced by a Python3 script that 1) reads the lines in the series file and 2) applies the patches using either `git apply` or `patch`. The same script can be used for all platforms.

This will be one less dependency for users to install. For Windows this is particularly beneficial beause it removes the need to install MSYS2. If we decide to use `patch`, the gnuwin32 binary can be downloaded automatically.

Eloston:
Alternatively, we could use [`python-quilt`](https://github.com/bjoernricks/python-quilt). But for non-developers, this adds a lot of complexity. It's simpler to just invoke `patch` directly.

> If we decide to use patch, the gnuwin32 binary can be downloaded automatically.

In the previous Windows instructions (before I attempted to update Windows support recently), I manually download MSYS2's `patch` binary from SourceForge (with the required DLL to make it run). I can't recall exactly why, but there may have been some issues with GNUWin32's old version of GNU `patch`. Since we can now automatically download and extract (almost) everything, we could just automate that procedure.

Eloston:
For now, I think it's better to play it safe by invoking `patch` directly and using MSYS2's `patch`. I think it'll be fine to include the patching code this as a new module `patching.py` with CLI command `patsrc` (for "patch source", unless you can think of a better name).

squalus:
What about using `git apply`? That works fine for me in testing on Windows. And we can download updated official binaries from https://git-scm.com in the extra deps.

Would the new buildkit command run between `subdom` and `genpkg`? And I guess we no longer need `genpkg` to copy patches or patching scripts to the tree.

squalus:
Also it appears `python-quilt` just invokes `patch` :)

https://github.com/bjoernricks/python-quilt/blob/3dd50bc77/quilt/patch.py#L26

squalus:
A few more things
- The Windows git binaries actually include patch.exe so it's possible we could use this
- If `patsrc` is run after `getsrc` it's no longer necessary to perform subdomain replacement on the patches, so that code could be deleted in the `subdom` step. Is this what you have in mind?

Eloston:
`git apply` requires Git to be downloaded. If you're not a developer, this is a large and non-portable dependency (IIRC it's a subset of MSYS). On the other hand, my GNU `patch` approach only requires two small downloads; of which only a single `exe` and `dll` are required from it.

[...]

PR #368, opened 2018-03-25
Add python patch apply script

squalus:
* create a script that uses 'patch' utility to apply patches
* update windows, macos, and linux_simple to use the new script
* quilt is no longer a build dependency for these platforms


squalus:
Windows, macos, and linux_simple are tested with the new script. Note that other platforms are free to continue using quilt if they want since the format of the files has not changed. Fixes #366

squalus:
I tested the Gnuwin32 patch binary and found that it didn't work. First, there's a bizarre problem with UAC that had to be worked around (http://irq5.io/2011/06/26/gnu-patch-and-windows-uac/). Next, I found that it crashed when applying some of the patches. I didn't investigate why.

So instead I got the portable git binaries and used the patch.exe from there. That worked fine for me locally and on Appveyor CI.

BUILDING.md
```diff
@@ -96,37 +96,39 @@ When installing the SDK, the "Debugging Tools for Windows" feature must be enabl
...
 ```
 mkdir buildspace\downloads
-py buildkit-launcher.py genbun windows
-py buildkit-launcher.py getsrc
-py buildkit-launcher.py subdom
-py buildkit-launcher.py genpkg windows
+python3 buildkit-launcher.py genbun windows
+python3 buildkit-launcher.py getsrc
+python3 buildkit-launcher.py subdom
+python3 buildkit-launcher.py genpkg windows
```
Eloston: The Python 3.5 and newer installers for Windows can install the `py` launcher. There's no `python3` command that is installed by the installer (AFAIK), and this can be confusing to users who copy these instructions verbatim.

From my experience, the `py` launcher will correctly read the shebang of `buildkit-launcher.py` and use Python 3 instead of Python 2 to run buildkit. If you're worried about this, you can just use `py -3` to force using Python 3.

Is there a reason why you chose to use `python3` here? If not, we could just explicitly tell users to ensure that they install the `py` launcher.

squalus: I think you're right here. I just wanted to make it clear it was Python 3, but your explanation makes more sense

PR #377, opened 2018-04-01
Add fingerprinting mitigation for getClientRects,  getBoundingClientRect

csagan5:
Scale the result of Range::getClientRects and Element::getBoundingClientRect
by a random +/-5% of the original value for each float in the Rect/Quad.
The random value is generated once for each Document and re-used for all the
attached elements.
The rationale is that such value, albeit constant within the rendered Document,
is within the same order of magniture of the floating point precision being
used for fingerprinting and sufficient to poison the well.

See also: http://www.gsd.inesc-id.pt/~mpc/pubs/fingerprinting-trustcom2016.pdf

### DIFF
--- a/resources/patches/bromite/BRM045_Add-fingerprinting-mitigation-for-getClientRects-getBoundingClientRect.patch
+++ b/resources/patches/bromite/BRM045_Add-fingerprinting-mitigation-for-getClientRects-getBoundingClientRect.patch
@@ -1,0 +1,130 @@
+From: csagan5 <32685696+csagan5@users.noreply.github.com>
+Date: Fri, 30 Mar 2018 10:09:03 +0200
+Subject: Add fingerprinting mitigation for getClientRects,
+ getBoundingClientRect
+
+Scale the result of Range::getClientRects and Element::getBoundingClientRect
+by a random +/-5% of the original value for each float in the Rect/Quad.
+The random value is generated once for each Document and re-used for all the
+attached elements.
+The rationale is that such value, albeit constant within the rendered Document,
+is within the same order of magniture of the floating point precision being
+used for fingerprinting and sufficient to poison the well.
+
+See also: http://www.gsd.inesc-id.pt/~mpc/pubs/fingerprinting-trustcom2016.pdf
+---
+ third_party/WebKit/Source/core/dom/Document.cpp | 14 ++++++++++++++
+ third_party/WebKit/Source/core/dom/Document.h   |  5 +++++
+ third_party/WebKit/Source/core/dom/Element.cpp  |  7 +++++++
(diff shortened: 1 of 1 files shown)


Eloston:
The 5% noise is actually from a cited source:

```
N. Nikiforakis, W. Joosen, and B. Livshits, “Privaricator: Deceiving
fingerprinters with little white lies,” in Proceedings of the 24th
International Conference on World Wide Web, 2015, pp. 820–830
```

That paper is available here: http://www.www2015.it/documents/proceedings/proceedings/p820.pdf

Unlike this patch, PriVaricator applies 5% offsets to more than just `get*ClientRect[s]`. Their results have shown minimal to no breakages for the top 1000 Alexa sites by comparing screenshots of the rendered sites.

However, there are still a few concerning aspects:
* I can't say with strong confidence that these sites are representative of the designs and structures of sites browsed by ungoogled-chromium users.
* The paper's breakage evaluation does not consider dynamic behavior for sites that have a heavier use of JavaScript for rendering, such as web games. Errors in these values may cause accumulative error throughout the runtime of said web applications.

While such concerns may be silly for people like us, I believe it's better to keep cautious when it comes to behavior-mutating changes like these, especially if they can introduce subtle (and potentially frustrating) bugs like this patch.

As such, I would like to have this feature behind a command-line flag or `chrome://flag` option. Not only does it offer configurability, but it also informs the user of potential issues this can cause.

I'm not too familiar with the DOM side of web application execution, but maybe we could read this flag once per tab, and then have it pass down through Document to individual Elements to reduce the cost of looking up command-line arguments. If we are to implement multiple fingerprinting methods in the future, perhaps we could create a generic fingerprinting argument that accepts a comma-separated list of fingerprinting deception methods, and store these as a bit field to be read by various components of the DOM implementation as necessary.

csagan5:
> The 5% noise is actually from a cited source:
> That paper is available here: http://www.www2015.it/documents/proceedings/proceedings/p820.pdf

I did not understand that the whole paragraph was about PriVaricator. Thanks.

> Unlike this patch, PriVaricator applies 5% offsets to more than just get*ClientRect[s]. Their results have shown minimal to no breakages for the top 1000 Alexa sites by comparing screenshots of the rendered sites.

Yet no source code was published, it would have helped similar implementations.

> As such, I would like to have this feature behind a command-line flag or chrome://flag option. Not only does it offer configurability, but it also informs the user of potential issues this can cause.

Sorry, no plans to add this. I will close this PR.

> I'm not too familiar with the DOM side of web application execution, but maybe we could read this flag once per tab, and then have it pass down through Document to individual Elements to reduce the cost of looking up command-line arguments.

There are similar flags already in Chromium, like those that are periodically added when a feature is being deprecated (see the deep and shadow CSS selectors for example). One can just look at how those implementations work to do the same.

PR #389, opened 2018-04-05
update_lists: Skip symlinks leading out of the buildspace tree

xsmile:
Symlinks leading out of the buildspace tree produce a ValueError on `relative_to(buildspace_tree)` and should be skipped when generating the binary pruning and domain substitution lists.


Eloston:
Good catch. Thanks.

PR #397, opened 2018-04-22
Update Windows build for 66

squalus:
Test build from this branch: https://github.com/squalus/ungoogled-chromium-binaries/releases/tag/66.0.3359.117-1-test1


xsmile:
I have been using squalus' build for a couple of days and it works fine so far.

ribatamu:
Yeap v66 is working, except the bug which is already reported for the crash when extensions access safebrowsing prefs.

PR #400, opened 2018-04-25
Fix macOS build

Tacolizard:
Build was failing because it couldn't find `//iridium:trknotify`. Not sure what that does, but I've managed to compile for version 65 by removing it from the macos patch order.


Eloston:
Closing this as wontfix in favor of #401.

PR #413, opened 2018-05-24
Update common files for 67

### LINKED ISSUE #386: widevine support under linux
https://github.com/Eloston/ungoogled-chromium/issues/73

Is this still required for widevine support to work under linux? If so, where are these files located?

xsmile:
**Toolchain patches**
- inox-patchset/chromium-arflags.patch - Re-formats arflags for llvm-ar
- inox-patchset/chromium-clang-compiler-flags.patch - Removes compiler flags not yet supported by the system toolchain

**Chromium patches**
- inox-patchset/chromium-enable_widevine.patch - Fixes building with 'enable_widevine=true' on Linux. The DRM stream test at https://bitmovin.com/demos/drm works if libwidevinecdm.so is installed, so hopefully the new release will fix #386

**Removed patches that are now upstream**
- inox-patchset/fix-crash-in-is_cfi-true-builds-with-unbundled-ICU.patch
- inox-patchset/fix-frame-buttons-rendering-too-large-when-using-OSX.patch
- debian/fixes/optional.patch 
- debian/fixes/duplicate-name.patch
- debian/fixes/overloaded-call.patch

**Misc**
- debian/system/icu.patch - The patch is updated but I'm not sure if it is still needed
- build target 'widevinecdmadapter' was removed - References to it and the resulting library need to be removed from the packaging scripts
- binary pruning and domain substitution lists need to be updated when the stable version is released


xsmile:
That was quick, great.

Issue #414, opened 2018-05-24
Update to version 67


Eloston:
This issue tracks the upgrade progress to Chromium 67, subsequent revisions, and related discussion.

Here are the steps necessary towards the first release in the upgrade process:
* [x] Update `common` and other abstract base bundles
* [ ] Update other base bundles and associated packaging types:
    * [ ] Arch Linux
    * [x] Debian and Ubuntu
    * [x] macOS
    * [x] openSUSE
    * [x] Portable Linux
    * [x] Windows

NOTE: A base bundle is considered "updated" once its current configuration has been added/removed/changed to produce working builds (i.e. runs without severe regressions) on the new version. It does **not** always consider new Chromium features.

After the initial release is published, this issue will remain open for status updates and discussions relating to new revisions or the version as a whole. It will be closed once a new Specific problems should still be posted in a separate issue.

Miscellaneous TODO:
* [x] build target 'widevinecdmadapter' was removed - References to it and the resulting library need to be removed from the packaging scripts
* [x] binary pruning and domain substitution lists need to be updated when the stable version is released

Eloston:
1b9bc64c465543101f7944aee7e8a52087d40344 can build Debian stretch packages that can't play any media. I need to refresh Portable Linux patches and try building.

Eloston:
**Status update**: Portable Linux seems to be fully functional as of fb6a3a450508ba633914aa6a2924aeb42a620e64 with [a build available here](https://github.com/Eloston/ungoogled-chromium-binaries/releases/tag/67.0.3396.62-1)

LeFroid:
The OpenSUSE build is also having the issue where media cannot be played.

tectiv3:
hey, I am building for macOS atm and this line https://chromium.googlesource.com/chromium/src/base/+/master/mac/sdk_forward_declarations.h#288
does not work on 10.11 
```
error: expected ';' after top level declarator
typedef NSString* VNImageOption NS_STRING_ENUM;
```

Got to replace it with ```typedef NSString* VNImageOption;```
Should be fine on newer systems though.
I am hesitant to creating a patch for this since it looks like an upstream problem and it affects only people on older systems.

xsmile:
Does media playback with the faulty builds produce any logs in the console or on the chrome:gpu, chrome:media-internals pages?

Eloston:
[...]

PR #418, opened 2018-06-04
Update patches for macOS

tectiv3:
Updated safe browsing related patches and added patch for older sdk compatibility.
#414


resources/patches/ungoogled-chromium/macos/fix-older-sdk-declarations.patch
```diff
@@ -0,0 +1,11 @@
+--- a/base/mac/sdk_forward_declarations.h
++++ b/base/mac/sdk_forward_declarations.h
+@@ -285,7 +285,7 @@
+ @end
+
+ // VNImageRequestHandler forward declarations.
+-typedef NSString* VNImageOption NS_STRING_ENUM;
++typedef NSString* VNImageOption;
```
Eloston: If you're worried about this breaking compatibility on newer systems, maybe we could use a compiler macro that goes by the SDK version?

resources/config_bundles/macos/patch_order.list
```diff
@@ -4,4 +4,5 @@ ungoogled-chromium/macos/fix-gn-bootstrap.patch
 ungoogled-chromium/macos/fix-gn-safe_browsing.patch
 ungoogled-chromium/macos/fix-mapped_file.patch
 ungoogled-chromium/macos/fix-visibility.patch
-ungoogled-chromium/macos/macos-fix-building-without-safebrowsing.patch
+ungoogled-chromium/macos/fix-building-without-safebrowsing.patch
```
Eloston: I'm thinking it may be better to move this to the `common` bundle because these changes may be nice for other platforms.

resources/patches/ungoogled-chromium/macos/fix-older-sdk-declarations.patch
tectiv3: thing is it's already inside such macro

resources/config_bundles/macos/patch_order.list
tectiv3: ok

PR #421, opened 2018-06-11
Create standalone-download-and-compile.bat

johnakki:
Standalone download and compile script for Windows.
If using this, you should not clone the entire repo. Just download and run the script.


Eloston:
Thanks for the PR, but I was already considering rewriting the build script in Python 3 for the redesign. The redesigned script will also be more automated, so I don't think there'll be much value in publishing a wrapper batch script.

PR #442, opened 2018-07-27
Update common files for 68

xsmile:
- inox-patchset
  - removed 0002-fix-building-without-reporting.patch - build finished without it
  - removed 9000-disable-metrics.patch - high-maintenance
- binary pruning list exclusions
  - `chrome/android/profiles/afdo.prof` - seems to be an Automatic Feedback Directed Optimizer (AFDO) profile for Clang
  - `chrome/browser/resource_coordinator/tab_ranker/example_preprocessor_config.pb` - seems to be a neural network model to aid in memory management of tabs


Eloston:
Looks like pylint updated and broke the entire check. The config looks good, so I'll merge.

PR #445, opened 2018-08-02
Fix domain_substitution.list to generate Visual Studio Project with gn.exe

Ti-R:
Fix domain_substitution.list to enable to generate Visual Studio Project with gn.exe

### DIFF
--- a/resources/config_bundles/common/domain_substitution.list
+++ b/resources/config_bundles/common/domain_substitution.list
@@ -10481,7 +10481,6 @@ tools/gn/args.cc
 tools/gn/bin/gn-format.py
 tools/gn/bin/roll_gn.py
 tools/gn/standard_out.cc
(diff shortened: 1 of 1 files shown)


Eloston:
Thanks for finding the files. I will add exceptions to the list generation script later after I finish up some work first.

TN0X:
@Eloston: you have updated the script "update_lists.py" in devtools with the exclude pattern for visual_studio_writer.cc, but it has no effect yet because of the following reasons: 

1.) the pattern is in the wrong section in PRUNING_EXCLUDE_PATTERNS. It should be instead in the DOMAIN_EXCLUDE_PREFIXES, because the substitution is executed in the domain_substition task.

2.) but these changes have no effects until any developer executes the update_lists.py script and pushs the changed domain_substitution.list to the repository

PR #446, opened 2018-08-02
Suggestion for replacing chromium logo by new ungoogled chromium logo

Ti-R:
To represent the "ungoogled" part inside the logo, a part of it got wash out of colors.
The other part of the logo got more vivid color to show how great this version is!


Eloston:
Thanks for the show of appreciation, but I currently do not want to change the Chromium branding as stated in the README.

PR #459, opened 2018-08-17
Update config and packaging for Arch Linux

xsmile:
CFI seems to break VA-API and the GN flag `use_cfi_icall=false` is used as a workaround. Should other distributions show the same issue, then it is probably better to move the flag to the `linux_rooted` config bundle instead.

### DIFF
--- a/config_bundles/archlinux/gn_flags.map
+++ b/config_bundles/archlinux/gn_flags.map
@@ -1,0 +1,1 @@
+'use_cfi_icall=false'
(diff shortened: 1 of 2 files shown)


Eloston:
This seems like it will break anything using VA-API, so even `linux_portable` should have this. I'll take care of moving the flag around.

BTW, this breakage was introduced in 68, right?

xsmile:
> BTW, this breakage was introduced in 68, right?

Yes, 67 worked well without the need for this flag.

PR #461, opened 2018-08-24
packaging/arch: Don't change directories while using buildkit

xsmile:
Plus some minor changes.

I didn't test the whole build process though.


Eloston:
Thanks. Didn't realize it was already in the repository directory.

PR #471, opened 2018-08-29
packaging/arch: Curl pkgver's hashes for sha256 verification

nikolowry:
I've been in the bad habit of locally changing the sha256sum to `SKIP`, so I was hoping we could dynamically fetch it instead. Graciously ripped from AUR's [`chromium-dev`](https://aur.archlinux.org/packages/chromium-dev/).

From a quick glance at the repo, I couldn't determine the significance or relevance of this source: `        chromium-$pkgver.txt::https://chromium.googlesource.com/chromium/src/+/$pkgver?format=TEXT`.  Is there any reason for it? If not could we remove it?


xsmile:
The text file was pulled by the official PKGBUILD a while back but it is not needed anymore, see the relevant [commit](https://git.archlinux.org/svntogit/packages.git/commit/trunk/PKGBUILD?h=packages/chromium&id=2d32bc7a7d23838f81fca30aada7eee6952b37f8). It should be removed.

nikolowry:
Thanks for the background @xsmile -- just removed it!

PR #474, opened 2018-08-29
[WIP] Update to 69

xsmile:
**Debian**
- used patches from the 69.0.3497.12-1 branch of chromium-browser

The rest of the changes mostly consists of refreshed and removed patches.

**Misc.**
- GN flag `use_cfi_icall` still causes issues
- `chrome://flags/#top-chrome-md` can be used to revert to the previous UI

EDIT: I really like the checks but I'm not sure what to do about error in *validate_patches*. The URL fails to load because gn was relocated from the chromium tree to its own repository.


Eloston:
Thanks for this. I'm going to merge this later since 69 isn't stable yet. At most, I will merge it as a new branch that will later be merged into `master`.

> EDIT: I really like the checks but I'm not sure what to do about error in validate_patches. The URL fails to load because gn was relocated from the chromium tree to its own repository.

Not only that, but the path is wrong. By inspecting the chromium official tar.xz for this version, it should now be located at `tools/gn/tools/gn/ninja_build_writer.cc`, not `tools/gn/ninja_build_writer.cc`.

It's not clear to me how they're downloading the GN code into the chromium official tar.xz yet; I'll need to dig a bit deeper so I can implement a non-hacky solution.

Eloston:
I extended the logic for `validate_patches.py`. You can use `devutils/update_patches.py` to speed up patch refreshing if you need it. (This and more is documented in `docs/developing.md`.)

xsmile:
I refreshed the patches but cannot figure out why `debian/gn/libcxx.patch` does not pass validation. Updating the patch manually with quilt or via `update_patches.py` does not alter its contents.

The remaining config bundles should probably be marked as outdated for now.

Eloston:
> I refreshed the patches but cannot figure out why debian/gn/libcxx.patch does not pass validation. Updating the patch manually with quilt or via update_patches.py does not alter its contents.

Before db11c83809c725d6e9b2aff5bbc80e8ee29993ab, I hardcoded the GN version used in the chromium-browser-official tar file. Since you changed the Chromium version, the GN version changed too. With db11c83809c725d6e9b2aff5bbc80e8ee29993ab, I implemented an idea I had to auto-detect the GN version (see the commit message for an explanation of the logic).

xsmile:
The current patches seem to be sufficient for the stable build 69.0.3497.81, they just need to be refreshed.

Eloston:
Awesome. Thanks for testing and the other PRs to fix my silly mistakes.

PR #482, opened 2018-09-05
Add searx. Resolves #282

### LINKED ISSUE #282: switch the default search from DDG to one that actually respects privacy
Ungoogled-Chromium replaces the default search from google to duckduckgo presumably to serve its privacy-consious users.  That's only a marginal improvement from a privacy standpoint.  DDG is yet another centralized service and it's also falsely positioned as a privacy-respecting option.  DDG does harm to privacy in [several ways](https://www.reddit.com/r/privacy/comments/4vgqrn/duckduckgo_illusion_of_privacy/d5y8tdw/), in partnership with bigger evils than google.

Consequently, I suggest a variety of relatively corporate-free options that are more suitable from a privacy standpoint:

 * searx.me ([FOSS](https://github.com/asciimoo/searx), decentralized, many [instances](https://github.com/asciimoo/searx/wiki/Searx-instances) [many of which offer onion URLs])
 * startpage.com (aka "ixquick.com"; results are largely google-influenced but users don't feed google)
 * deusu.de ([FOSS](https://github.com/MichaelSchoebel/DeuSu))
 * metager.de
 * gigablast.com
 * qwant.com
 * swisscows.com

I know I'm missing [some](http://searchenginearchive.com/open_closed.php), but I highly recommend searx.  Results are quite good, in fact better than duckduckgo b/c fewer cloudflare sites end up in the top results.  DDG's first page results are often polluted with cloudflare pages.  Startpage.com (my 2nd choice) would be interesting in the sense that results are close to Google's results without touching google, which may be in alignment with ungoogled-chromium apparent philosophy to be minimal and conservative in the ungoogling effort.

I propose populating the pre-configured "default" list of search engines with all of the above, using a searx instance as the default of the defaults, and removing the politically incorrect ones like Bing, Yahoo, and AOL, or at least tagging them as "[PRISM](https://en.wikipedia.org/wiki/PRISM_%28surveillance_program%29)" or "ill-advised". 

This will also help in light of this annoyance: users cannot control which search engines are short-listed  on the "default search settings" and which are in the "other search settings" list.

dakka2:
This will become the default engine unless it's already been changed as far as I can tell.

### DIFF
--- a/config_bundles/common/patch_order.list
+++ b/config_bundles/common/patch_order.list
@@ -103,5 +103,6 @@ ungoogled-chromium/add-flag-to-enable-potentially-annoying-security-features.pat
 ungoogled-chromium/disable-mei-preload.patch
 ungoogled-chromium/add-flag-to-force-punycode-hostnames.patch
 ungoogled-chromium/fix-building-without-safebrowsing.patch
+ungoogled-chromium/searx.patch
 
 bromite/fingerprinting-flags-client-rects-and-measuretext.patch
--- a/patches/ungoogled-chromium/searx.patch
+++ b/patches/ungoogled-chromium/searx.patch
@@ -1,0 +1,696 @@
+diff --git a/components/search_engines/prepopulated_engines.json b/components/search_engines/prepopulated_engines.json
+index 4e050c0..7ad8362 100644
+--- a/components/search_engines/prepopulated_engines.json
++++ b/components/search_engines/prepopulated_engines.json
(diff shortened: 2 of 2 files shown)


dakka2:
I don't know why this fails validation with your CI setup. It seems to apply cleanly for me.

Eloston:
Thanks. There was a small bug in my script which I solved in da719778ad9987e1d2fde3d2f2a70cb9663e134c

PR #493, opened 2018-09-11
Add build instruction for fedora opensuse and google cloud to building…

intika:
update building.md


intika:
maybe also adding the new section to the main index

PR #506, opened 2018-09-17
dont throw error when folder exists

pastmaster007:
this change skips re-download same archive file from remote location..
this will trim approx 600mb file download time for the next build..

### DIFF
--- a/packaging/macos/build.sh.ungoogin
+++ b/packaging/macos/build.sh.ungoogin
@@ -22,7 +22,7 @@ mkdir out/Default
 DOWNLOAD_CACHE="$packaging_dir/../../download_cache"
 
 pushd ungoogled_packaging
-mkdir "$DOWNLOAD_CACHE" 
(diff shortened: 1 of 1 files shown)


Eloston:
> this change skips re-download same archive file from remote location..

I'm not sure what you mean. Can you clarify? The only changes I see made to the packaging scripts is `-p` to mkdir...

Also, it looks like macOS patches apply cleanly now. You can fix the CI check by removing the `patches_outdated` marking from `bundlemeta.ini` of `macos`.

pastmaster007:
when i build first time, remote archives(chromium, llvm, google toolbox) downloaded into download_cache folder. I got few errors, becauses of patches are not up-to-date. Then i modified a patch and try to build again. I got error saying download_cache directory already exists can not be created and build process exited. So i made command to ignore if the folder already exists by adding -p flag. Now, build process does not fail, if folder exists. So, no need to re-download archives again.

PR #551, opened 2018-10-07
Add AppImage Build Script

intika:
For rhel based system gconf2 or lib64gconf2 need to be installed, but normally this will be the case because i think it's needed to build but not sure...

Build instructions : 
On portable build instruction after : 
```
./ungoogled_packaging/build.sh
./ungoogled_packaging/package.sh
```

Just run 
`./ungoogled_packaging/package.appimage.sh`


packaging/linux_simple/ungoogled-chromium.yml
```diff
@@ -0,0 +1,30 @@
+# # TODO: Add https://www.i-dont-care-about-cookies.eu/ - https://github.com/uBlockOrigin/uBlock-issues/issues/196#issuecomment-417963577
```
Eloston: I believe this TODO should be removed since it goes against the objective of sticking to the default Chromium experience.

Eloston:
By the way, the instructions for buliding an AppImage, Flatpak, or Snap should be updated and possibly renamed.

packaging/linux_simple/package.appimage.sh.ungoogin
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -eux
+
+CURRENTDIR=$(dirname $(readlink -f $0))
+
+pushd "$CURRENTDIR"
+
+if [ ! -d "./AppImages" ]; then
+  git clone https://github.com/AppImage/AppImages.git
```
Eloston: If `pkg2appimage` is a standalone script, it may be better just to retrieve it directly via HTTPS.

Otherwise, I think it would be better to use a shallow clone here to save a little bandwidth.

packaging/linux_simple/ungoogled-chromium.yml
intika: Yes indeed i just forget to remove it was on the original script

intika:
> By the way, the instructions for buliding an AppImage, Flatpak, or Snap should be updated and possibly renamed.

Will do that ;)

intika:
Updated and resolved as needed ;) 
Also tested
all good ;)

Issue #520, opened 2018-09-24
Proposal: exclusion of account/signin/gaia code


csagan5:
In light of recent directions of upstream ([1](https://blog.cryptographyengineering.com/2018/09/23/why-im-leaving-chrome/), [2](https://blog.lukaszolejnik.com/am-i-logged-in-or-not-gdpr-case-study-on-the-example-of-chrome-browser-change/)) I would like to completely rip out the relative code, more as a statement than anything else (the relevant functions have been always dormant in Bromite, and I think it is not a supported feature of ungoogled-chromium as well?).

This should cause some more maintenance work (as this area I bet is one of the most active upstream), but on the other hand work as a sort of "preservation project" to let us understand how Chromium without cloud works/worked, and be capable of keeping it in this state for longer by learning more about the internals.

My idea would be to change `chrome/browser/BUILD.gn` this way:

```diff
--- chrome/browser/BUILD.gn	2018-09-23 17:16:40.264000000 +0200
+++ chrome/browser/BUILD.gn.new	2018-09-24 23:25:36.239894176 +0200
@@ -1366,50 +1366,10 @@
     "shell_integration_mac.mm",
     "shell_integration_win.cc",
     "shell_integration_win.h",
-    "signin/about_signin_internals_factory.cc",
-    "signin/about_signin_internals_factory.h",
-    "signin/account_consistency_mode_manager.cc",
-    "signin/account_consistency_mode_manager.h",
-    "signin/account_fetcher_service_factory.cc",
-    "signin/account_fetcher_service_factory.h",
-    "signin/account_investigator_factory.cc",
-    "signin/account_investigator_factory.h",
-    "signin/account_reconcilor_factory.cc",
-    "signin/account_reconcilor_factory.h",
-    "signin/account_tracker_service_factory.cc",
-    "signin/account_tracker_service_factory.h",
-    "signin/chrome_signin_client.cc",
-    "signin/chrome_signin_client.h",
-    "signin/chrome_signin_client_factory.cc",
-    "signin/chrome_signin_client_factory.h",
-    "signin/chrome_signin_helper.cc",
-    "signin/chrome_signin_helper.h",
-    "signin/gaia_cookie_manager_service_factory.cc",
-    "signin/gaia_cookie_manager_service_factory.h",
-    "signin/identity_manager_factory.cc",
-    "signin/identity_manager_factory.h",
     "signin/investigator_dependency_provider.cc",
     "signin/investigator_dependency_provider.h",
[...]

PR #553, opened 2018-10-11
Update for 70

xsmile:
### Patches
- removed `patches/iridium-browser/safe_browsing-support-trk-prefix.patch`:
The safebrowsing protocol was updated from v3 to v4 and the relevant code was likely moved to `components/safe_browsing/db/`

### Features
- disabled `prefs::kSigninAllowedOnNextStartup` by deault. The setting can be found at  `chrome://settings`:
"Allow Chromium sign-in - By turning this off, you can sign in to Google sites like Gmail without signing in to Chromium"


Eloston:
> removed patches/iridium-browser/safe_browsing-support-trk-prefix.patch:

I don't think we need to preserve this patch; we don't support Safe Browsing.

> disabled prefs::kSigninAllowedOnNextStartup by deault. The setting can be found at chrome://settings:

Is this related to the Sync functionality (such as https://github.com/Eloston/ungoogled-chromium/issues/535#issuecomment-425679921 and #520)? I'm wondering how it's related. Also, does this flag still do anything even though the signin code is partially disabled?

xsmile:
@Eloston It is definitely related but I didn't have the chance to investigate yet. I assume that it won't do anything since users cannot sign in into Chromium. In any case, disabling the preference should be the better option.

Eloston:
I see, thanks. It definitely won't hurt to have it for now.

Eloston:
I'm planning to merge after 70 becomes stable, unless you think there's a reason to merge earlier. I'll ask if you're ready to merge when the time comes.

xsmile:
Alright, I will most likely refresh the patches and add the lists once the stable build is published.

csagan5:
I had tried myself to port the safebrowsing-eviction patch to v70, didn't manage; glad you have a working version here.

I have some additional changes for Android (the v69 addendum is here: https://github.com/bromite/bromite/blob/master/patches/BRM080_Fixes-for-building-without-safebrowsing-on-Android.patch), I will publish the new version adapted for v70. Not sure you should include it now if you're not building for Android (might quickly go stale) but I thought to mention it here for other people trying to build for Android without safe browsing.

xsmile:
I refreshed the existing and updated patches for Debian which include an important change:
The path to the master_preferences file was changed from `/usr/share/chromium/master_preferences` to `/etc/chromium/master_preferences`.

@csagan5 I did a few Android builds to test compatibility with the ungoogled patchset but I don't think I will continue to do it. It's probably better to keep the patches separated for now.
Nevertheless, it's nice to see the safebrowsing patches included in Bromite.

Eloston:
@csagan5 I agree with @xsmile. It's better to exclude Android-specific patches until Android support gets added into ungoogled-chromium. I don't see any benefit of including those patches ahead of time.

@xsmile Thanks for updating the config. I've fixed the patch validator to attempt retries with a backoff.

csagan5:
@Eloston we all agree 

> Not sure you should include it now if you're not building for Android (might quickly go stale)

csagan5:
On a separate topic, is the canvas mitigation still effective? I got an issue report (https://github.com/bromite/bromite/issues/154) but not investigated yet.

xsmile:
@csagan5 No issue here. The canvas fingerprint changes each time.

csagan5:
The issue with canvas was due to an invalid automatic patch application with fuzziness, my bad; I have also updated the patch to use the optimisation of reduced rand() calls: https://github.com/bromite/bromite/blob/70.0.3538.71/patches/BRM053_Canvas-fingerprinting-mitigations-for-image-data-and-webGL.patch#L142

On another topic (sorry for going kinda-off-topic here): have you experienced issues at fetching the `webrtc` third-party dependency via git?
In v69 and v70 (and in the past as well) I cannot fetch the git commit specified in `DEPS` but rather I can get it via the web interface (https://chromium.googlesource.com/external/webrtc/).

@xsmile are you doing the same? Otherwise, how do you fetch webrtc via git? Just curious if for some reason I am not able to fetch it via git because the tree is not published via git but only via HTTP.

xsmile:
@csagan5 This is the workaround I used for the WebRTC repository for 69: https://github.com/xsmile/ungoogled-chromium/blob/android/packaging/android/build.sh.ungoogin#L35.
I'm not sure why depot_tools doesn't find the commit and I assumed this is not an issue if you stick to the original build instructions.

csagan5:
Thanks for the pointer! I use `gclient sync --upstream`, and in your script `--no-history --shallow` I guess have the same limitation/feature: they do not fetch extra refs, hence that's why it is needed to use the explicit fetch for the separate WebRTC tree.

PR #564, opened 2018-10-24
Update fix-libva1-compatibility.patch for 70.x

ian-moone:


Eloston:
I didn't realize there were patches between fix-libva1-compatibility.patch that needed to be refreshed. I took care of that.

PR #565, opened 2018-10-24
Disable 'use_gold' if 'use_lld' is true

ian-moone:
As `use_gold=true` will be redundant in this case.


Eloston:
If I'm understanding what you're saying, `use_gold=true` is already implied when `use_lld=true`. Could you explain what your choice of `use_gold=false` with `use_lld=true` does?

ian-moone:
> Could you explain what your choice of `use_gold=false` with `use_lld=true` does?

Oh, sorry for my poor explanation. Because it's not possible to use both linker at the same time. `lld` is the preferred linker to use with `is_clang`. The logic is defined in `build/config/compiler/compiler.gni`. 

```
declare_args() {
  # Set to true to use lld, the LLVM linker. This flag may be used on Windows,
  # Linux or Fuchsia.
  # TODO(pcc): Enable lld on more architectures on Linux. E.g. we probably need
  # to fix some of crbug.com/742655 to enable it on ARM.
  use_lld =
      is_clang &&
      (is_win || is_fuchsia || (use_thin_lto && target_os != "chromeos") ||
       (is_linux && current_cpu == "x64" && target_os != "chromeos") ||
       (is_android && (current_cpu != "arm" || arm_version >= 7) &&
        current_cpu != "mipsel" && current_cpu != "mips64el"))
}

declare_args() {
  # Whether to use the gold linker from binutils instead of lld or bfd.
  use_gold = !use_lld && !(is_chromecast && is_linux &&
                           (current_cpu == "arm" || current_cpu == "mipsel")) &&
             ((is_linux && (current_cpu == "x64" || current_cpu == "x86" ||
                            current_cpu == "arm" || current_cpu == "mipsel" ||
                            current_cpu == "mips64el")) ||
              (is_android && (current_cpu == "x86" || current_cpu == "x64" ||
                              current_cpu == "arm" || current_cpu == "arm64")))
}
```

We could even remove both lines, as long as `is_clang` is true.

Eloston:
I see, thanks for explaining.

> We could even remove both lines, as long as is_clang is true.

I like this idea more, since it reduces potential breakages due to our GN configuration. I'll take care of this.

Issue #362, opened 2018-03-21
Large memory leak in main process over several days


Eloston:
There seems to be a memory leak in the Debian 9 amd64 binaries I've built.

Current reports: On my machine, I have seen the memory footprint of the main process grow from about 170 MB to over 400 MB in the span of 3 days (with [my binaries here](https://github.com/Eloston/ungoogled-chromium-binaries/releases/tag/65.0.3325.162-1)). @ik9999 has reported growth of *over 2 GB* on the main process within 3-5 days for my 64.0.3282.186-1 Debian 9 amd64 binaries on Debian testing (buster).

Some potential avenues to investigate:
* Sift through the patches to see if anything there could be causing any problems.
* See if any of ungoogled-chromium's deriving projects, such as Inox or Debian's Chromium, have memory leaks. This can narrow down the potentially problematic patches.
* See if `linux_portable` (as of version 65) is affected by memory leaks to this degree.
* Use [LeakSanitizer](https://www.chromium.org/developers/testing/leaksanitizer) based on `linux_portable` (as of version 65).

ik9999:
Are you sure that its not a [chromium](https://github.com/chromium/chromium) issue? I can't check it right now.

Eloston:
@ik9999 It's hard to say. Given that this is a pretty large leak and that you've been affected by it in previous versions, I think it's more likely that it's due to some change in ungoogled-chromium.

Eloston:
@ik9999 67 is out. Are you still experiencing major memory leaks?

ik9999:
@Eloston I feel like memory usage grows slower now, it was 721mb yesterday, now its 889mb.

tonowoe:
I'm using those pre-built binaries as well with Debian 9 amd64, and I've noticed the same. If I don't reboot my browser or computer for couple of days, I see that the browser is being heavy on RAM usage. It haven't bothered be that much because it's really fast to just restart the browser so the RAM usage will go back to normal again. I have had this problem for many versions now, and it's about the same in the latest 67 version. But it seems that RAM usage will not grow as fast as in earlier versions.

If I remember correct, this happened in the normal chromium too, not just in ungoogled, not sure though.

csagan5:
I have had at least one [similar report](https://github.com/bromite/bromite/issues/111); my suggestion is to look upstream for possible Chromium issues. Difficult to say whether it's the same root cause or not, since Bromite builds for Android and there are a lot of differences also in the memory model.

[...]

PR #591, opened 2018-11-18
Portability on Windows

### LINKED ISSUE #444: Portability under Windows OS
I use Chrome as a portable application under WIndows OS. In the past Chrome was partially portable. If chrome browser was moved to another system, extensions were kept. Although, some settings were reset, it was good enough. 

Version 55 of the ungoogled chrome was portable. However, after version 60 or so Chrome became unportable. When chrome is carried to another system, chrome deletes the extensions. I think google is checking file "Secure Preferences" for system changes, but I am not sure.

There are a few chrome based browsers that are portable such as Opera, maybe that would give an idea what needs to be changed.

I like ungoogled chrome, but because of portability, I am stuck at version 55. I am not able to use the latest builds.

If someone can look at what needs to be done, I will be happy.  

Thanks.

crazy-max:
Closes #444

## About

The purpose of this PR is to allow the portability of Chromium on Windows with the addition of two new switches :

* `disable-machine-id` : Allows disabling the machine ID generation on Windows.
* `disable-encryption` : Allows disabling encryption of cookies, passwords, settings. Should be ONLY used if your hard drive is encrypted or if you know what you are doing :upside_down_face:

This allows the portability of everything, except certificates.

Related threads :

* portapps/brave-portable#4
* brave/brave-browser#694
* brave/brave-core#795

### disable-machine-id

* Launch chormium with switch `--disable-machine-id`
* Install an extension
* Close chormium and copy data to a second computer
* Launch chormium on the second computer
* The extension must be ok

### disable-encryption

* Launch chormium with switch `--disable-encryption`
* Login to a known website with credentials and save them (will add unencrypted cookies and passwords to database)
* Close chormium and copy data to a second computer
* Launch chormium on the second computer
* You must be still logged in on the website and credentials must be kept intact in the database

### use both switches

* Previous and following settings must be ok too :
  * Default search engine
  * Homepage
  * Startup pages
  * Pinned tabs

### DIFF
--- a/patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
+++ b/patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
@@ -1,0 +1,53 @@
+# Disable machine ID generation on Windows.
+# See https://github.com/Eloston/ungoogled-chromium/issues/444
+
+--- a/services/preferences/tracked/device_id_win.cc
++++ b/services/preferences/tracked/device_id_win.cc
+@@ -10,11 +10,15 @@
+
+ #include <memory>
+
++#include "base/command_line.h"
+ #include "base/logging.h"
+ #include "base/macros.h"
+
+ MachineIdStatus GetDeterministicMachineSpecificId(std::string* machine_id) {
+   DCHECK(machine_id);
++  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-machine-id")) {
++    return MachineIdStatus::NOT_IMPLEMENTED;
++  }
+
+   wchar_t computer_name[MAX_COMPUTERNAME_LENGTH + 1] = {};
+   DWORD computer_name_size = arraysize(computer_name);
+--- a/components/metrics/machine_id_provider_win.cc
++++ b/components/metrics/machine_id_provider_win.cc
+@@ -9,6 +9,7 @@
+ #include <winioctl.h>
+ 
+ #include "base/base_paths.h"
++#include "base/command_line.h"
+ #include "base/files/file_path.h"
+ #include "base/path_service.h"
+ #include "base/threading/scoped_blocking_call.h"
+@@ -18,6 +19,9 @@ namespace metrics {
+ 
+ // static
+ bool MachineIdProvider::HasId() {
++  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-machine-id")) {
++    return false;
++  }
+   return true;
+ }
+ 
+@@ -25,6 +29,11 @@ bool MachineIdProvider::HasId() {
+ // is running from.
+ // static
+ std::string MachineIdProvider::GetMachineId() {
++  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-machine-id")) {
++    NOTREACHED();
++    return std::string();
++  }
++
+   base::ScopedBlockingCall scoped_blocking_call(base::BlockingType::MAY_BLOCK);
+ 
+   // Use the program's path to get the drive used for the machine id. This means
--- a/patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
+++ b/patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
@@ -39,12 +39,11 @@
    return true;
  }
  
-@@ -25,6 +29,11 @@ bool MachineIdProvider::HasId() {
+@@ -25,6 +29,10 @@ bool MachineIdProvider::HasId() {
  // is running from.
  // static
  std::string MachineIdProvider::GetMachineId() {
 +  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-machine-id")) {
-+    NOTREACHED();
 +    return std::string();
 +  }
 +
--- a/config_bundles/windows/patch_order.list
+++ b/config_bundles/windows/patch_order.list
@@ -6,3 +6,5 @@ ungoogled-chromium/windows/windows-disable-rcpy.patch
 ungoogled-chromium/windows/windows-fix-building-without-safebrowsing.patch
 ungoogled-chromium/windows/windows-fix-enum-conflict.patch
 ungoogled-chromium/windows/windows-fix-building-gn.patch
+ungoogled-chromium/windows/windows-disable-encryption.patch
+ungoogled-chromium/windows/windows-disable-machine-id.patch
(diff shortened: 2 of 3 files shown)


patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
```diff
@@ -0,0 +1,53 @@
...
++  }
+   return true;
+ }
+ 
+@@ -25,6 +29,11 @@ bool MachineIdProvider::HasId() {
+ // is running from.
+ // static
+ std::string MachineIdProvider::GetMachineId() {
++  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-machine-id")) {
++    NOTREACHED();
```
Eloston: What is the purpose of `NOTREACHED()` here? Does `disable-machine-id` break debug builds or something?

Eloston:
Looking good, but a couple questions:

1. Why is this marked as a WIP? What else needs to be done?
2. What happens if you disable machine ID but leave encryption enabled? Is encryption still tied to the current machine? Are the machine ID and encryption mechanisms independent of each other?

crazy-max:
@Eloston

1. WIP to handle certificates but I think we cannot do something about it so feel free to remove the WIP :)
2. For me encryption is not tied to machine-id. Machine ID seems to be linked to extensions and only used on Windows.

patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
Eloston: The purpose of `NOTREACHED()` is to raise an assertion (i.e. crash the browser) when the code is reached. So, `NOTREACHED()` in this context doesn't make sense. The reason why this works is because `NOTREACHED()` is implemented as `DCHECK(false)`, which only triggers in debug builds; ungoogled-chromium is built in release mode by default. [See the documentation here for more info](https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md#check_dcheck_and-notreached). (Sidenote: The fact that Linux and macOS have `NOTREACHED()` further confirms that these platforms aren't supposed to use the machine ID mechanism.)

I don't see anything other troublesome code here, but please check that the code immediately affected by your changes makes sense in the future. Otherwise, we'd be accumulating technical debt that can be challenging to solve (e.g. #362).

Eloston:
BTW on Windows, does Chromium store credentials internally in the user data directory by default? Or is there a credential storage mechanism that Chromium uses? I'm trying to understand why passwords are included in the encryption.

crazy-max:
And Chromium uses [CryptProtectData](https://docs.microsoft.com/en-us/windows/desktop/api/dpapi/nf-dpapi-cryptprotectdata) to encrypt passwords in the `password_value` BLOB.

patches/ungoogled-chromium/windows/windows-disable-machine-id.patch
crazy-max: You're right, I should have asked about this behavior. But indeed this stub allows to test the code without dealing with this dependency directly.

Eloston:
I see, it kinda sucks that the Windows version doesn't use a credentials storage. Thanks for the info.

caspertone2003:
@crazy-max 
@Eloston 
Thanks!
Good job, much desired behaviour.

Jason4774:
请问这两个参数现在还支持最新的chrome吗？

GeographicCone:
Thank you for this patch. Works great with 76.0.3809.100.

crazy-max:
Hi @Eloston,

Looks like the patches for portability on Windows have been removed?

Eloston:
@crazy-max This was merged before the fragmentation refactor (i.e. before platform repos were created). They now live in here: https://github.com/ungoogled-software/ungoogled-chromium-windows/tree/master/patches/ungoogled-chromium/windows

crazy-max:
@Eloston Perfect thanks

gilluc:
Why appdata/local/chromium is not stored just beside chromium.exe to be really portable??
This could make it possible for me to use ungoogled chromium on my two pcs from my Nextcloud drive...

mikhoul:
@Eloston @crazy-max  I have a question I want to transfer my "user data" profile (from my  Windows desktop) to my Windows laptop without having to reinstall all my extensions  and configure their settings.

There is also unpacked customized extensions in dev mode that I want to transfer too (they are inside a specific sub-folder inside the "user data" folder.  Of course I want to keep all the custom settings and flags related to the browser.

My question is: **How I can transfer all my extensions (packed/unpacked with their settings) + the browser/flags settings with the `disable-machine-id/disable-encryption` switch ???   🤔

I'm not sure of the right procedure since my extensions were installed without the "`disable-machine-id"` switch

Regards :octocat:

iGom:
Do we also lose history, settings, extensions and bookmarks even when we update/upgrade ungoogled-chromium on the same machine on Windows??? Because that's what happened to me right now.

MarvinFS:
I have been using command line switches **disable-machine-id and disable-encryption** for ages now, all was good, but I noticed on the main project website these 2 parameters are now crossed out, but I can't find any details what happened. Also with the latest build I noticed all my cookies doesn't survive browsers restart and I'm forced to login to absolutely all sites from scratch each time, which is a huge issue for me.  I initially thought that is related to 3PCD, but doesn't seems like, as I manually marked all my required websites as allowed for 3rd party cookies still if I restart the browser I'm forced to login again everywhere absolutely. I was never the case and something changed only in the latest build 122.xxx, but I'm unable to find what it was exactly. 
image

mavorse:
Do these switches still work with the current version of Chromium? I tried using them on Vivaldi, and they didn't seem to function.

tomasz1986:
> @mavorse Apparently it's `--disable-encryption-win` now, see [macchrome/winchrome#135](https://github.com/macchrome/winchrome/issues/135)

According to https://github.com/ungoogled-software/ungoogled-chromium-windows/blob/master/patches/ungoogled-chromium/windows/windows-disable-encryption.patch, the switch should still be just `--disable-encryption-win`. Please keep in mind that the binaries from are not official and may have additional modifications in comparison to the official Windows binaries distributed by https://github.com/ungoogled-software.

Nifury:
Well, passing `--disable-encryption-win` to chromium is the same as passing nothing. It's just an invalid argument. 
`disable-encryption-win` doesn't even exist in either UG or winchrome. 
image

AJolly:
Fyi if you reaaally want its possible to export dpapi keys and also resign chrome/chromiums securepreferences.

AJolly:
Also the issue with the current patches is if you ever accidentally run the exe without having those cli flags set, it'll wipe out the existing settings.

PF4Public:
@AJolly Please file proper issues into Windows repository if you think those issues are a significant ones.

PR #627, opened 2018-12-13
Fix macOS build

tie:
This PR should fix build on macOS systems (still waiting for the build to finish).

The following issues I've encountered before running `ninja` are addressed:

1. Missing `--build-path` argument — `bootstrap.py` generated `out/Release` by default.

2. Since safe browsing is disabled, `disk_image_type_sniffer_mac` is never defined and should not be added to `deps` in that case.


Eloston:
> Missing --build-path argument — bootstrap.py generated out/Release by default.

Could you elaborate? What is being generated in `out/Release`?

Eloston:
I see, it doesn't really matter then. We can leave it alone. Thanks for the explanation.

PR #628, opened 2018-12-14
Added implicit https scheme patch

tie:
All URLs entered in omnibox without explicit scheme will have `https://` prefix instead of `http://`.  That is, HTTPS by default without any extension.
This feature is not configurable because plain-text requests directly affect user's privacy.

Also, it could be my imagination, but looks like Chromium loads web pages significantly faster with this patch.

As shown below, some sites do not redirect to HTTPS by default.

__Before__

vanilla

__After__

patched

### Alternatives

- [https-by-default](https://github.com/Rob--W/https-by-default) uses similar approach for Chromium, but patches omnibox autocomplete instead of the URL fixer.

- [HTTPS Everywhere](https://www.eff.org/https-everywhere), etc.

  - Hard to setup if (e.g. in my case Chromium runs from tmpfs).
  - Third-party code (which probably has performance issues) running in the browser.
  - Usually just a database of sites known to support HTTPS.

- Entering URL with explicit scheme manually.

  Inconvenient and error-prone.

### DIFF
--- a/config_bundles/common/patch_order.list
+++ b/config_bundles/common/patch_order.list
@@ -117,6 +117,7 @@ ungoogled-chromium/add-flag-to-show-avatar-button.patch
 ungoogled-chromium/no-such-option-no-sysroot.patch
 ungoogled-chromium/add-suggestions-url-field.patch
 ungoogled-chromium/add-flag-to-hide-crashed-bubble.patch
+ungoogled-chromium/default-to-https-scheme.patch
 
 bromite/fingerprinting-flags-client-rects-and-measuretext.patch
 bromite/flag-max-connections-per-host.patch
--- a/patches/ungoogled-chromium/default-to-https-scheme.patch
+++ b/patches/ungoogled-chromium/default-to-https-scheme.patch
@@ -1,6 +1,7 @@
 # Default to https for (non-standard) URLs without scheme.
 #
 # This patch handles URLs like user:pass@example.com and trims scheme in suggestions if it is implicit.
+# URLs entered in md_bookmarks (chrome://bookmarks) will also get https:// prefix.
 #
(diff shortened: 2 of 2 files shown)


tie:
I've fixed suggestions in the last commit.  Scheme is trimmed in suggestion dropdown for URLs without explicit scheme.

tie:
>However, will this patch affect other ways of loading web pages?

Thanks for noticing that.

I've looked through the [users of FixupURL](https://github.com/chromium/chromium/search?q=fixup+url+http) on GitHub chromium mirror and, except omnibox (which is already covered by this patch), only [searchbox.cc](https://github.com/chromium/chromium/blob/59a4f503269b8ca4bee6df573770c658e8c7f46e/chrome/renderer/searchbox/searchbox.cc#L175-L182) depends on old FixupURL behavior.

Also, `url_formatter::FixupURL` is used in _most_ of the user URL input parsing, so this should affect command-line usage too.

I'll let you when I come home because I don't have the build right now.

saltama:
As  long time user of https everywhere, great patch, thanks.

tie:
With the last commit URLs without explicit (i.e. invalid URLs) scheme entered in chrome://bookmarks will be prefixed with `https://` and parsed again.

Command line usage works as expected — `chromium example.com` opens a new tab for `https://example.com` URL.

Eloston:
Awesome. Thanks for looking into it.

intika:
Nice patch :+1: but should not this added with a flag or or something there is a lot site using only http, i know that most of them will be opened after searching a web-engine... any way just my 2 cents not really important

Eloston:
@intika That isn't affected by this patch. Only URLs without a scheme are considered.

PR #638, opened 2018-12-18
Changes in Archlinux building script (PKGBUILD)

seppiofish:
This pull request is actually not complete since the harfbuzz.patch file should be put in the same directory of the PKGBUILD by buildkit scripts for the build process to work.

I would therefore like to discuss whether sould be created an ad hoc directory to put PKGBUILD and patch in, or simply copy the .patch file in the same folder where PKGBUILD is generated.

### DIFF
--- a/config_bundles/archlinux/patch_order.list
+++ b/config_bundles/archlinux/patch_order.list
@@ -1,0 +1,1 @@
+debian_buster/system/harfbuzz.patch
--- a/packaging/archlinux/PKGBUILD.ungoogin
+++ b/packaging/archlinux/PKGBUILD.ungoogin
@@ -16,7 +16,7 @@ url="https://github.com/Eloston/ungoogled-chromium"
 license=('BSD')
 depends=('gtk3' 'nss' 'alsa-lib' 'xdg-utils' 'libxss' 'libcups' 'libgcrypt'
          'ttf-font' 'systemd' 'dbus' 'libpulse' 'pciutils' 'json-glib'
-         'desktop-file-utils' 'hicolor-icon-theme')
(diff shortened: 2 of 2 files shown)


xsmile:
The harfbuzz patch is already present in `patches/debian_buster/system` and can be used by listing it in `config_bundles/archlinux/patch_order.list`. ungoogled' build scripts will take care of the rest.

ian-moone:
Instead of `config_bundles/archlinux/patch_order.list`, I would like to request to be added to `config_bundles/linux_rooted/patch_order.list`. So downstream _gentooers_ can be benefited too. For the time being I'm [sed-fixing](https://gitlab.com/chaoslab/chaoslab-overlay/commit/b98a58050ddd56a69024e9e95c1f4ad7040f0d3f#10ee16d7a5fecdaf39c6d9e342880f1ffc2d44ab_283_282) it.

seppiofish:
If I am not mistaken archlinux config bundle depends on linux_rooted bundle, so adding it there shoud do. But will it not be a problem for systems that don't yet have harfbuzz 2 and build with linux_rooted bundle to have chromium patched that way? I truly don't know so I'm genuinely asking.

Eloston:
@seppiofish is correct. Adding the harfbuzz patch will break Debian stretch, since its harfbuzz library is too old. Thus, I think adding the harfbuzz patch to Arch Linux's patch_order is the best solution here.

Also, I would appreciate it if you all could give some input to #640. Thanks!

seppiofish:
ok that was unintended... I must have done something wrong. Hope reopening the PR fixes this. Anyway I deleted the previous two commits since one was unnecessary and the other wrong (I recived an advice from an AUR user and basically jsoncpp is an actual dependecy and not only a build one), and now committed the correct changes.

PR #649, opened 2018-12-29
Add use_sysroot to portable version

intika:
fix https://github.com/Eloston/ungoogled-chromium/issues/556

### DIFF
--- a/config_bundles/linux_portable/gn_flags.map
+++ b/config_bundles/linux_portable/gn_flags.map
@@ -3,3 +3,4 @@ host_toolchain="//build/toolchain/linux/unbundle:default"
 linux_use_bundled_binutils=false
 optimize_for_size=false
 use_kerberos=false
+use_sysroot=true
--- a/packaging/linux_simple/build.sh
+++ b/packaging/linux_simple/build.sh
@@ -18,6 +18,7 @@ mkdir $DOWNLOAD_CACHE || true
 python3 -m buildkit downloads retrieve -b config_bundles/$BUNDLE -c $DOWNLOAD_CACHE
 python3 -m buildkit downloads unpack -b config_bundles/$BUNDLE -c $DOWNLOAD_CACHE ../
 python3 -m buildkit prune -b config_bundles/$BUNDLE ../
+../build/linux/sysroot_scripts/install-sysroot.py --arch=amd64
 python3 -m buildkit patches apply -b config_bundles/$BUNDLE ../
 python3 -m buildkit domains apply -b config_bundles/$BUNDLE -c domainsubcache.tar.gz ../
 python3 -m buildkit gnargs print -b config_bundles/$BUNDLE > ../out/Default/args.gn
--- a/packaging/linux_simple/ungoogled-chromium.yml
+++ b/packaging/linux_simple/ungoogled-chromium.yml
@@ -9,7 +9,9 @@
 app: ungoogled-chromium
 
 ingredients:
-  package: libgconf2-4
+  package: 
+    - libgconf2-4
+    - libatomic1
   dist: trusty
   sources: 
     - deb http://us.archive.ubuntu.com/ubuntu/ trusty main universe


Eloston:
If I'm not mistaken, `install-sysroot.py` will download binaries from Google to work. That doesn't make sense since this project is trying to remove Google integration. Could you look into using the script to create the sysroot? If possible, I'd like to cache the created sysroot instead of re-creating it every time.

intika:
Yes it is grabbing it from google i thought about that too i guess its why sysroot is removed in the first place... i am investigating that and will report back

intika:
To install the sysroot `install-sysroot.py` is used and to "build" them `sysroot-creator-sid.sh BuildSysrootAmd64` is used **BUT** in both cases the binaries are downloaded 

either from 
https://commondatastorage.googleapis.com/chrome-linux-sysroot/toolchain/e7c53f04bd88d29d075bfd1f62b073aeb69cbe09/debian_sid_amd64_sysroot.tar.xz

or from
http://ftp.us.debian.org/debian//dists/experimental/main/binary-amd64/Packages.xz

I know its possible to build all those package from sources etc. but it a huge work what do you think ? should be stop here or go into a gentoo venture ? loool :p just kidding. the right thing to do is to create a script to build all the needed package from sources, and to be honest i don't have time to do it not right now. 

So decide and let me know

intika:
or may be provide both binaries with and without sysroot

Eloston:
The Debian binaries are fine. The main thing is that they aren't from Google. For example, on Windows and macOS, we download the LLVM toolchain from LLVM.

intika:
> The Debian binaries are fine. The main thing is that they aren't from Google. For example, on Windows and macOS, we download the LLVM toolchain from LLVM.

looooooool it's truly about ungoogling the stuff hahahah ok, i will see how to implement that because i am building on suse and the main method give some errors if not on debian...

intika:
i am facing a little problem the command to build is `./build/linux/sysroot_scripts/sysroot-creator-sid.sh BuildSysrootAmd64` and as i mentioned i am building on suse... and finally this can not be done on suse nor on non debian based system the reason i am building one suse is because of the glibc version v2.22 for a good backward compatibility. 

possible solution host a built pre-built sysroot from debian on uc repo and adapt the command to grab that file instead on the one coming from google. otherwise this will break compatibility with non debian system based on the portable build

what should we do ?

Eloston:
Maybe instead of sysroots, we can require people to build ungoogled-chromium inside a container or chroot of Debian stable?

intika:
To fix the issue i was having with non debian system i started by adding manually the Debian signature keys... (it's a single command like to fix that)... this fixed the first step of the process, then i was asked dpkg-deb (or something similar)... i stopped there... i think if i go further in fixing the compatibility i will may be end up having 5 or 6 additional commands... i don't think we should break compatibility. 
may be going further in debugging and fixing the compatibility on non Debian system would be the way to go (this is what is mostly done on the different build wiki over here)
This would be faster and more reliable than a container or a chroot.
What do you think ? 

(i will see look into that later on, i am a little out of time lool)
Note : i became a github junkie loooool

intika:
@Off-topic by the way are you going to the FOSDEM event in Brussels ?

Eloston:
> i think if i go further in fixing the compatibility i will may be end up having 5 or 6 additional commands... i don't think we should break compatibility.
> This would be faster and more reliable than a container or a chroot.

The sysroot creation script is not under our control. I think this will be more work for us to maintain if Google changes it. This is why a chroot or a container of Debian stable are more practical options, since we are in total control of them.

> @Off-topic by the way are you going to the FOSDEM event in Brussels ?

Nope.

intika:
i reevaluated the situation here, i may just end up building on debian... i am summarizing 

**Needs:** 
use_sysroot to increase systems compatibility

**Issue:**
use_sysroot is using a pre compiled binary hosted on google servers

**Note:**
chromium build tools are using debian sysroot other distro are not implemented 
 
**Solution:**
build/gather sysroot from debian repo **>** this require a debian system/chroot/container

**Solution implementation 1:**
Limit portable **building** integrating **sysroot** to debian (this somehow what google is doing by its unique debian sysroot) user who want to build on other system would just build without sysroot (as generally most of the time the system we build in are the target system for the application to be run on)
This solution is the easiest one 
- Pro: 
-- Easy to implement  
-- Almost zero maintenance 
-- User who want to build on system other than debian can do so while disabling the sysroot feature (this is not a problem as long as the targeted distro is the same)
-- Users building on system A (other than debian) and targeting system B are a very small user base
- Con:  
-- Portable build with sysroot feature (not including google binary) will be limited to debian
-- Users who wants to build on system A (other than debian) and target system B, can encounter https://github.com/Eloston/ungoogled-chromium/issues/556 
 
**Solution implementation 2:** 
Debian chroot/container:
- Pro: 
-- Portable release would be build-able on any system and targeting any system
-- That chroot/container could be hosted and maintained on ug git that would be kind a VM for building portable uc that devs would just use without facing compatibility/missing-package issues  
- Con:  
-- Not easy to implement 
-- All the build environment needs to be integrated/added/installed on that container 
-- Need considerable maintenance in case it's hosted 
-- If it's implemented as a script/instruction to install/add a chroot/container that would make building uc super complicated and it is already the case i don't think we need to supercharge the building system
-- Space needed to build 
-- Time needed to build 
-- Do-ability regarding noobs 
-- Issue generator for unqualified users and thus wasting time helping them.    

**Solution implementation 3:** 
Hosting a debian prebuild sysroot on a separate git
Basically doing the following on a debian system and uploading the result at every release.
```
$ cd build/linux/sysroot_scripts
$ ./sysroot-creator-stretch.sh BuildSysrootAll
```
-Pro:
-- Super easy to do
-- Solve all the headache
-- No change to the current build system 
-- All distro will remain compatible without changes
-- Implementation would be super easy as well just changing the url from the google server to the generated file
-Con:
-- Sysroot need to be generated from time to time or on every release

**Solution implementation 4:** 
implement a container just to build the sysroot not the whole chromium thing... this is kind a solution 2 with less con. 

**Alternative:** 
Google is claiming their binary are coming from debian repo https://chromium.googlesource.com/chromium/src/+/HEAD/docs/linux_sysroot.md
Using the binary from the google hosting may be a solution 

**Personal situation**
I have limited time for UG i personally can not go in a road that require more maintenance time than the current situation, i will then close this PR as current solution is not suitable, temporarily i will build next releases on debian system to be able to add sysroot (not the one from google servers) **OR** i will build that sysroot on a debian system and use it in my current build system (opensuse) this would be the easier way for me, if this solution is suitable long term, i may maintain a build script for sysroot and PR it. That script would be used as solution 3... 

Any way let me know what you decide and i will see if i can do something about it ;) otherwise i am not alone contributing for portable ;) others may help in case of the container solution.

intika:
> I don't understand what this solution is. Are we requiring people to build on Debian machines (no chroot/container)?

When we **add** sysroot using google tools/script it download sysroot from google and thus this works on any system BUT if we want to build/package it `./sysroot-creator-stretch.sh BuildSysrootAll` this require a debian system and can not be run on any system 

> I don't understand this solution. Why does using a container to build the sysroot help?

If we want to build the sysroot with the current build script it require debian, now we can implement a new script that build sysroot on any system but it would require a lot of maintenance... 
So the theoretical idea here was to use a debian container to build sysroot while keeping the current build method/script

> They also have hashes for the entire sysroot images (in tar files), so maybe we could have some way of verifying these hashes are valid by building the images ourselves too? I'm not sure if the tar file is reproducible though...

Yeah tar would produce a different hash, what we could do instead is verify/check that google is indeed using debian binary (this require same effort as solution 3) we would decompress the tar and check the hash of all file inside it (comparing them to debian build) this could get complicated if there are version miss-match between google's hosted binary and debian binary... (i mean miss-match between latest and used version)

> I like this solution the most because there are commands to build a sysroot image relatively easily (therefore, any user can build it themselves if they choose). But, how do we make Chromium use the sysroot image?

Me too... to make Chromium use the sysroot image is relatively super easy, current patching system is removing the google link (link of the google's hosted sysroot) we would just replace that link with our link. (would just require only one additional line on the patch that remove google's links from sources)

I can implement solution 3, would be relatively easy (may be one day or half day) to get everything done... we would need an additional sysroot git... if you want to go for it just create an other git for the hosted sysroot under ungoogled-chromium organization, i will PR to it later on once implemented. 

So solution would require: 
- a git/repo for hosting the script to build sysroot and the built sysroot
- a PR to UC similar to the current one (with one line difference or so...)

Eloston:
I see, I didn't realize that building the sysroot image requires Debian. In that case, my view is different.

There are essentially two groups of users with their own trade-offs we need to consider for controlled build environments (e.g. sysroot, chroot, container, VM, dedicated system):

1. **Those that only build ungoogled-chromium for their own Linux systems**. Technically, these users don't need to worry about a controlled build environment as long as they run their binary on a system that is compatible with theirs.
    * *Not using a controlled environment*:  However, there still can be issues if their system libraries are too different from Chromium's expected library versions, or they end up running the binary on a system that is no longer compatible (e.g. by switching distros, or maybe package updates). Even then, I'd expect the number of people in this group to have issues to be relatively small.
    * *Using a controlled environment*: Using a controlled environment will eliminate the problems from above, but it is more costly (time and space) for the user. The methods for creating controlled build environments are described in the next group.
2. **Those that build ungoogled-chromium for use on all\* Linux systems**. Backwards-compatibilty with the system libraries is important to ensure runtime compatibility. This group will want to have a binary they can readily download and run anywhere (e.g. from Contributor Binaries). Therefore, binaries for this group **must** be built in a controlled environment.
    * *Using sysroots*: Since this must be built in a Debian environment, it is silly to require the users to make a container just to build the sysroot. Instead, you mentioned that we could provide pre-built sysroot images. However, this will be more for us to maintain (even if we automate the building and publishing of sysroot images via CI). Also, we'd need to patch Google's script to use our own sysroot image, which means we will need to keep track of changes Google makes to this script.
    * *Using other methods*: This removes all complications involving sysroots. In fact, with a solution like Docker, we could provide a `Dockerfile` that could automate the entire build for us. Sysroots don't provide us that kind of power. However, I'm not sure what kinds of downsides we would have using something like Docker.

\* That is, distros supported by Chromium

Essentially, we need to answer these two questions:

* For Group 1, should we enforce a controlled build environment?
* To make a controlled build environment, should we use sysroots or another method like Docker?

intika:
> you mentioned that we could provide pre-built sysroot images. However, this will be more for us to maintain (even if we automate the building and publishing of sysroot images via CI). Also, we'd need to patch Google's script to use our own sysroot image, which means we will need to keep track of changes Google makes to this script.

Yes but it does not require a lot of time 

> - For Group 1, should we enforce a controlled build environment?

Theoretically that group does not need sysroot the build script could be adapted to add a "local portable version". Enforcing a controlled build environment will depend on how we implement that controlled build environment if that would require installing docker and setting it up it may be a downside (additional heavy steps) may be splitting the build system could be a solution here 
- Build a portable widely compatible with a controlled build environment
- Build a portable for local usage

If the controlled build environment solution is fairly easy to implement we would just enforce it and avoid all the headache  

> - To make a controlled build environment, should we use sysroots or another method like Docker?

i know how docker works but am not familiar with it (not using it) do you have the time to implement this ? basically what would that `Dockerfile` do ? the container would be a debian system ? in short what's required for the docker solution ? 

Just pick a solution that works best according to what's required, keep in mind some important things

- Over complicating a build system that is already complicated may not help 
- The solution would need to be easily maintainable by you or any contributor (what i mean is that this should not be a time consuming solution otherwise we would remove the portable and add more distributions support)

Eloston:
Here's an example I just found for building Chromium in Docker: https://github.com/chromedp/docker-chromium-builder

I have worked with Docker before, so I could put something together relatively easily. However, this requires users to install Docker in order to use it. This is more work for the user compared to the sysroot solution.

While I don't like that we would have to maintain patches for the sysroot scripts, I think it's better than Docker for users who want to make builds. I will re-open this PR and create a new repo under ungoogled-software for you to PR against.

Eloston:
I created [ungoogled-chromium-sysroot](https://github.com/ungoogled-software/ungoogled-chromium-sysroot). I'm not sure if you can PR against an empty repo, so let me know if it doesn't work.

Also I should note that Docker will be our fallback if sysroots don't work as well as we think they will.

Eloston:
Closing since the fragmentation refactor has been merged. Please re-submit when ungoogled-chromium-portablelinux has been updated.

PR #650, opened 2018-12-30
Remove "optimize_for_size" redundancy

ian-moone:
`build/config/compiler/compiler.gni` has the following entry:

```
# If true, optimize for size. Does not affect windows builds.
# Linux & Mac favor speed over size.
# TODO(brettw) it's weird that Mac and desktop Linux are different. We should
# explore favoring size over speed in this case as well.
optimize_for_size = is_android || is_ios || is_fuchsia
```


ian-moone:
> Is there fuzz in debian_buster/fixes/as-needed.patch now?

I'll refresh the patches.

ian-moone:
All systems go!

ian-moone:
For historical purposes, upstream has left the following entry:

```
# If true, optimize for size. Does not affect windows builds.
# Linux & Mac favor speed over size.
# TODO(brettw) it's weird that Mac and desktop Linux are different. We should
# explore favoring size over speed in this case as well.
optimize_for_size = is_android || is_ios || is_fuchsia
```

PR #664, opened 2019-01-31
Update for 72.0.3626.81

xsmile:
- disabled GN flag `closure_compile` to avoid the use of Java and of a pre-shipped jar file
- replaced VA-API patch with https://src.fedoraproject.org/rpms/chromium/pull-request/18 as the refreshed version of the old patch didn't enable acceleration

@Eloston While I didn't follow the workflow for Debian-related patches a suggested in the previous PR, I hope I didn't forget anything this time.


xsmile:
Alright, the remaining errors seem to be related to the other bundles I didn't update.

Eloston:
Works for me. Thanks.

PR #674, opened 2019-02-11
Update building.md

pouledodue:
General knowledge for newbie builders https://github.com/Eloston/ungoogled-chromium/issues/672#issuecomment-462431896


Eloston:
Near the top of the building document, it indicates the following:

> **Choosing a version**: *It is highly recommended to choose a tag version for building.* `master` and other branches are not guaranteed to be in a working state.

Therefore, this note is irrelevant. If there's anything that should be changed, it's the section at the top to make this more obvious to new builders.

(Also, trying to help new developers use unstable code is a recipe for disaster).

PR #685, opened 2019-02-20
 Improved macOS building documentation

TCB13:
Updates macOS building documentation with a more detailed setup guide.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -176,13 +176,13 @@ A zip archive will be created in `build\src`
 5. Install Python 3.x: `brew install python`
 6. Install Python's pyenv to manage python version: `brew install pyenv`
 7. Install Python 2.7.13: `pyenv install 2.7.13`
-8. Set Terminal's `python` command to use Python 2.7.13 instead of 3.x. Edit/create `~/.bash_profile` and add:
-```sh
-alias python="~/.pyenv/versions/2.7.13/bin/python2.7"
-```
-9. Restart your Terminal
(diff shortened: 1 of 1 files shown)


docs/building.md
```diff
@@ -171,8 +171,18 @@ A zip archive will be created in `build\src`
...
 1. Install Ninja via Homebrew: `brew install ninja`
 2. Install GNU coreutils (for `greadlink` in packaging script): `brew install coreutils`
+3. Install GNU readline: `brew install readline`
+4. Install the data compression tools xz and zlib: `brew install xz zlib`
+5. Install Python 3.x: `brew install python`
+6. Install Python's pyenv to manage python version: `brew install pyenv`
+7. Install Python 2.7.13: `pyenv install 2.7.13`
+8. Set Terminal's `python` command to use Python 2.7.13 instead of 3.x. Edit/create `~/.bash_profile` and add:
+```sh
+alias python="~/.pyenv/versions/2.7.13/bin/python2.7"
```
Eloston: Does pyenv support creating a symlink to Python as `python`? I don't think bash aliases will work in all of the cases that use Python 2 (correct me if I'm wrong).

TCB13: @Eloston yeah we can do it with pyenv, but I initially did it via bash aliases and it complied just fine, multiple times. I'll check this later today again.

TCB13:
@Eloston I've updated the file as discussed.

Issue #675, opened 2019-02-11
BUILD.gn error during compilation on Windows


hardhub:
### Description

Error during patching:
```
patching file chrome/BUILD.gn
Hunk #2 FAILED at 83.
```
### Environment details

* ungoogled-chromium version: all tried versions starting 69.*
* Configuration bundle: windows
* Packaging type (if applicable): windows
* OS and version for building: Windows 7 x64
* Compiler toolchain: VS 2017

### Building and packaging steps

py ungoogled_packaging\build.py

TCB13:
Related to https://github.com/Eloston/ungoogled-chromium/issues/639

TCB13:
Maybe related to the safe browsing mess? https://github.com/Eloston/ungoogled-chromium/issues/639#issuecomment-461983120

I was stuck on this for a long time, manage to get around it on macOS but under windows looks like its harder...

TCB13:
Update: this isn't actually related to safe browser. Looks like a new issue of this version.

hardhub:
@TCB13 

I simply made all as specified in instructions.  
Maybe you used different version of chromium source.  
To be honest I do not know exactly.

TCB13:
@hardhub can you cofirm that your build process was like this?

1. Install:
- 7z1806-x64.msi
- python-2.7.15.amd64.msi (add it to path)
- python-3.6.5-amd64.exe (don't add to path; enable py launcher)
2. Run `pip install pypiwin32`
3. Install VSCode (vs_community__735157147.1549731593.exe) with the following components:
```
- Microsoft.VisualStudio.Workload.NativeDesktop
- microsoft.visualstudio.component.debugger.justintime
- microsoft.visualstudio.component.vc.diagnostictools
- microsoft.visualstudio.component.vc.cmake.project
- microsoft.visualstudio.component.vc.testadapterforboosttest
- microsoft.visualstudio.component.vc.testadapterforgoogletest
- microsoft.component.vc.runtime.ucrtsdk
- microsoft.visualstudio.component.vc.atlmfc
- microsoft.visualstudio.component.vc.cli.support
- microsoft.visualstudio.component.vc.140
- component.linux.cmake
```
4. Open `Settings > Apps & features > Windows SDK 10.0.17763.132 > Modify` and enable `Debugging Tools For Windows`
[...]

PR #705, opened 2019-03-15
[dev][windows] Refresh patches for 72.0.3626.122

r4sas:
Changes according to #675.

*Note:* dev tree only. This will fix patching only, but compiling still (maybe) need more changes.


r4sas:
> `windows-disable-reorder-fix-linking.patch` still doesn't apply cleanly according to the CI check, but I'll merge anyway to let you guys make progress on Windows.

I can create PR with fix of this fuzz'es if needed.

Issue #651, opened 2018-12-30
The "fragmentation" refactor


Eloston:
This is a refactor of ungoogled-chromium as a result of #640. Development will occur in the "fragmentation" branch and new branches of other repos. **The refactor will not affect regular development until it is ready to be used.**

Essentially, this refactor is the separation of platform-specific code into individual repositories under the [ungoogled-software organization](//github.com/ungoogled-software). It makes maintenance easier for myself and the contributors for each platform.

Rough TODO:

* [x] Refactor file structure
* [x] Separate patches into "core" and "extra"
* [x] Update README with new Motivation and Philosphy section to reflect what ungoogled-chromium has been effectively striving for.
* [x] Split up and trim `buildkit` into individual Python scripts under `utilities`
    * [x] Create script to merge patches into quilt-based patches
    * [x] Create script to download and unpack files based on multiple download .ini files
* [x] Trim devutils and update CI checks
* [x] Update ungoogled-chromium-debian `debian_buster` branch, then other branches
    * [x] Create script to retrieve ungoogled-chromium files at certain revision. Perhaps read revision from a text file?
    * [x] Create script to validate and update patches
    * [x] Move Debian-specific patches back into repo
    * [x] Document new tagging system for platform
* [x] Update documentation with new links and references (mainly building, design)
* [x] Create repos for all other platforms (i.e. `ungoogled-chromium-*`)
* [x] Merge changes into master
* [x] Update Wiki, ungoogled-chromium-binaries

intika:
I will add "Clean issues.." 57 are open and almost haft of then should be closed... just to keep things organized and avoid wasting time ;)

Eloston:
Main repository and Debian have finished migration. However, other platform repositories have quite a bit more work.

PR #708, opened 2019-03-17
Fix missing substitutions in log messages

riyad:


Eloston:
Thanks for taking the time to submit a bug fix for this script, but I'll be deprecating it soon as part of #651. The new system won't need to synchronize changes like this anymore.

PR #727, opened 2019-04-10
devutils/validate_patches: Allow comments

xsmile:
Lines starting with '#' will be ignored.

### DIFF
--- a/devutils/validate_patches.py
+++ b/devutils/validate_patches.py
@@ -603,7 +603,8 @@ def _get_files_under_test(args, required_files, parser):
 def _get_series_iterable(series_path):
     """Returns an iterable of the relative patch paths in the series file"""
     with series_path.open(encoding=ENCODING) as series_file:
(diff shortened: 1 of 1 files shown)


Eloston:
Hmm, I didn't realize that quilt allowed comments in series files. If we're going to allow comments, then we should update all the `utils` and `devutils` to be consistent. I'll need to look into this further.

xsmile:
Usually I write comments directly in the series file. The final version of the file should probably not contain any, but it certainly helps while refreshing patches.
I would appreciate such an option, if it doesn't involve too many changes.

Eloston:
So far I have left notes about patches in the patch files themselves. I'm not entirely sure what I would use comments for in the series file. How are you using this in your workflow?

xsmile:
Mostly to toggle inclusion of patches and sometimes for notes while updating them. There are of course other ways to handle it.

PR #742, opened 2019-04-23
Enabled ability to use Paste and Go with wheel on New Tab Button.

23rd:
Another feature that is hidden from non-linux users is the `Paste and Go` with wheel on the New Tab Button.
It would be cool to have this feature on other platforms.
I dont see any reason to add setting in `chrome://flags` for it — if someone does not need this function, then just do not use it, still pressing the New Tab Button with wheel did nothing before.

image

### DIFF
--- a/patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
+++ b/patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
@@ -1,0 +1,31 @@
+--- a/chrome/browser/ui/views/tabs/new_tab_button.cc
++++ b/chrome/browser/ui/views/tabs/new_tab_button.cc
+@@ -64,10 +64,8 @@ const gfx::Size NewTabButton::kButtonSize{28, 28};
+ NewTabButton::NewTabButton(TabStrip* tab_strip, views::ButtonListener* listener)
+     : views::ImageButton(listener), tab_strip_(tab_strip) {
+   set_animate_on_state_change(true);
+-#if defined(OS_LINUX) && !defined(OS_CHROMEOS)
+   set_triggerable_event_flags(triggerable_event_flags() |
+                               ui::EF_MIDDLE_MOUSE_BUTTON);
+-#endif
+ 
+   // Initialize the ink drop mode for a ripple highlight on button press.
+   ink_drop_container_ = new views::InkDropContainerView();
+--- a/chrome/browser/ui/views/tabs/tab_strip.cc
++++ b/chrome/browser/ui/views/tabs/tab_strip.cc
+@@ -2662,15 +2662,12 @@ void TabStrip::ButtonPressed(views::Button* sender, const ui::Event& event) {
+     if (event.IsMouseEvent()) {
+       const ui::MouseEvent& mouse = static_cast<const ui::MouseEvent&>(event);
+       if (mouse.IsOnlyMiddleMouseButton()) {
+-        if (ui::Clipboard::IsSupportedClipboardType(
+-                ui::CLIPBOARD_TYPE_SELECTION)) {
+           ui::Clipboard* clipboard = ui::Clipboard::GetForCurrentThread();
+           CHECK(clipboard);
+           base::string16 clipboard_text;
+           clipboard->ReadText(ui::CLIPBOARD_TYPE_SELECTION, &clipboard_text);
+           if (!clipboard_text.empty())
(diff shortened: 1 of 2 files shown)


patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
```diff
@@ -0,0 +1,34 @@
+--- a/chrome/browser/ui/views/tabs/new_tab_button.cc
++++ b/chrome/browser/ui/views/tabs/new_tab_button.cc
+@@ -63,10 +63,8 @@ const gfx::Size NewTabButton::kButtonSize{28, 28};
+ NewTabButton::NewTabButton(TabStrip* tab_strip, views::ButtonListener* listener)
+     : views::ImageButton(listener), tab_strip_(tab_strip) {
+   set_animate_on_state_change(true);
+-#if defined(OS_LINUX) && !defined(OS_CHROMEOS)
+   set_triggerable_event_flags(triggerable_event_flags() |
+                               ui::EF_MIDDLE_MOUSE_BUTTON);
```
Eloston: Just to be certain, is this constant defined in macOS and Windows?

patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
```diff
@@ -0,0 +1,34 @@
...
+--- a/chrome/browser/ui/views/tabs/tab_strip.cc
++++ b/chrome/browser/ui/views/tabs/tab_strip.cc
+@@ -2699,15 +2699,15 @@ void TabStrip::ButtonPressed(views::Button* sender, const ui::Event& event) {
+     if (event.IsMouseEvent()) {
+       const ui::MouseEvent& mouse = static_cast<const ui::MouseEvent&>(event);
+       if (mouse.IsOnlyMiddleMouseButton()) {
+-        if (ui::Clipboard::IsSupportedClipboardType(
+-                ui::CLIPBOARD_TYPE_SELECTION)) {
++        //if (ui::Clipboard::IsSupportedClipboardType(
++                //ui::CLIPBOARD_TYPE_SELECTION)) {
```
Eloston: It's not really clear from this patch alone why this needs to be commented out. Could you elaborate on this?

23rd: https://github.com/chromium/chromium/blob/2d57e5b8afc6d01b344a8d95d3470d46b35845c5/ui/base/clipboard/clipboard.h#L40-L50

Here is `IsSupportedClipboardType` method, that always returns `false` when type is `CLIPBOARD_TYPE_SELECTION` and platform is non-Linux. So I hided this condition.
I thought that changing something in this method is quite dangerous, because something can break in a completely different place.

patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
23rd: https://github.com/chromium/chromium/blob/2d57e5b8afc6d01b344a8d95d3470d46b35845c5/chrome/browser/ui/views/toolbar/toolbar_view.cc#L174-L222

This is the place where the `ui::EF_MIDDLE_MOUSE_BUTTON` is actively used for all platforms.

Eloston: Alright, thanks for pointing that out

patches/extra/ungoogled-chromium/enable-paste-and-go-new-tab-button.patch
Eloston: Hmm I see. The reason I asked about this is [Line 27 (of the patch file itself)](https://github.com/Eloston/ungoogled-chromium/pull/742/files#diff-44c235527014ca5a4e29a16670d058f6R27) uses `CLIPBOARD_TYPE_SELECTION`, so I was wondering if this could cause any problems on non-Linux platforms.

23rd: It works fine on OS X.
Unfortunately, windows build is broken and I can't test it there. =(

But it always has condition for non-empty text from clipboard.

Eloston: I see. As long as the clipboard always exists (even if it only returns empty strings), then that works for me. Thanks.

PR #758, opened 2019-05-08
Update to 74.0.3729.136

saltama:
Patches refreshed and validated.


qvint:
According to https://chromereleases.googleblog.com/search/label/Desktop%20Update, 74.0.3729.131 is still the latest stable *desktop* release...

saltama:
Thanks @qvint for pointing this out, will check there from now on. I guess you were hinting to the fact that this could be a problem from the browser fingerprinting point of view...

Eloston:
Hmm, I didn't consider the fingerprinting perspective. I think it should be fine for now since users can just change the user agent.

PR #807, opened 2019-07-30
Update to Chromium 76.0.3809.87

Zoraver:
The most notable change is that changes to a number of files were removed from disable-gaia.patch, as they appeared to no longer be necessary (in Chromium 76, the files in question no longer contain any reference to gaia or no longer exist).


Zoraver:
I just tried to compile a portablelinux build using this patchset, and I ran into a build issue due to a typo. I would hold off merging this to master until it is fixed.

Zoraver:
It should be fixed now, but I'll let you know once I have a successful build.

leo-lb:
@Zoraver Thanks a lot for this work. I hit some of these safebrowsing issues while building in debug mode in 75.x releases, release mode was working fine.

xsmile:
@Zoraver Please remove the patch `core/inox-patchset/chromium-exclude_unwind_tables.patch` and the GN flag `exclude_unwind_tables=true`. If `is_official_build=true` is set, it is now implicitly enabled.

mbakke:
`core/inox-patchset/chromium-exclude_unwind_tables.patch` is still needed for GCC builds, because `is_official_build` requires Clang.

I don't mind "downstreaming" the patch, but the change seems out of scope of this PR, as it has been obsolete in ungoogled-chromium for many versions now.

Zoraver:
Unfortunately, no. I am still working on resolving build issues related to building without safebrowsing. I am making progress, but each build attempt takes between 3 to 4 hours, which has slowed me down.

leo-lb:
@Zoraver Hey, what forces you to recompile from scratch every time instead of using incremental builds by re-running `ninja`?

Eloston:
@leo-lb @Zoraver Even with [the workflow outlined here](https://github.com/Eloston/ungoogled-chromium/blob/master/docs/developing.md#steps-for-fixing-patches-after-a-failed-build-attempt), I believe all the files that are touched by domain substitution will cause ninja to rebuild them because of the file metadata. That may add about an hour of rebuilding the same files.

I believe ninja uses the file timestamps to determine whether to rebuild files, so perhaps we could preserve timestamps when making the domain substitution cache...

Zoraver:
I am pleased to announce that these changes are ready to merge. I have successfully built ungoogled-chromium 76.0.3809.87 for both the linux-portable and macOS targets.

@leo-lb Thanks for reminding me about this feature. I have used it before, but recently it has slipped my mind to use it.

leo-lb:
@Zoraver Just merged in my (fork)[https://gitlab.com/lle-bout/ungoogled-chromium] :-] Thanks.

CI running: https://gitlab.com/lle-bout/ungoogled-chromium/-/jobs/264393763

PR #810, opened 2019-08-09
Update to Chromium 76.0.3809.100

Zoraver:
No manual edits to the patchset were needed. I am currently running a build for the portablelinux target and expect that it will be successful.


Zoraver:
The build for portablelinux succeeded; feel free to merge.

PR #823, opened 2019-09-11
Update for 77.0.3865.75

xsmile:
- `patches/core/ungoogled-chromium/block-trk-and-subdomains.patch` might need updating. Specifically `chrome/browser/profiles/profile_io_data.cc` has changed
- `patches/extra/debian_buster/fixes/inspector.patch` removed, as it breaks the build


patches/core/ungoogled-chromium/block-trk-and-subdomains.patch
```diff
@@ -162,37 +162,17 @@
        url::kBlobScheme,
        base::WrapUnique((*protocol_handlers)[url::kBlobScheme].release()));
    DCHECK(set_protocol);
---- a/chrome/browser/profiles/profile_io_data.cc
-+++ b/chrome/browser/profiles/profile_io_data.cc
```
Eloston: Created new issue to address this.

tangalbert919:
Can a tag be created for this so we can update the projects using this repo as a submodule?

Eloston:
@tangalbert919 Whoops, looks like I forgot after I closed the issure regarding the `trk` scheme patch. Done.

PR #832, opened 2019-09-28
Add common flags; allow change of logging level

wchen342:
- Add a function to `_common.py` for common cmdline args
 - Add a function to set logging level separately

### DIFF
--- a/devutils/validate_patches.py
+++ b/devutils/validate_patches.py
@@ -26,7 +26,7 @@ sys.path.pop(0)
 
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / 'utils'))
 from domain_substitution import TREE_ENCODINGS
-from _common import ENCODING, get_logger, get_chromium_version, parse_series
+from _common import ENCODING, get_logger, set_logging_level, get_chromium_version, parse_series, add_common_params
 sys.path.pop(0)
 
 try:
@@ -626,8 +626,8 @@ def main():
         metavar='DIRECTORY',
         default='patches',
         help='The patches directory to read from. Default: %(default)s')
-    parser.add_argument(
-        '-v', '--verbose', action='store_true', help='Log more information to stdout/stderr')
+    add_common_params(parser)
+
     file_source_group = parser.add_mutually_exclusive_group(required=True)
     file_source_group.add_argument(
         '-l',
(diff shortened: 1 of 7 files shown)


devutils/validate_patches.py
```diff
@@ -661,10 +661,7 @@ def main():
     if not args.patches.is_dir():
         parser.error('--patches path is not a directory or not found: {}'.format(args.patches))
 
-    if args.verbose:
-        get_logger(initial_level=logging.DEBUG)
-    else:
-        get_logger(initial_level=logging.INFO)
+    set_logging_level(verbose=args.verbose, quiet=args.quiet)
```
Eloston: Instead of calling `set_logging_level` everywhere, I'm wondering if there's a more concise way to do this, like:

1. Pass `args` into a function that handles all common arguments, e.g. `process_common_params`. We can call `set_logging_level` in this new function.
2. Create a subclass of `Action` and set it during the `add_argument` invocation in `add_common_params`. That way, we don't even need to invoke `process_common_params` in all the callbacks. This will also discourage interdependency between flags, which I think is an anti-pattern.

utils/_common.py
```diff
@@ -54,6 +52,22 @@ def get_logger(initial_level=logging.INFO):
     return logger
 
 
+def set_logging_level(verbose=False, quiet=False):
+    """Sets logging level of logger and all its handlers"""
+
+    default_level = logging.INFO
+    logging_level = default_level + 10 * (quiet - verbose)
```
Eloston: This is a clever and succinct solution, but I think this is not great from a user perspective because:

1. The verbose and quiet flags have an implicit dependency on each other. And when both are applied, surprising behavior occurs (based on the help messages you wrote)
2. There are no comments to describe how this works; e.g. it is implicit that the value `10` will change it to different logging levels. Also, is this behavior a guarantee from the Python API, or just an implementation detail?

I think the better solution is to let the user decide which logging level they want. Either a numerical value or a string to specify the logging level would make things more clear (maybe even accept both)?

Eloston:
Also, the `code_check` CI task is failing due to some small formatting issues; would you mind fixing that up?

devutils/validate_patches.py
wchen342: Ah, didn't know there is an `Action` class. Then it will be much clear to use this class instead of a global function.

utils/_common.py
wchen342: The values are in Python documents so it should be guaranteed, but I think you are right, it will be better for users to specify the logging level. I think I can just make verbose and quiet alias of some logging levels.

wchen342:
> Also, the `code_check` CI task is failing due to some small formatting issues; would you mind fixing that up?

I will fix that once everything else is ready.

PR #836, opened 2019-10-10
Update Chromium to 77.0.3865.120

tangalbert919:


howaboutsynergy:
there's 121 already xD https://chromium.googlesource.com/chromium/src.git/+refs
oh wait, actually, nevermind, i forgot this project uses `.tar.xz` instead of git (and I don't recommend trying to use git, there's stuff downloaded from google and ran as executable before building)

jstkdng:
welp, gonna build this once I get home.

PR #850, opened 2019-11-02
Some Android-specific fixes which are beneficial/neutral for all variants

csagan5:
This PR contains a commit to make the first part of https://github.com/ungoogled-software/ungoogled-chromium-android/blob/master/patches/android-prune-domain-fix.patch redundant and another commit to not replace google.com in the prepopulated search engines list.

### DIFF
--- a/domain_substitution.list
+++ b/domain_substitution.list
@@ -4771,7 +4771,6 @@ third_party/blink/renderer/core/dom/class_collection.cc
 third_party/blink/renderer/core/dom/class_collection.h
 third_party/blink/renderer/core/dom/comment.idl
 third_party/blink/renderer/core/dom/container_node.cc
-third_party/blink/renderer/core/dom/document.cc
 third_party/blink/renderer/core/dom/document_test.cc
 third_party/blink/renderer/core/dom/dom_implementation.idl
 third_party/blink/renderer/core/dom/element.cc
--- a/domain_substitution.list
+++ b/domain_substitution.list
@@ -2831,7 +2831,6 @@ components/safe_search_api/stub_url_checker.cc
 components/safe_search_api/url_checker_unittest.cc
 components/search_engines/default_search_policy_handler_unittest.cc
 components/search_engines/keyword_table_unittest.cc
-components/search_engines/prepopulated_engines.json
 components/search_engines/template_url.cc
 components/search_engines/template_url.h
 components/search_engines/template_url_data_unittest.cc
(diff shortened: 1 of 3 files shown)


domain_substitution.list
```diff
@@ -2831,7 +2831,6 @@ components/safe_search_api/stub_url_checker.cc
 components/safe_search_api/url_checker_unittest.cc
 components/search_engines/default_search_policy_handler_unittest.cc
 components/search_engines/keyword_table_unittest.cc
-components/search_engines/prepopulated_engines.json
```
Eloston: Unless I am missing some context on the changes in Android, this won't help us because of https://github.com/Eloston/ungoogled-chromium/blob/5931f25336c8bbd083a08d564534fbffdc031217/patches/core/ungoogled-chromium/replace-google-search-engine-with-nosearch.patch

wchen342: I think it is probably a slight different problem, that replacing links in that file will prevent users who wants to switch back to Google from using search directly?

csagan5: @Eloston that patch would have to be updated.

I am proposing to replace it **only** in the patch, not via domain substitution; this is also because in Bromite that search engine is allowed.

domain_substitution.list
```diff
@@ -4772,7 +4771,6 @@ third_party/blink/renderer/core/dom/class_collection.cc
 third_party/blink/renderer/core/dom/class_collection.h
 third_party/blink/renderer/core/dom/comment.idl
 third_party/blink/renderer/core/dom/container_node.cc
-third_party/blink/renderer/core/dom/document.cc
```
csagan5: I am adding this because there are no remotely-accessed URLs to replace and it would break https://github.com/bromite/bromite/blob/78.0.3904.63/build/patches/Inject-scripts-for-AMP-tracking-ads-and-video-functionality.patch

domain_substitution.list
Eloston: Fair enough. But you need to add this path to `PRUNING_EXCLUDE_PATTERNS` in `devutils/update_lists.py`, otherwise your change will be overridden.

domain_substitution.list
Eloston: Sure. This also needs to be added to `devutils/update_lists.py`

domain_substitution.list
csagan5: @Eloston I have added both exclusions to `DOMAIN_EXCLUDE_PREFIXES` (I think you meant that one)

csagan5:
Thanks; I will propose a small change in `ungoogled-chromium-android` for the regex part.

PR #852, opened 2019-11-04
Update README.md

travankor:
Related:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=943909
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=941060


Eloston:
Seems strange that they'd depend on Debian patches. But if this is the only Chromium browser in Guix, then it makes some sense. Otherwise, I think the ungoogled-chromium Guix package should depend on those patches instead of Debian's.

PR #862, opened 2019-11-16
Remove filenames matching 'test' from the domain substitution list

csagan5:
@Eloston do you agree with this change and is there a way to use a regex with DOMAIN_EXCLUDE_PREFIXES/DOMAIN_INCLUDE_PATTERNS?


Eloston:
The trkbar comment change LGTM, thanks for that.

What is the purpose of removing filenames matching "test"?

csagan5:
> The trkbar comment change LGTM, thanks for that.

Feel free to cherry-pick it, it's a small one compared to the other commit.

> What is the purpose of removing filenames matching "test"?

For two reasons: it is unnecessary and if we were to run the tests it would break them.

Eloston:
> it is unnecessary

How is this a problem? The more complicated logic you add, the more difficult it is to be certain that you'd be catching all the cases that you want to catch. I don't want to add rules that are not absolutely necessary.

> if we were to run the tests it would break them

What kind of tests and how? Any tests requiring a connection to Google should be modified as part of ungoogling.

csagan5:
> How is this a problem?

It is a patch maintenance burden.

Eloston:
@csagan5 I'm still having troubles understanding what you mean. Are you saying that domain substitution is breaking code that shouldn't be domain substituted (and thus requiring patches to fix it), or is the time it takes to substitute all the testing files a burden?

csagan5:
@Eloston no I am saying that those tests are not being run, but with the domain substitution changes I expect them to fail.

> the time it takes to substitute all the testing files a burden?

Sometimes I use a temporary commit which contains all the domain substitution changes and it slows down a bit rebasing etc

Basically in my workflow I do not allow "dirty trees" and thus having a lot of unnecessary changes is an annoyance, but I do not insist on this as we obviously have different workflows :relaxed:

Eloston:
> Sometimes I use a temporary commit which contains all the domain substitution changes and it slows down a bit rebasing etc
> 
> Basically in my workflow I do not allow "dirty trees" and thus having a lot of unnecessary changes is an annoyance, but I do not insist on this as we obviously have different workflows

Interesting, I always considered domain substitution to be more like a temporary patch, rather than a permanent patch for the source code. That's why I made it reversible by implementing the domain substitution cache and timestamp manipulation.

However, you're right that substituting all the test files slows down the process a bit; if substitution is not I/O bound, I could parallelize it.

csagan5:
> I consider domain substitution to be more like a temporary patch that is only applied when building the browser.

Me too, but I learnt the hard way that ignorning modified files before a build starts can lead to not building what you think you are building, so I enforce that there are no modified files before starting a build.

> if substitution is not I/O bound, I could parallelize it.

My thinking was that the test files are not used by the build process (because we do not build and run tests, but just build), thus changing them equates to noise. I was also pretty confident that no file matching `*test*` gets built, but I might be wrong there.

PR #878, opened 2019-12-02
Add tag automation

### LINKED ISSUE #866: Automated tag creation from GitHub
Because I'm lazy, we should have a way to create a tag based on some metadata in a PR. This can include:

* A comment or review on the PR that tells some automated process to create a tag after it is merged (but not closed).
* Same as above, but we use a label on the PR instead.

Ideally, this automated process would be something that runs on GitHub rather than a 3rd-party bot/agent (Is this possible via GitHub Actions?). It may also need to make a commit to revert `revision.txt` back to 1 on a Chromium update.

This feature would be especially useful for platform repos.

tangalbert919:
This should handle making new tags for us when a PR is merged.

This is basically a scaled-down version of [GitHub Tag Bump](https://github.com/marketplace/actions/github-tag-bump), and has been specifically designed for use in this repository.

EDIT: ~~Upon merging this PR, tag `78.0.3904.108-3` will automatically be created.~~ This PR will create a tag the next time `chromium_version.txt` or `revision.txt` are modified.

EDIT 2: ~~It might be best to merge #883 and #880 first. Then this can be merged afterwards to automatically create tag `79.0.3945.79-1`.~~

EDIT 3: Should've mentioned this, but resolves #866


tangalbert919:
I'll go ahead and address a few things now, while I have time:

> I want to specify when new tags will be created. I don't want tags to be created on every push; otherwise documentation changes will create new tags.

On this, the only way to specify when new tags will be created is by modifying either `chromium_version.txt` and `revision.txt`. I made a minor revision that stops the script if the contents of both files put together match the most recent tag.

> The script should know when to bump a revision when the same Chromium version is used, and reset the revision when a new Chromium version is used. The action should create the commit to update this revision automatically.

Bumping the revision, I'm not sure how I can make the script know when to do that. Resetting the revision, I could do that if GitHub Actions (and the API) allow making commits and pushing it, but I want to keep the script simple, which is why bumping or resetting the revision is still a manual thing.

> I'd like to have a message in the tag body, such as "Update to Chromium x.x.x.x" and then a list of commit titles that are included in the change. But if only the revision changes, then the message should start with "Bump revision to x" or something more meaningful.

That sounds like something for the Releases tab, which sounds like something outside of tag automation, which is what I want to avoid in the sake of keeping the tag automation as simple as possible.

> I'd like to use git to create the tag, unless using the GitHub APIs have some advantage(s) over using git that I'm not aware of.

Using the GitHub APIs is the better option just to keep things simple.

> I don't really understand why there is both a `.github/workflows` file and a `action.yml`, but these are aspects of GitHub Actions that I will need to learn about at some point. I also don't understand the choices of Alphine for the Dockerfile, and Ubuntu for the GitHub Workflow.

Those two files just define how the action will work, and what the action is. As for why I picked Alpine and Ubuntu for the Dockerfile and GitHub Workflow, I just went with the examples.

> I also don't like how this adds more files to the root of the repository. We don't need any more clutter at the root of the repository. Ideally, we should keep all these changes contained within one directory.

I'll see if I can make that work.

tangalbert919:
Those new files at the root of the repository (`action.yml` and `Dockerfile`) have been removed (we don't need them), but for the most part, manual editing of either `chromium_version.txt` and `revision.txt` is still required for a new tag to be made.

PF4Public:
Sorry for jumping in, but I would suggest renaming .yml file according to its intended use. I'm planning soon to do a PR with [another GitHub action](https://github.com/PF4Public/ungoogled-chromium/blob/master/.github/workflows/new_version_check.yml). I think that having more than one is a sufficient amount to require proper names.

tangalbert919:
#1435 seems to have resolved this, so I'll just close this PR.

PR #903, opened 2020-01-05
Removed Arch Linux untrusted AUR package

Rowisi:
The `ungoogled-chromium-bin` package should have not been added here from the beginning, its not only an unofficial package but also very suspicious that doesn't download from the official built binaries but from the package owner's link itself.

the package creator made the download link to his own path: `download.opensuse.org/repositories/home:/justkidding:/arch/standard/x86_64/ungoogled-chromium-$pkgver-1-x86_64.pkg.tar.xz` which give him freedom to do anything with it.

`ungoogled-chromium-bin` is now deleted from the AUR but it can be recreated by other users so its better to be deleted from README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,6 @@ These binaries are known as **contributor binaries**.
 Also, ungoogled-chromium is available in several **software repositories**:
 
 * Arch Linux: [Available in AUR as `ungoogled-chromium`](https://aur.archlinux.org/packages/ungoogled-chromium/)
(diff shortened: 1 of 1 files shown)


jstkdng:
Dude wtfm why would you delete it?
"Official" packages aren't really official as in built by Eloston as he uses debian, all the "Official" binaries can be submitted by anyone. They suffer the same trust issue.
At least my packages from the OBS have all the sources and build logs there. If I ever did anything suspicious anyone could call my out on it.

Rowisi:
> Dude wtf, why would you delete it?

Its already deleted from the AUR, so what's the point of keeping it here anyway?

PR #904, opened 2020-01-08
Update to Chromium 79.0.3945.117

tangalbert919:
Patches were refreshed with the `--sort` flag, which is why there's so many changes.


Eloston:
Hmm, I didn't know of the `--sort` flag. Thanks.

PR #907, opened 2020-01-12
Remove lines causing Android build to fail

wchen342:
The removed lines in the patch will lead to wrong number of arguments when building for Android.

### DIFF
--- a/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
+++ b/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
@@ -601,22 +601,6 @@
  
  namespace policy {
  BrowserSigninPolicyHandler::BrowserSigninPolicyHandler(Schema chrome_schema)
-@@ -49,7 +48,6 @@ void BrowserSigninPolicyHandler::ApplyPo
-             // The new kSigninAllowedOnNextStartup pref is only used on Desktop.
-             // Keep the old kSigninAllowed pref for Android until the policy is
-             // fully deprecated in M71 and can be removed.
--            prefs::kSigninAllowed,
- #else
-             prefs::kSigninAllowedOnNextStartup,
- #endif
-@@ -61,7 +59,6 @@ void BrowserSigninPolicyHandler::ApplyPo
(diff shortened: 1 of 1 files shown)


wchen342:
So the original source code looks like this:
```
      case BrowserSigninMode::kEnabled:
        prefs->SetValue(
#if defined(OS_ANDROID)
            // The new kSigninAllowedOnNextStartup pref is only used on Desktop.
            // Keep the old kSigninAllowed pref for Android until the policy is
            // fully deprecated in M71 and can be removed.
            prefs::kSigninAllowed,
#else
            prefs::kSigninAllowedOnNextStartup,
#endif
            base::Value(true));
        break;
      case BrowserSigninMode::kDisabled:
        prefs->SetValue(
#if defined(OS_ANDROID)
            // The new kSigninAllowedOnNextStartup pref is only used on Desktop.
            // Keep the old kSigninAllowed pref for Android until the policy is
            // fully deprecated in M71 and can be removed.
            prefs::kSigninAllowed,
#else
            prefs::kSigninAllowedOnNextStartup,
#endif
            base::Value(false));
        break;
```

The current patch actually does nothing for non-Android systems because if you look at the macro, the line current patch deleted is wrapped in `#if defined(OS_ANDROID)`. However, on Android because of the macro and the deleted line, now `prefs->SetValue()` will get one argument instead of two. Not sure about `signin_pref_names.h` but I have sucessfully built Android version with the original source of this part, so I think it should be fine.

wchen342:
Ok so when I built my Android build, I actually put `prefs::kSigninAllowedOnNextStartup` there which is not supposed to exist in Android version. That's why the build was successful. So I think the two `case` sections should be removed altogether. Will test on Arch version and report back.

Eloston:
> I actually put prefs::kSigninAllowedOnNextStartup there which is not supposed to exist in Android version. That's why the build was successful. So I think the two case sections should be removed altogether. 

I see, thanks for investigating. I must've forgotten about Android when writing this patch.

wchen342:
I have updated the patch. I tested it on Arch and it works all right.

PR #934, opened 2020-02-11
Update to Chromium 80.0.3987.106

tangalbert919:
~~Do not merge! Google just released version 80.0.3987.106, and I need to get those changes in.~~

EDIT: This PR can be merged now.


kramred:
I've built 80.0.3987.100 for macOS (based on an earlier version of this PR) and I've noticed that this version is sending [ssdp](https://en.wikipedia.org/wiki/Simple_Service_Discovery_Protocol) UDP multicast messages to 239.255.255.250:1900
ssdp_multicast_chromium

The payload is: 
```
M-SEARCH * HTTP/1.1
HOST: 239.255.255.250:1900
MAN: "ssdp:discover"
MX: 1
ST: urn:dial-multiscreen-org:service:dial:1
USER-AGENT: Chromium/80.0.3987.100 Mac OS X
```

Might be related to Chromecast – Anyone else noticed this or any ideas what this is about?

jstkdng:
yeah, probably chromium looking for a chromecast, or maybe the other way around, just chromium exposing its capabilities to other devices running on the network, I don't think it is harmful since it is not sending anything to google.

PF4Public:
We should probably petition @Eloston to issue commit rights for @tangalbert919  ;)

PR #961, opened 2020-03-12
Add WinRAR support

### LINKED ISSUE #956: Option to use WinRAR in the unpack script instead of 7-Zip
Hey everyone,

I downloaded the source code using the download.py script and tried to unpack it when I noticed that the script tried to call 7-Zip. I don't have 7-Zip installed on my computer, nor plan to, due to a conflict between my country's upcoming election (United States) and the Russian Federation (where the 7-Zip maintainer is located). Please respect my beliefs in this matter and not try to convince me to use 7-Zip just because it's FOSS.

Anyways, I wanted to put in a request for WinRAR support in the unpack script as an alternative since WinRAR has a CLI and is made in Germany/the European Union. I figured WinRAR would be the next common archiving tool on Windows after 7-Zip.

tangalbert919:
This PR resolves #956, and may be really unpopular among certain users (specifically, those who would advise installing 7-zip).

### DIFF
--- a/utils/_common.py
+++ b/utils/_common.py
@@ -14,7 +14,7 @@ from pathlib import Path
 
 ENCODING = 'UTF-8' # For config files and patches
 
-SEVENZIP_USE_REGISTRY = '_use_registry'
+USE_REGISTRY = '_use_registry'
 
 LOGGER_NAME = 'ungoogled'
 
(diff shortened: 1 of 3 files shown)


tangalbert919:
The linter errors have been fixed.

PR #969, opened 2020-03-19
Update maintainer's nickname for Ubuntu PPA

TotalCaesar659:
I've noticed that Ubuntu PPA maintainer's nickname is @braewoods, not @ryou. If it is, please, let me know.


braewoods:
That would be correct. The name changed awhile ago and that is a different user.

PR #1006, opened 2020-04-19
Returned checkbox to remember association of app with URL Scheme.

23rd:
After [this commit](https://github.com/chromium/chromium/commit/9f8e95c96f5de02811f5d3cbc74da72714498533) the "Always open" checkbox became controlled by an additional policy that has `false` as a default value. 
So, regular users who don't want to deal with the policies on their OS can no longer remember opening external applications and must always make an extra click.
This simple PR adds a new patch file that replaces `false` default value with `true` and returns the previous behaviour.

image


PF4Public:
Could this help with non-working #extension-mime-request-handling?

23rd:
I'm not really sure.

PR #1021, opened 2020-04-29
Add about page branding patch

kramred:
This PR addresses issue https://github.com/Eloston/ungoogled-chromium/issues/659
It adds a string (with hyperlink to the github repo) to the "About Chromium" page at  `chrome://settings/help` (does not affect strings in `chrome://version/`)

Screen Shot 2020-04-29 at 20 43 56

Feel free to modify this however you'd like to have it displayed. 
E.g. anchor target is currently __not__ a new tab and I chose square brackets instead of parentheses.


PF4Public:
Would it be more unambiguous to write sometheing like this:

Version 81.0.4044.129 (Official Build) (64-Bit) (with ungoogled-chromium patches)

also without the link, as someone intentionally installing ungoogled-chromium should already know all the links. Besides simple text won't be standing out a lot unlike a link.

qvint:
There is already [an environment variable](https://github.com/chromium/chromium/blob/856dcc8c9aee22a12732323ee9eb668a715b914d/chrome/tools/build/linux/chrome-wrapper#L138) to modify the version string:
```bash
CHROME_VERSION_EXTRA="(with ungoogled-chromium patches)"
```

This additional text is displayed in all possible places:
- chrome://settings/help
- chrome://version/
- chrome --version

PF4Public:
> There is already [an environment variable](https://github.com/chromium/chromium/blob/856dcc8c9aee22a12732323ee9eb668a715b914d/chrome/tools/build/linux/chrome-wrapper#L138) to modify the version string:

This makes it a platform repo responsibility of adding this into version string?

kramred:
Then I would be in favour of using the `CHROME_VERSION_EXTRA` environment variable instead of adding another patch.

> There is already [an environment variable](https://github.com/chromium/chromium/blob/856dcc8c9aee22a12732323ee9eb668a715b914d/chrome/tools/build/linux/chrome-wrapper#L138) to modify the version string:
> 
> ```shell
> CHROME_VERSION_EXTRA="(with ungoogled-chromium patches)"
> ```
> 
> This additional text is displayed in all possible places:
> 
> * chrome://settings/help
> * chrome://version/
> * chrome --version

Eloston:
Yeah I agree using the environment variable would be better for platforms that can easily support it (e.g. Debian, Arch Linux). For platforms that are essentially just the Chromium binary only like Portable Linux, macOS, and Windows, I think it's fine to include a patch like this in each of their repos.

PR #1022, opened 2020-04-30
Replace PPA with OBS

braewoods:


jstkdng:
Can't one use the OBS as a sort of ppa? I ask since an OBS project can be set up to be used as a repository under arch.

braewoods:
Yes it can. The updated documentation replaces the PPA with a set of instructions for how to enable the OBS repository. I thought that much was obvious. Is there still some confusion about something?

jstkdng:
calm down there m8, was just asking since I don't use the OBS for ubuntu stuff.
It's all good now.

braewoods:
Ok. Sorry. I was just wondering if there was more to explain.

PR #1046, opened 2020-05-22
Update to chromium 83.0.4103.61

jstkdng:
First PR, so go easy on me, ty


jstkdng:
set_quilt_vars.sh don't seem to work with zsh, will use bash for now, that could be fixed.

jstkdng:
Advance: They changed a lot of stuff from safe browsing, it is getting way complex

jstkdng:
Ok, that should do it I think, will try to build with arch

jstkdng:
It seems they added some type of enhanced safe_browsing, tried to remove those references as well.

Issue #1076, opened 2020-06-15
Updating to Chromium 83.0.4103.106


Zoraver:
## Notes for the developer

Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:

![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

Finally, make sure to reference this issue in your PR.

Feel free to ask questions and discuss issues here along the way. Thanks!

## Notes for others

Feel free to raise issues or questions throughout the process. However, please refrain from asking ETAs unless no visible progress has been made here or on the developer's PR for a while (e.g. 2 weeks).

Zoraver:
I am planning on working on this.

Eloston:
Need to test a new build with the PDF name changes before tagging.

PF4Public:
✔️ Compiled at f08ce8b3f1300ef0750b5d6bf967b9cbbfd9a56d (after both PDF-related commits). Seems to be working fine.

Eloston:
Thanks @PF4Public, tag created.

PR #1077, opened 2020-06-15
Upgrade to Chromium 83.0.4103.106

Zoraver:
Addresses #1076. I will let you know once I have a successful build.


lsfxz:
My build fails with

```
../../chrome/common/chrome_content_client_constants.cc:23:33: error: cannot define or redeclare 'kChromePDFExtensionPluginName' here because namespace '' does not enclose namespace 'ChromeContentClient'
```

which seems to be unrelated to this PR but instead to 583b3e4c4b62c6c53b4efba2377c0c1f9845ad7a?

Eloston:
@lsfxz 

> which seems to be unrelated to this PR but instead to 583b3e4?

Thanks for testing, it's probably a typo I made. I tried to test the code on Portable Linux but got a somewhat cryptic STL error (something with EraseIf template declaration not being seen) with LLVM 8.0.1 from Debian buster backports. What platform are you building on? And what LLVM version?

Zoraver:
My build succeeded and is running well; feel free to merge. Note that I started my build before either of the PDF plugin name patches were introduced.

Eloston:
@Zoraver How are you testing the builds? I'm trying to build Portable Linux with these changes using LLVM 8.0.1 from Debian buster backports, but it isn't working for me (strange STL errors).

Zoraver:
I have been building the Portable Linux target in a Fedora 31 container using clang 9.0.1. I should probably move to Fedora 32 + clang 10, but I haven't had time to...

Eloston:
Does that mean the instructions in the Portable Linux README are out of date? They still recommend LLVM 8

Zoraver:
They might be; the majority of the instructions were written before I started contributing to the project. I haven't built on Debian, so I can't attest to the accuracy of the Debian-specific info. Looking at the value of `RELEASE_VERSION` in `tools/clang/scripts/update.py`, Google appears to be using clang 11.0.0, which seems to be still in development.

lsfxz:
@Eloston building on Arch (x86_64) and Manjaro ARM (aarch64); llvm 10.0.0 at the moment.
The x86_64 build also succeeded btw; I've also started another build from the fix at f08ce8b – will report back if any issues remain with it :)

The whole thing, jobs, pipelines etc. is over at [gitlab](https://gitlab.com/ohfp/caidao/-/pipelines)  – I'm building it for a rather small audience there with minor changes (changing the name / branding so you can run it alongside chromium / ungoogled-chromium; building for aarch64 as well and including the vaapi/vdpau-patches). There are not many changes to the way it gets built compared to the way things are done by [ungoogled-chromium in the AUR](https://aur.archlinux.org/packages/ungoogled-chromium) – so there's probably not much informational stuff there in addition to [ungoogled-software/ungoogled-chromium-archlinux](https://github.com/ungoogled-software/ungoogled-chromium-archlinux) and [jstkdng/ungoogled-chromium-archlinux](https://github.com/jstkdng/ungoogled-chromium-archlinux) anyway ^^.

jstkdng:
hey @lsfxz , do you know the limits of gitlab runners? Last I checked they only have 1 core a few gb of ram. Are you using your own runners?
I ask as I'm trying to use github runners to build and I'm already past the 10 hour mark and you seem to be able to build in less than 10 hours.

lsfxz:
The gitlab runners are rather limited as well; I can't recall the constraints from the top of my head, but definitely not enough to build chromium.

So yeah, after having looked around quite a bit (esp. for aarch64), there were a few interesting discoveries like https://cloud.drone.io – but nothing (free-ish) that was enough for rather large builds like building chromiunm. 

In the end I caved in; now I'm using my own runners – the x86_64 builds are running on an "average" VPS that I already had anyway, while I always spin up a temporary (c1.large.arm instance)[https://www.packet.com/cloud/servers/c1-large-arm/] at packet.com for the aarch64 builds.

PF4Public:
Had anyone try Hetzner? They provide some of the cheapest nodes:
image

jstkdng:
Well, cheap is relative to ones economy. Maybe for someone 4 euros is a lot of money. Now, free has the same meaning for everyone no matter your economy.
I was even thinking about using google's own cloud build service to build UC, quite ironic if you ask me. I'd be doxing myself by that so... yeah, not an option.

PF4Public:
> Well, cheap is relative to ones economy.

> They provide some of the cheapest nodes

I meant "some of the cheapest nodes on market"

jstkdng:
yeah, OVH also has great prices.
I don't want to spend more money than what I already am, especially during covid times.

lsfxz:
Automating spinning up (and spinning down) hourly instances (does Hetzner offer an API for that?) for builds would make it possible to keep costs as low as possible – but yeah, still wouldn't be free.

One can always use spare devices "at home" as well – you can just drop a gitlab-runner on some reasonably isolated device and build things on it. This, though, still requires having a spare device at hand, needs electricity and has potential security aspects that one might want to think about beforehand.

Sooo, yeah, somewhat unsatisfying.

For Open Source projects in general though it _might_ be possible to ask for support at packet.com (https://www.packet.com/community/open-source/) or ask at https://cloud.drone.io/ for extending the time limit of the free instances – iirc. they do that if you're asking nicely / have a good reason; might need to take a look at their forums for that (and there's https://github.com/WorksOnArm/cluster for armv7/aarch64).

This would probably make the most sense for the ungoogled-chromium project as a whole – I doubt a "subset" of it (eg. "building for Arch") would be a "good enough reason" to get some support there.

jstkdng:
That could work, but they are asking for your real name, so unless anyone here is willing to dox themselves or packet accepts pseudonyms then it won't be possible.

PF4Public:
> does Hetzner offer an API for that?

Honestly saying, haven't tried, but the do.

> somewhat unsatisfying

Somewhat unsatisfying is Google dropping jumbo support. With it full compilation could fit into free allowances.

lipici:
give some love to windows version :(

jstkdng:
@lipici
github actions support windows. You could adapt https://github.com/ungoogled-software/ungoogled-chromium-archlinux/blob/master/.github/workflows/build.yml to build windows binaries.

lipici:
Sorry, im just a creative person, not a tech, put me to create you a commercial, ideas for any application and so on.
    On Tuesday, July 7, 2020, 8:54:47 PM GMT+3, JustKidding <notifications@github.com> wrote:  
 
 


@lipici
github actions support windows. You could adapt https://github.com/ungoogled-software/ungoogled-chromium-archlinux/blob/master/.github/workflows/build.yml to build windows binaries.

PR #1098, opened 2020-07-07
Add patch for Bookmark-Bar on New-Tab-Page

toluschr:
I really dislike the Bookmark-Bar, since it takes up more space than necessary.
I only toggle it when I really need it, since I still keep some Bookmarklets (For installing chrome store addons, etc.).
Not being able to toggle it on the new-tab-page was quite annoying for me.

This patch makes the Bookmark-Bar toggleable on the NTP (Maybe the flags description is a little confusing).
I hope I didn't do anything wrong, since this is my first time modifying the chromium source code.

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-for-bookmark-bar-ntp.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-for-bookmark-bar-ntp.patch
@@ -1,0 +1,52 @@
+--- a/chrome/browser/about_flags.cc
++++ b/chrome/browser/about_flags.cc
+@@ -273,6 +273,13 @@ const FeatureEntry::Choice kShowAvatarBu
+      "never"}
+ };
+ 
++const FeatureEntry::Choice kBookmarkBarNewTab[] = {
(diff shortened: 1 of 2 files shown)


jstkdng:
Isn't this how chromium already works? You can hide the bar and it will not show (also on the new tab page) unless you manually add a bookmark. It gives more control I guess, but it could become a burden to maintain on future releases. This patch could be a good adition to the contrib repo.

Eloston:
@jstkdng Are you on version 83? On version 81, the bookmark bar unconditionally shows in the NTP.

I agree the benefit is marginal at best, but it's also a very small patch that shouldn't be hard to keep updated. In the event that it does become a pain to maintain, it's fine with me if the patch is dropped because it's an `extra` patch.

toluschr:
I never used quilt before. It should work now

toluschr:
> Isn't this how chromium already works? You can hide the bar and it will not show (also on the new tab page) unless you manually add a bookmark. It gives more control I guess, but it could become a burden to maintain on future releases. This patch could be a good adition to the contrib repo.

@jstkdng 
I created a new profile and noticed, that this doesn't seem to be the case.

PR #1101, opened 2020-07-09
Add patch for Bookmark-Bar on New-Tab-Page

toluschr:
#1098
I'm not smart enough to use git correctly...
`git push fork --mirror` broke the old pr


toluschr:
After reading [this](https://github.com/Eloston/ungoogled-chromium/blob/master/docs/developing.md#adding-command-line-flags-and-chromeflags-options), I refreshed the patches twice and it now seems to work

PR #1116, opened 2020-07-28
Upgrade to Chromium 84.0.4147.105

Zoraver:
Addresses #1115. I will let you know once I have a successful build.


evs-ch:
Can confirm a successful build on aarch64 and an ongoing almost-finished one for x86_64 for Manjaro and Arch.
(Actually with just the previous patch – but with sufficiently fuzzy patching, the patches were applied in the same way your commit would've made them apply).

Zoraver:
My build succeeded and is running well. Feel free to merge.

@evs-ch Thanks for the kind words!

Issue #1126, opened 2020-08-10
Hide 'Current user' button next to the address bar


sgtpep:
Is that possible to hide this button, that allows to sing in the Google account, for those who uses Chromium without singing it to it?

Eloston:
It doesn't look like there's any configuration flag for it. I think the only way is to modify the source code.

PF4Public:
@sgtpep Is this what you're asking for?
![image](https://user-images.githubusercontent.com/10319700/89892574-898b0200-dbdf-11ea-8658-748c1b5be6ad.png)
It is added by [patches/extra/ungoogled-chromium/add-flag-to-show-avatar-button.patch](../tree/master/patches/extra/ungoogled-chromium/add-flag-to-show-avatar-button.patch)

sgtpep:
@PF4Public Exactly what I needed. Thank you so much!

Eloston:
Huh.

PF4Public:
Come to think of it, could we / should we probably change the name of this flag and reword its description, so that its purpose would be clearer and more obvious, for it to be easier to find? I do use this flag, which is obvious from a screenshot above, but if I didn't, I likely wouldn't be searching for an "avatar" or "button", I would have been searching for an "account", "user", "profile" etc.

qvint:
The upstream calls it "[avatar toolbar button](https://source.chromium.org/search?q=kAvatarToolbarButton)".

PF4Public:
And we may call it whatever we want:
https://github.com/Eloston/ungoogled-chromium/blob/4f574006aca69964995acfa0718e7097ce36c059/patches/extra/ungoogled-chromium/add-flag-to-show-avatar-button.patch#L27-L30

Eloston:
I'd be okay with adding to the title and description. The flag name itself should remain since it's more of a technical detail.

PF4Public:
> The flag name itself should remain since it's more of a technical detail.

Good point, some people might be already using it and changing the flag name will break this functionality for them.

I'll do a PR soon.

PR #1134, opened 2020-08-13
avatar -> avatar/people/profile

PF4Public:
Reference: #1126

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-to-show-avatar-button.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-show-avatar-button.patch
@@ -25,8 +25,8 @@
  #endif  // defined(OS_CHROMEOS)
  
 +    {"show-avatar-button",
-+     "Show avatar button",
-+     "Show avatar button in the browser toolbar", kOsDesktop,
++     "Show user account button",
++     "Show user account button in the browser toolbar", kOsDesktop,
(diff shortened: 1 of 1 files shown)


qvint:
As per the project objective №2 (see the README),
I believe we should use the upstream terminology and be consistent with the existing flags.

----

PF4Public:
> I believe we should use the upstream terminology and be consistent with the existing flags.

And answer the same questions again and again due to a poor wording choices made upstream? @Eloston himself couldn't find this flag!

Eloston:
> @Eloston himself couldn't find this flag!

To be fair, I was being negligent and rushing things.

Anyway, I think @qvint makes a good point. Could you change the description to instead show "avatar/user button"? This has the additional benefit of working with `chrome://flags`'s "Search flags" function when you search "avatar".

PF4Public:
Should it be then `Show avatar / user account button` or `Show avatar / user button`?

Eloston:
Actually, I noticed that `chrome://settings` and the avatar menu calls profiles "people", not "users". So maybe we should call it `Show avatar/person button`? Or maybe `Show avatar/people/profile button` since the support docs still use the term "profile" (https://support.google.com/chrome/answer/2364824)?

([At one point it was called "user"](https://www.chromium.org/user-experience/multi-profiles), but I don't see that term used to describe the current iteration of the profile button)

PF4Public:
As funny as it gets 🤣

PR #1138, opened 2020-08-15
Added ability to always open dev context menu on reload button.

23rd:
This small PR adds ability to use the dev context menu on the reload button even in a non-dev mode.
The reload button does not have any context menu by default, so I don't see any reason to complicate my commit with a command-line flag.
<details>

image

</details>


Eloston:
The patch is failing validation, could you refresh it?

23rd:
@Eloston 👌

Eloston (approved):
Thanks!

PR #1152, opened 2020-08-23
Disable DIAL repeating discovery

### LINKED ISSUE #991: What's with hammering port 1900.
What is with ugC hammering port 1900. I have set the media router to disabled and turned off "hide local ip with mDNS". I still get literally hundreds of broadcasts a day while chrome is open.

Ahrotahn:
The DIAL registry has a timer set to send out SSDP broadcasts every two minutes.  
This removes the timer to prevent network spam.  
Fixes #991

### DIFF
--- a/patches/extra/ungoogled-chromium/disable-dial-repeating-discovery.patch
+++ b/patches/extra/ungoogled-chromium/disable-dial-repeating-discovery.patch
@@ -1,0 +1,16 @@
+# Disables the dial registry's repeating discovery timer
+# This caused unnecessary SSDP network spam
+
+--- a/chrome/browser/media/router/discovery/dial/dial_registry.cc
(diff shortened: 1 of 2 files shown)


Ahrotahn:
The design doc makes it seem like this belongs in **core**.  Let me know if it would better to put this in **extra** instetad, or if it should be rolled into an existing patch.

Eloston:
Could we put this under extra? It's not very clear whether this should be core or extra, but we already have `fix-building-without-mdns-and-service-discovery.patch` in extra. Also while this is a privacy-enhancing change, it doesn't do so by contacting a web service directly; it's for peer-to-peer connections. So this falls into a grey area at best.

Ahrotahn:
Makes sense to me.  I've moved it over to extra.

Eloston (approved):
LGTM, Thanks :+1:

PR #1153, opened 2020-08-23
Revise Debian/Ubuntu OBS instructions

braewoods:
This should resolve the problem of people using the wrong set of instructions.


Eloston (approved):
Thanks

Issue #1130, opened 2020-08-12
Remove Sync (ie: "Not Signed In", "Turn on sync...", "Sync and Google services", etc) UI 


prurigro:
**Is your feature request related to a problem? Please describe.**

This isn't really a problem so much as that the UI is showing options for and suggesting the use of sync functionality which has been disabled.

**Describe the solution you'd like**

The package would feel much more complete if the account dropdown didn't have a broken sync icon, the message `Not signed in` and a big `Turn on sync` bubble. The `Sync and Google Services` menu and other settings that require being signed into a google account in the settings are much less in your face but is also a bit redundant.

**Describe alternatives you've considered**

I'm not sure if there's really an alternative way to remove something, but the dialogs in the account dropdown are definitely the most jarring.

**Additional context**

![redundant-sync-dropdown](https://user-images.githubusercontent.com/1149238/90060466-c3790880-dcb2-11ea-916d-7b9bfa1964bc.png)

![redundant-sync-setup](https://user-images.githubusercontent.com/1149238/90060469-c5db6280-dcb2-11ea-81c6-068c8debb872.png)

wchen342:
Should this be merged with #224, #369, #999 to be one meta issue?

Eloston:
Good idea, I'll add a meta issue to track these kinds of problems.

LoganDark:
Looks like "Not signed in" is present again on Ungoogled Chromium 129

![image](https://github.com/user-attachments/assets/44913db3-3d63-4d8f-b61f-84ed6324e5b7)

could this issue be reopened?

Issue #369, opened 2018-03-25
Remove unused toggles for Google services from the UI


irlcatgirl:
Certain settings toggles are displayed which could harm privacy if turned on by mistake. It might be better if they are patched out. Thoughts?
<img width="690" alt="screen shot 2018-03-25 at 6 21 56 pm" src="https://user-images.githubusercontent.com/8227641/37880612-70253260-3059-11e8-97fd-837ad120b59a.png">

Eloston:
It's fine if someone wants to patch it out for aesthetics, but they won't do anything if switched on.

Eloston:
It should be enough to add the hidden attribute to the entries that should be removed. You can submit a PR if you want.

PR #1156, opened 2020-08-26
UI and translation functionality changes

### LINKED ISSUE #224: Remove non-functioning Google Print UI
Came across this application today and decided to use going forward. Is there a reason as to why the Google Print menu under Settings still exists?

Settings -> Show advanced settings...

Personally I would like this to be removed.

Running on Windows 7 x64.

### LINKED ISSUE #999: Remove "Translate to ..." context menu entry and other UI
**Describe the bug**
Right-click on any web page + "Translate to ____ (English in my setup)" does not work

**To Reproduce**
1. Open any web page
2. Right click anywhere on the page, choose "translate to ___ (English appears autmatically in my setup)
3. See error "This page could not be translated"

**Expected behavior**
The page should get translated

 - OS/Platform and version:
 - ungoogled-chromium version: 80.0.3987.149 / Windows 7 (64-bit)

### LINKED ISSUE #1131: Removing dead Google web service-specific UI
This is a meta-issue that tracks issues regarding the removal of unused/dead UI as a result of removing Google web service integration.

Currently: #224, #369, #999, #1130

Ahrotahn:
Fixes #1131 

The first commit removes unneeded UI elements.
Fixes #224, #369, and #1130 

The second commit allows translation to function if the `--translate-script-url` and `--translate-security-origin` flags are set.  
It also removes the "Translate to" context menu item by default when the flag is not set.  
This removes two existing patches:
`core/inox-patchset/0014-disable-translation-lang-fetch.patch`
`core/ungoogled-chromium/disable-translate.patch`
Fixes #999 

The third commit re-adds the `core/ungoogled-chromium/disable-translate.patch` and implements the the changes from `core/inox-patchset/0014-disable-translation-lang-fetch.patch`.  The new implementation now checks for the `--translate-script-url` flag and disables the functionality when not present.  
This commit also separates the component download prevention into a new patch:  `core/ungoogled-chromium/block-component-downloads.patch`.


PF4Public:
> This removes two existing patches:
> `core/inox-patchset/0014-disable-translation-lang-fetch.patch`
> `core/ungoogled-chromium/disable-translate.patch`

Are there any downsides due to this?

Ahrotahn:
None that I can see.  The URL substitution prevents it from functioning and there doesn't seem to be any visible changes in how it behaved previously.  

It would be good to have some more people testing the change before it's accepted to be sure there aren't any side-effects I'm missing.

PF4Public:
I mean, my concern is that removing these patches could open some holes. Wouldn't it?

Ahrotahn:
I agree with you.  After giving it more thought, it's bad form to rely on the substitution like that.  

I'm working on a third commit for this PR that will check the translate URL flag and disable the features if there is no intention for them to be used.  It'll work much like what I already set up to hide the translate UI.

Ahrotahn:
I've added the third patch.  It now will now disable the translation functionality if the flag is not set.  

I also separated out the parts of the patches that prevent downloads into it's own patch.  I figured it would be a good idea to have that in it's own patch so that in changing or removing things in the future doesn't have a chance of affecting whether or not the browser attempts to download things.  It could also serve as a good place for future updates that may try to download new components.

Eloston (approved):
This is some nice work. Thanks!

Gonna merge this so more people can test it. I am considering adding a "staging" or "devel" branch in the future because testing is becoming increasing more important and difficult.

Eloston:
Actually I probably should've merged this after #1159. Sorry about that.

Eloston:
@Ahrotahn Please resubmit this PR after #1159. Apologies again for the inconvenience.

PF4Public:
> I am considering adding a "staging" or "devel" branch in the future because testing is becoming increasing more important and difficult.

Should we probably agree on a some kind of feature-freeze prior to major releases, so that no pull-request could spoil next release? So that @Eloston won't be merging any PR for the week before new major release, keeping them for the time after all the patches updated for a new version and it confirmed to be OK. After that we may have more time (till next release at-least) to test these pending PR's with new version?

Ahrotahn:
> @Ahrotahn Please resubmit this PR after #1159. Apologies again for the inconvenience.

No worries, I should be the one apologizing for submitting this so soon before a major update.  
I'll resubmit the UI changes and the translation changes separately.  I should have done that the first time around.  

I like @PF4Public's idea of implementing a freeze before the major releases.  They have a schedule [here](https://www.chromium.org/developers/calendar) that can help with deciding cut-off dates.

Eloston:
I'm not sure if a feature freeze is really necessary here. Consider the following observations:

* An issue is created roughly about the same time that a maintainer chooses to start updating to a new version
* All PRs are tested (to an extent) before they are merged. Major version update PRs usually get tested more than others.
* If multiple PRs are open simultaneously, then major version PRs take precedence.
* Any enhancing feature (i.e. code under `patches/extra`) will be dropped during a major update if the developer cannot update it.

So even if a PR is merged the day before a developer starts updating ungoogled-chromium to a new major version, there shouldn't be any problems because:

1. `core` patches are pretty stable and don't change much except during major version updates. So any PRs for core will probably not break anything
2. `extra` patches can be dropped if the developer cannot update it or it breaks something.

Speaking of `extra`, the number of patches it'll have will only grow over time. I'm concerned that this will only become an increasing burden for developers who update the patches to maintain. Maybe it'd help to have a system that notifies other developers to update them at a later time?

PR #1175, opened 2020-09-07
Translation functionality changes

Ahrotahn:
This PR adds a check for the `--translate-script-url` flag to the existing translation changes and also hides the "Translate to" context menu when it is not set.  This allows users to enable translation functionality as needed if a URL and security origin are supplied.  

The changes were significant enough that incorporating `core/inox-patchset/0014-disable-translation-lang-fetch.patch` into `core/ungoogled-chromium/disable-translate.patch` made sense, and the changes that prevented the ranker model download and spellcheck dictionary download have been moved into `core/ungoogled-chromium/block-requests.patch`.  

Addresses #1131, Fixing #999


patches/core/ungoogled-chromium/disable-translate.patch
```diff
@@ -1,38 +1,111 @@
...
---- a/components/translate/content/renderer/translate_agent.cc
-+++ b/components/translate/content/renderer/translate_agent.cc
-@@ -74,8 +74,7 @@ TranslateAgent::TranslateAgent(content::
-     : content::RenderFrameObserver(render_frame),
-       world_id_(world_id),
-       extension_scheme_(extension_scheme) {
--  translate_task_runner_ = this->render_frame()->GetTaskRunner(
--      blink::TaskType::kInternalTranslation);
-+  CancelPendingTranslation();
+# Disables translation and removes the "Translate to" context menu when --translate-script-url flag is not set
```
Eloston, line 1: This patch filename and this comment here is a little confusing because we don't always disable translation. Maybe we could rename the patch to something like `toggle-translation-via-switch.patch` or something to that effect?

Eloston (changes requested):
Code looks great, but minor complaint with naming.

patches/core/ungoogled-chromium/disable-translate.patch
Ahrotahn: Makes sense to me!

Eloston (approved):
LGTM, Thanks :+1:

PR #1186, opened 2020-09-15
Update ipv6 probing patch to use feature flag

### LINKED ISSUE #1172: Resolving host takes too long compared to other browsers or original chromium
**Describe the bug**
Accessing a website that has not been opened yet in the current session always takes 1-2 seconds before it starts loading making the browsing experince terrible. The status bar at the bottom of the browser says "Resolving host" during this wait, so my assumption is that there're something slowing DNS queries down. 

What I've tried:
- --set-ipv6-probe-false
- removing ~/.config/chromium dir for a fresh install
- other browsers work just fine
- even original Chromium works ok
- different networks
- changing DNS servers

**To Reproduce**
As I couldn't find any open issue that would describe my problem, I assume it is for some reason specific to my machine. I'm not aware of anything special on my computer however. I'm not even using a system-wide DNS cache. No IPv6.

 - OS/Platform and version: Linux version 5.8.5-arch1-1, Arch Linux, i3
 - ungoogled-chromium version: Version 84.0.4147.125 (Official Build) Arch Linux (64-bit)

--
**Could anyone push me in the right direction so that I can figure out what's the problem, please?**

Ahrotahn:
At some point the code this patch affects was moved off the main thread and no longer has access to the commandline switches.  
This caused `HasSwitch("set-ipv6-probe-false")` to always return false.  

Updating the patch to use a feature flag instead solves the issue while keeping the changes minimal.  
To make use of the change, the executable will now need to be passed `--enable-features=SetIpv6ProbeFalse` instead of `--set-ipv6-probe-false`.

Fixes #1172


Eloston (approved):
LGTM. Interesting that the feature flag has the simplest implementation here. Thanks for documenting it too!

PR #1189, opened 2020-09-19
Add flag to set whether a window should close with the last tab

Ahrotahn:
This PR adds a flag that allows the user to set whether a window should close when the last tab is closed.  
Closing the last tab will instead open a new tab when `close-window-with-last-tab` is set to `never`.  The window can still be closed via the menu, the titlebar, keyboard shortcut, etc.  

I was previously using an addon to handle this.  Unfortunately it needed to keep a blank tab pinned to function, polluted the tab history, and it also prevented dragging tabs between windows.  
It made more sense to add a few lines of code to implement the functionality.


Eloston (changes requested):
Huh, interesting use-case. Code LGTM!

Could you also document this in `docs/flags.md`? Thanks!

Ahrotahn:
Sure thing!  I've updated the commit.

Eloston (approved):
LGTM, Thanks :+1:

snacktary:
Is it possible to close the whole chromium application after closing the last tab?

Issue #1205, opened 2020-10-07
Updating to Chromium 86.0.4240.75


PF4Public:
## Notes for the developer

Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:

![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

Finally, make sure to reference this issue in your PR.

Feel free to ask questions and discuss issues here along the way. Thanks!

## Notes for others

Feel free to raise issues or questions throughout the process. However, please refrain from asking ETAs unless no visible progress has been made here or on the developer's PR for a while (e.g. 2 weeks).

Ahrotahn:
Odd to see no takers on this yet, hopefully everyone's just busy and doing well!

I had been working on a patch that deals with flag offsets and decided to try working on this to see what's involved and how that patch would affect the process.  I'll set up a draft if no one more experienced starts work on this by the time I get that done.

lipici:
maybe @tangalbert919 manages to pull this again and help us.

Zoraver:
I am doing well, I just had a very busy week. I am going to try to get the patches ported over the weekend.

Ahrotahn:
Life takes precedence, please don't feel rushed!  

I don't know if most major version upgrades require as much work as this one, but doing this has given me a new perspective on the process and some ideas about how it could be made easier.  I certainly appreciate the work you and the other maintainers put into this project.

PR #1219, opened 2020-10-21
New version check

PF4Public:
This PR adds a GutHub workflow, which checks for a new Chromium version and creates an "Updating to …" issue.

I allowed it to run [for a while](https://github.com/PF4Public/ungoogled-chromium/actions) in [my fork](https://github.com/PF4Public/ungoogled-chromium/issues?q=is%3Aissue+), while improving it.

Addressed concerns are:
* Running every 6 hours, not a heavy one
* Does not by default run in forks and [it is indeed so](https://github.com/PF4Public/ungoogled-chromium/actions), now that I have changed `if: github.repository ==` condition
* `check-if-issue-exists-action` didn't originally consider closed issues, which is a show-stopper for this use-case. I have however fixed that and [created a PR](https://github.com/nickderobertis/check-if-issue-exists-action/pull/1), which is unfortunately not yet accepted, hence `PF4Public/check-if-issue-exists-action@master`.
* It skips as much as possible if nothing changed/nothing to do
* To avoid the [confusion](https://github.com/Eloston/ungoogled-chromium/pull/1210#issuecomment-709610422), when a new issue was created by this workflow in my fork, which I took for granted for all platforms, but stable was only bumped on Mac, it also provides a link to the original news item

I believe it is now ready to be merged

PS: after merging it should (within 6 hours) create [Updating to Chromium 86.0.4240.111](https://github.com/PF4Public/ungoogled-chromium/issues/15)

### DIFF
--- a/.github/workflows/new_version_check.yml
+++ b/.github/workflows/new_version_check.yml
@@ -1,0 +1,40 @@
+name: New version check
+
+on:
+  # running every 6 hours
+  schedule:
+    - cron: '48 */6 * * *'
+jobs:
+  check:
+    # do not run in forks
+    if: github.repository == 'Eloston/ungoogled-chromium'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Read news items
+        id: rss-out
+        # a little bit of Shell-Fu
+        # obtaining latest version and a link to a corresponding news item
+        run: echo "::set-output name=version::$( rdom () { local IFS=\> ; read -d \< E C ; }; while rdom; do if [[ $C = 'Stable Channel Update for Desktop' ]]; then while rdom; do if [[ $C =~ [[:space:]]([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)[[:space:]] ]]; then echo ${BASH_REMATCH[1]}; while rdom; do if [[ $E = 'feedburner:origLink' ]]; then echo "::set-output name=link::$C"; exit; fi done fi done fi done < <(wget -q -O- 'https://chromereleases.googleblog.com/feeds/posts/default') )"
+      - uses: PF4Public/check-if-issue-exists-action@master
+        name: Check if Issue Exists
+        id: check_if_issue_exists
+        with:
+          repo: Eloston/ungoogled-chromium
+          token: ${{ secrets.GITHUB_TOKEN }}
+          title: Updating to Chromium ${{ steps.rss-out.outputs.version }}
+          labels: enhancement, help wanted
+      - uses: actions/checkout@v2
+        # we need this to use create-an--updating-to-chromium-x-x-x-x-.md
+        # skip it if there's nothing new
+        if: steps.check_if_issue_exists.outputs.exists == 'false'
+      - name: Create Issue
+        # skip it if there's nothing new
+        if: steps.check_if_issue_exists.outputs.exists == 'false'
+        uses: JasonEtco/create-an-issue@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # variables from Shell-Fu step
+          VERSION: ${{ steps.rss-out.outputs.version }}
+          LINK: ${{ steps.rss-out.outputs.link }}
+        with:
+          filename:  .github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
--- a/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
+++ b/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
@@ -1,13 +1,16 @@
 ---
 name: Create an "Updating to Chromium x.x.x.x"
 about: For letting the community track progress to a new stable Chromium
-title: Updating to Chromium [VERSION_HERE]
+title: Updating to Chromium {{ env.VERSION }}
 labels: enhancement, help wanted
 assignees: ''
 
 ---
 
-### Please respond if you would like to update ungoogled-chromium to the new stable Chromium version.
+Chromium stable channel has been updated to a newer version. See [the original posting]({{ env.LINK }}) for more information.
+Feel free to ask questions and discuss issues here along the way. Thanks!
+
+### Please respond if you would like to update ungoogled-chromium to the new stable Chromium version
 
 ## Notes for the developer
 
@@ -15,9 +18,7 @@ Once you claim it, it is advisable that you create a [Draft Pull Request](https:
 
 ![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)
 
-Finally, make sure to reference this issue in your PR.
-
-Feel free to ask questions and discuss issues here along the way. Thanks!
+Finally, make sure to reference this issue in your PR
 
 ## Notes for others
 


.github/workflows/new_version_check.yml
```diff
@@ -0,0 +1,40 @@
...
+  check:
+    # do not run in forks
+    if: github.repository == 'Eloston/ungoogled-chromium'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Read news items
+        id: rss-out
+        # a little bit of Shell-Fu
+        # obtaining latest version and a link to a corresponding news item
+        run: echo "::set-output name=version::$( rdom () { local IFS=\> ; read -d \< E C ; }; while rdom; do if [[ $C = 'Stable Channel Update for Desktop' ]]; then while rdom; do if [[ $C =~ [[:space:]]([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)[[:space:]] ]]; then echo ${BASH_REMATCH[1]}; while rdom; do if [[ $E = 'feedburner:origLink' ]]; then echo "::set-output name=link::$C"; exit; fi done fi done fi done < <(wget -q -O- 'https://chromereleases.googleblog.com/feeds/posts/default') )"
```
qvint, line 17:
```bash
curl -s 'https://chromiumdash.appspot.com/fetch_releases?channel=Stable&platform=Linux' |
    jq -r '.[0].version'
```

PF4Public: Does it provide a URL to news item?

qvint: You are trying to guess the latest version from the blog which is written by humans and for humans. It is not in a machine-friendly format. Also there were no links in previous issues: #1205 #1194 #1179.

PF4Public: Since when is xml not machine-friendly?

Ahrotahn: It would be more correct to use Google's [OmahaProxy](https://omahaproxy.appspot.com/) site rather than chromiumdash.  There are feeds for the current releases and the release history near the bottom.  You can even get a list of the commits since the last version there which could be added to the issue.  
The current version for a givien platform can be obtained simply by appending it to the url, for example:  https://omahaproxy.appspot.com/linux
There's less to parse and less chance for errors that way. 

The blog gives a good overview for what was changed in a given update, and would be nice to have included in the issue.  I think the way PF4Public implemented is fine since Google's FeedBurner seems to be the cleanest way to get the blog link for a version update.  Not only that, but it's actually a simpler script since the feed should update for each platform.  
Recently version 86.0.4240.80 was released, but it was a mac-only release.  So using a linux-specific version link would not have caught that version update.  This means that the action would become more complicated since it needs to check the version for each platform in order to be correct, whereas checking the blog feed should catch every stable update.  

If there isn't a delay between when a new version is released and the blog feed is updated, then I think PF4Public's implementation seems to be the best way to handle this.

qvint: Since ungoogled-chromium has the [official Android platform repo](https://github.com/ungoogled-software/ungoogled-chromium-android), "updates for desktop" might not be sufficient. Right now the latest Android release is 86.0.4240.114 (and not 86.0.4240.111). To handle that, it's possible to use the following simple script:

```bash
for p in Linux Windows Mac Android; do  # for all ungoogled-chromium platforms
  curl -s "https://chromiumdash.appspot.com/fetch_releases?channel=Stable&platform=$p&num=1"
done | jq -rs 'flatten | sort_by(.time) | .[-1].version'
```

> You can even get a list of the commits since the last version there

Since we already have the latest Chromium release in chromium_version.txt, the link is just `"https://chromium.googlesource.com/chromium/src/+log/${OLDVER}..${NEWVER}"`. There's no need to consult "OmahaProxy" here. (BTW Chromium devs do use Chromium Dash along with "OmahaProxy" [[1]](https://www.chromium.org/developers/calendar#:~:text=https://chromiumdash.appspot.com/schedule)[[2]](https://chromium.googlesource.com/devtools/devtools-frontend/+/HEAD/docs/workflows.md#:~:text=https://chromiumdash.appspot.com/branches).)

> The blog gives a good overview for what was changed in a given update,
> and would be nice to have included in the issue.

I agree, but the human-written blog is not a good source for automated version check. The link could be *optionally* added *after* obtaining the latest version,—by searching the *exact* version string over the blog.

PF4Public: > "updates for desktop" might not be sufficient

Like you said: … there were no ~links~ Androids in previous issues … ;)

Besides, "official Android platform repo" could use a similar workflow, why not!

> To handle that, it's possible to use the following easy-to-maintain script

How does it handle it? It returns 4 version strings, what next? 4 issues? An issue each time a single version changes? A combination?

I offer a _complete_ solution to the problem I see. You are suggesting some _snipplets_, which are not solving anything, which need more work to make use of.

Lets start over again, shall we?

>You are trying to guess the latest version

False

>written by humans

To me it looks like an automated blog, populated via templates. 

>not in a machine-friendly format

this "not machine-friendly format" is perfectly parsed by Bash alone, isn't it?

I don't get it, are you saying that this approach is wrong or simply suggesting something else? How _exactly_ is my approach wrong?

qvint: > It returns 4 version strings, what next? 4 issues?

What version of `jq` do you use? On my system `jq-1.6-4.fc32` returns exactly one version string from the [stream](https://stedolan.github.io/jq/manual/#:~:text=--slurp) of four JSON documents.

> need more work to make use of

Reading JSON using `jq` is much simpler, safer and *easier to maintain* comparing to [bash-based ad-hoc XML parsing](https://stackoverflow.com/questions/893585/how-to-parse-xml-in-bash/7052168#:~:text=Just%20because%20you%20can%20write%20your%20own%20parser,%20doesn't%20mean%20you%20should.). Also, the blog posts contain [a lot of redundant styling tags](https://user-images.githubusercontent.com/16843577/97010579-41c51700-154e-11eb-8963-3ff059b0b959.png) which could mean that the texts are copy-pasted from MS Office or similar software.

PF4Public: >Also, the blog posts contain a lot of redundant styling tags which could mean that the texts are copy-pasted from MS Office or similar software.

I'm not parsing blog posts, I'm parsing xml feed.

>Reading JSON using jq is much simpler, safer and easier to maintain comparing to bash-based ad-hoc XML parsing solutions.

Please be more specific on the issues, that may arise from parsing xml?

qvint: > Please be more specific on the issues, that may arise from parsing xml?

Being edited by a human with a WYSIWYG-editor, the produced HTML can easily break your XML reader. E.g., does your script handle something like this—?

```html
<span>... updated to 86</span><span>.0.4240.111 for Windows ...</span>
```

It will not work for [this post](https://chromereleases.googleblog.com/2019/07/stable-channel-update-for-desktop_30.html) just because the version is surrounded by non-`[[:space:]]` characters:

```html
<b>76.0.3809.87</b>      // or, in XML feed:  &lt;b&gt;76.0.3809.87&lt;/b&gt;
```

It will not work if there is an `&nbsp;` right after the version ([see the post](https://chromereleases.googleblog.com/2020/07/stable-channel-update-for-desktop.html)):

```html
84.0.4147.89&nbsp;       // or, in XML feed:  84.0.4147.89&amp;nbsp;
```

I don't think we should rely on WYSIWYG-editor outputs here. Parsing HTML with bash is too error-prone.

PF4Public: >error-prone

Worst case - missing "Updating to…" issue. How bad is that?

qvint: > Worst case - missing "Updating to…" issue

You *will* create an issue for the *next* occurrence of `[[:space:]]([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)[[:space:]]`.

If `86.0.4240.111` matching fails, you will notify the others that it's time to update to… **`88.0.4292.2`**.

PF4Public:
Manual issue creation it is then.

PR #1233, opened 2020-11-07
Fix gn conflict with Android

wchen342:
The component name used for ungoogled_switches.h will conflict in v86 on Android:
```
ERROR at //build/config/BUILDCONFIG.gn:576:5: Duplicate output file.
    target(_target_type, _target_name) {
    ^-----------------------------------
Two or more targets generate the same output:
  libswitches.cr.so

This is can often be fixed by changing one of the target names, or by 
setting an output_name on one of them.

Collisions:
  //components/ungoogled:switches
  //device/gamepad/public/cpp:switches
```

This should solve the problem.


wchen342:
@Eloston Also I accidentally forked the main repo into `ungoogled-software` but I do not have permission to delete it. Can you do it?

Eloston:
> @Eloston Also I accidentally forked the main repo into `ungoogled-software` but I do not have permission to delete it. Can you do it?

Done.

Eloston (approved):
LGTM, Thanks :+1:

PR #1250, opened 2020-11-17
Flag overhaul

Ahrotahn:
I noticed the TODO comment about adding a flag to `chrome://flags` in `add-ipv6-probing-option.patch` while working on a previous change.  I decided to delay adding the flag and create it's own PR, which has since evolved into this one!  

I've made this a draft for now to get feedback on the implementation and to make sure that this doesn't cause problems for anyone.  I also need to set up a companion PR for the Windows-specific flags.  Not to mention the major version update due in a day or so will require this to be updated anyways.  

# Overview
* Creates flag entries for patches that did not have them previously.  Now all features have `chrome://flags` entries!
* Added "ungoogled-chromium flag" to the end of all flag descriptions.  This makes it easier to differentiate between Chromium flags and ones that have been added by ungoogled-chromium.  It also has the benefit of being able to pull up all the ungoogled-chromiumn flags when searching on the flag page.  
* Updates the flag documentation.  I organized the flags by the platform it's available for and used a table layout for readability.  ([preview](https://github.com/Eloston/ungoogled-chromium/blob/9cdae7734149dfac08dcfc421a616e0b81cad6c5/docs/flags.md))  
* Adds a patch which creates headers to organize flags added by Ungoogled Chromium.  This is where the most significant change happens and will be explained in more detail below.  
* Updates existing patches to use the new headers.  
* Fixes a small bug in `validate_patches.py` that this processes uncovered.

### DIFF
--- a/devutils/validate_patches.py
+++ b/devutils/validate_patches.py
@@ -525,7 +525,7 @@ def _apply_file_unidiff(patched_file, files_under_test):
         assert len(patched_file) == 1 # Should be only one hunk
         assert patched_file[0].removed == 0
         assert patched_file[0].target_start == 1
-        files_under_test[patched_file_path] = [x.value for x in patched_file[0]]
+        files_under_test[patched_file_path] = [x.value.rstrip('\n') for x in patched_file[0]]
     elif patched_file.is_removed_file:
         # Remove lines to see if file to be removed matches patch
         _modify_file_lines(patched_file, files_under_test[patched_file_path])
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -1,31 +1,52 @@
 # List of flags and switches
 
-This is an exhaustive list of command-line switches and `chrome://flags` introduced by ungoogled-chromium
+This is an exhaustive list of command-line switches and flags introduced by ungoogled-chromium.  
+Each switch has a corresponding entry on the `chrome://flags` page which can be filtered by searching for `ungoogled-chromium`.  
 
-**NOTE**: If you add a command-line argument that is also in `chrome://flags`, the flag's state will not be indicated in `chrome://flags`. There is no universal way to ensure command-line flags are taking effect, but you can find if they're being seen by checking `chrome://version`.
+If a switch requires a value, you must specify it with an `=` sign; e.g. flag `--foo` with value `bar` should be written as `--foo=bar`.
 
-If a flag requires a value, you must specify it with an `=` sign; e.g. flag `--foo` with value `bar` should be written as `--foo=bar`.
+> **NOTE**: If you add a command-line argument that is also in `chrome://flags`, the flag's state will not be indicated in `chrome://flags`. There is no universal way to ensure command-line flags are taking effect, but you can find if they're being seen by checking `chrome://version`.
 
-* `--bookmark-bar-ntp` - Sets the visibility of the bookmark bar on the New Tab Page. Only takes the value `never`.
-* `--close-window-with-last-tab` - Determines whether a window should close once the last tab is closed.  Only takes the value `never`.  Also configurable under `chrome://flags`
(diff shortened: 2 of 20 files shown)


Ahrotahn:
# Further Detail

## Documentation Formatting
Github's formatting tries to make the flag column narrow which causes the flags to break over multiple lines on the hyphens.  I used non-breaking spaces in the headers to prevent that from happening.  I'd be interested to know if there's a better way to handle that!  

## Flag Headers
Currently flag entries and choices are all added into `chrome/browser/about_flags.cc` without too much organization.  Updating or adding new flags generally requires updating the line offsets in other patches.  And when Google changes things in `about_flags.cc` each patch will need to be double-checked afterwards if `quilt refresh` had fuzz updates.  

This PR organizes all of the flag entries and choices into separate headers.  Those headers are then included in `about_flags.cc`.  
The idea is that only the line offset for `add-ungoogled-flag-headers.patch` would be updated when running `quilt refresh`!  Adding new flags would also be an easier process since it could simply be added after the last one in its header.  
The Bromite flags have their own set of headers separate from the Ungoogled Chromoium ones.  This is so PRs adding new flags to Ungoogled Chromium don't also need to update the offsets in the Bromite patches.  

This creates new files, so if you do dirty builds then you would need to handle the new files the same way you do for `add-components-ungoogled.patch`.  

An alternative approach would be to have the entries and choices be included in the `add-ungoogled-flag-headers.patch` itself.  That would make the list more apparent to anyone looking at the patch and would allow the series file to be reordered more easily as well.  
I didn't go with this option since I thought it would be weird to have the patches and their flags decoupled.  It also means that flags like the Windows-specific ones would need to make a PR to add their flags to the list.  Granted they would have to make a PR to add them to the flag doc as well, so maybe that isn't really a problem.  

## Bugfix for `validate_patches.py`
I managed to find a small bug while working on this.  Existing files are split on newlines, so there shouldn't be newline characters in the lines in the list that patches are compared against.  When a patch *creates a new file* the hunk is read in line-by-line but those lines included a newline!  This caused validation to fail for any patch that attempts to change a newly created file.

Eloston (commented):
This is some great work. Not only is it good at a technical level, your description makes this excellent by explaining the context for the non-trivial changes and design choices you've made. Thanks :)

I'll review again after it has been updated for 87.

Eloston:
@Zoraver I'd like your feedback on this PR since you've put in a lot of work into maintaining our patches.

Zoraver (commented):
@Ahrotahn I appreciate the strategy that you are using here. This will definitely make maintaining the flags that we add easier. The only minor nitpick that I have is that the updated `docs/flags.md` does not really make it clear that "all features have chrome://flags entries" - it reads like only `SetIpv6ProbeFalse` does.

Ahrotahn:
@Eloston I don't have a preference!  I guess that's just how the name occupies my mind and I didn't give it a second thought.  I'll change them to "ungoogled-chromium" when I do the update.

@Zoraver Going over doc again also just made me realize that I have "flags" for the table header, but most of them are actually the commandline switches.  I'll try to make it more clear that every entry has a corresponding flag and switch.

Ahrotahn:
I've updated the PR for 87 and hopefully improved the language on the flags doc.  

I'll mark this for review after I create the PR on the Windows repo later today.

Ahrotahn:
I created the PR and have also added one more change.  

Originally I had the Windows flags added to the ungoogled headers like the rest.  Having it that way meant that they would have to update their patches any time that the ungoogled-chromium repo adds/removes/reorders flag entries.  

So I added a set of platform headers that can be used by any of the platform repos to add entries to `chrome://flags`.  That should allow them to add their flags without having to worry about changes here causing extra maintenance.

Eloston (approved):
LGTM. Thanks for documenting your changes along the way!

Issue #1252, opened 2020-11-17
Updating to Chromium 87.0.4280.66


Ahrotahn:
## Notes for the developer

Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:

![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

Finally, make sure to reference this issue in your PR.

Feel free to ask questions and discuss issues here along the way. Thanks!

## Notes for others

Feel free to raise issues or questions throughout the process. However, please refrain from asking ETAs unless no visible progress has been made here or on the developer's PR for a while (e.g. 2 weeks).

## Additional Info
[Change Log](https://chromium.googlesource.com/chromium/src/+log/86.0.4240.0..87.0.4280.0?pretty=fuller&n=10000)

jstkdng:
Oh my, this is a big update afaik.

Zoraver:
I am working on this currently.

victor03303:
I even thought there was an issue with updating since I'm still on version 84.

PR #1261, opened 2020-11-21
Upgrade to Chromium 87

Zoraver:
Addresses #1252. I will let you know once I have a successful build. For update notes, view the commit messages.


Luka-Filipovic:
Did you encounter this error while building? It always popups early on.
```
[18/19200] LIBTOOL-STATIC obj/chrome/browser/safe_browsing/libsafe_browsing.a
FAILED: obj/chrome/browser/safe_browsing/libsafe_browsing.a 
rm -f obj/chrome/browser/safe_browsing/libsafe_browsing.a && TOOL_VERSION=1605638353 python ../../build/toolchain/mac/filter_libtool.py /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool -static -D  -o obj/chrome/browser/safe_browsing/libsafe_browsing.a -filelist obj/chrome/browser/safe_browsing/libsafe_browsing.a.rsp
fatal error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: file list file: obj/chrome/browser/safe_browsing/libsafe_browsing.a.rsp is empty
```
After that, build continues with no errors. I'm asking because I have some other patches I'm using as well but they shouldn't trigger this error.

Building on macos.

Ahrotahn:
Builds and runs fine so far on Arch.  Thanks Zoraver!  
<hr>
@Luka-Filipovic, I didn't encounter that error on my build.  I've seen some previously submitted issues with similar output.  Maybe the macos platform patches need to be updated?

Luka-Filipovic:
You are right, somehow I forgot to turn on patch for fixing safebrowsing build on macos. 
Anyways, build working good for me on macos. Will test on windows tomorrow.
Thanks!

jeromegn:
> Builds and runs fine so far on Arch. Thanks Zoraver!

@Ahrotahn How have you been able to build it? I'm trying to use [ungoogled-chromium-archlinux](https://github.com/ungoogled-software/ungoogled-chromium-archlinux) as a base, but none of the patches seem to be relevant anymore.

Ahrotahn:
The easiest way is to check out the changes foutrelis has made for the [Arch Chromium package](https://www.archlinux.org/packages/extra/x86_64/chromium/).  There is a link to the [source](https://github.com/archlinux/svntogit-packages/tree/packages/chromium/trunk) in the box on the side of the page.  Then you can update the base you have with the same changes.  

When it's early enough that you don't have the luxury of being able to cheat off of that then you would need to get a little more familiar with what the PKGBUILD is doing.  The first thing to check is [stha09's repo](https://github.com/stha09/chromium-patches) for patches that have been made that allow building with libstdc++.  Find the version you're building for on the releases page and expand the assets.  Set `_gcc_patchset` in your PKGBUILD to the same version number that is used there.  Swap out the patches in your PKGBUILD with those.  You don't need the gcc ones since this is being built with clang (unless you're building with gcc of course!).  Don't forget to set `pkgver` to the new version of Chromium you're building for and update the source URLs for Zoraver's repo and branch.

jeromegn:
@Ahrotahn Thank you! I eventually figured out just that. Still getting build failure much further out, but they might be due to something else.

perfect7gentleman:
```
Version 87.0.4280.66 (Official Build) (with ungoogled-chromium patches) (64-bit)
```

Zoraver:
My build succeeded and is running well; feel free to merge.

@Eloston Chromium 87.0.4280.67 is considered stable for macOS only. Would you like me to bump the version up to 87.0.4280.67 so that both 87.0.4280.66-1 and 87.0.4280.67-1 can be tagged? The macOS platform repo would then use 87.0.4280.67-1 and the other platform repos would use 87.0.4280.66-1.

mbakke:
It seems Ozone and Wayland support is built by default in this version.  Tested both X11 and `--enable-features=UseOzonePlatform --ozone-platform=wayland`, works fine on Guix, thanks!

@Ahrotahn Chromium has not been able to build with GCC for a long time.  What the GCC patchset does is enable compatibility for GCC's `libstdc++.so`. They can be omitted if you build Chromium with the bundled libcxx.

jstkdng:
@mbakke I've been waiting for this version to come, finally ungoogled-chromium-ozone can be taken down.

Eloston:
@Zoraver [The changes look pretty minimal between `.66` and `.67`](https://chromium.googlesource.com/chromium/src/+/refs/tags/87.0.4280.66..87.0.4280.67), I think it is fine to just tag `.67` this time.

Eloston (approved):
LGTM, I'll wait until we bump to `.67`

Eloston (approved):

Zoraver:
@Eloston no changes were required when changing from 87.0.4280.66 to 87.0.4280.67. Ungoogled-chromium 87.0.4280.67 compiled and is running well.

PR #1266, opened 2020-11-28
Split distribution-independent packages out to a separate README section.

mbakke:
Greetings,

I thought it would be good to clarify that Flatpak, Guix, and Nix can be used to obtain ungoogled-chromium on any GNU/Linux distribution.

This PR adjusts for that.

(the Nix package seems stuck on 81, but that's another issue)


Eloston (approved):
LGTM, Thanks :+1:

PR #1278, opened 2020-12-10
Performance improvements for update_lists.py

Ahrotahn:
Running `update_lists.py` for the last update took around 7m35s on average with my hardware.  I made two changes to improve the performance:

*  I changed the loops in `should_prune` to filters.  It seems that pathlib adds a bit of overhead that becomes more apparent when looped over multiple times.  This change on it's own was enough to reduce the script's run-time on my system to 4m54s!
*  I implemented multiprocessing to the script.  Chromium has a lot files that have to be processed, spreading the work out over multiple cores helps get that accomplished in less time.  This change brought the run-time down to 1m10s!

There's a handful of different ways the multiprocessing can be implemented, but I believe that this is one of the most readable ways to handle it in this instance.  

I made `source_tree` and `search_regex` into globals.  These weren't changed during runtime after being set by the args, so it made sense to me to make them global so they don't have to be passed to each process.  Let me know if you'd prefer to not have them set up that way.

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -15,6 +15,8 @@ import argparse
 import os
 import sys
 
+from itertools import repeat
+from multiprocessing import Pool
 from pathlib import Path, PurePosixPath
 
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent / 'utils'))
@@ -160,22 +162,23 @@ def _dir_empty(path):
     return False
 
 
-def should_prune(path, relative_path, unused_patterns):
+def should_prune(path, relative_path, used_pep_set, used_pip_set):
     """
     Returns True if a path should be pruned from the source tree; False otherwise
 
     path is the pathlib.Path to the file from the current working directory.
     relative_path is the pathlib.Path to the file from the source tree
-    unused_patterns is a UnusedPatterns object
+    used_pep_set is a list of PRUNING_EXCLUDE_PATTERNS that have been matched
+    used_pip_set is a list of PRUNING_INCLUDE_PATTERNS that have been matched
     """
     # Match against include patterns
     for pattern in filter(relative_path.match, PRUNING_INCLUDE_PATTERNS):
-        unused_patterns.pruning_include_patterns.discard(pattern)
+        used_pip_set.add(pattern)
         return True
 
     # Match against exclude patterns
     for pattern in filter(Path(str(relative_path).lower()).match, PRUNING_EXCLUDE_PATTERNS):
-        unused_patterns.pruning_exclude_patterns.discard(pattern)
+        used_pep_set.add(pattern)
         return False
 
     # Do binary data detection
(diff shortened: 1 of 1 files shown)


devutils/update_lists.py
```diff
@@ -205,84 +205,106 @@ def _check_regex_match(file_path, search_regex):
...
+    symlink_list = manager.list() # POSIX resolved path -> set of POSIX symlink paths
+
+    # Launch multiple processes iterating over the source tree
+    with Pool() as procpool:
+        procpool.starmap(
+            compute_lists_proc,
+            zip(
+                _source_tree.rglob('*'), repeat(used_pep_list), repeat(used_pip_list),
+                repeat(used_dep_list), repeat(used_dip_list), repeat(domain_sub_list),
+                repeat(pruning_list), repeat(symlink_list)))
```
Eloston, line 294: We both want to improve performance here, so I am wondering if `multiprocessing`'s Managers is the right synchronization mechanism to use. I'm looking a bit into Managers, but I'm still not clear on the details. For example, what underlying IPC mechanism do Managers use to share data?

But first, let's be clear on the higher-level design. Looking at the code, I believe these two invariants are true:

1. Each worker process uses data from the main process without modifying them.
2. Each worker process only needs to write to its output lists; it never needs to read its output.

Based on these invariants, we can conclude that **each worker process can work independently from each other**. 

Now let's consider your use of Manager. Whenever a worker modifies a list, the Manager needs to employ a locking mechanism to prevent a race condition with other worker processes, right? However, this is very inefficient because we just proved that worker processes don't need to share data among themselves (something that a Manager is useful for). In other words, workers should be free to run completely independently from each other (i.e. without synchronization) until they are ready to return their output.

Therefore, we should consider another synchronization mechanism. For example, can we simply return the lists at the end of each worker function and aggregate them in the main process?

devutils/update_lists.py
```diff
@@ -205,84 +205,106 @@ def _check_regex_match(file_path, search_regex):
...
+    used_pip_list = manager.list() # pruning_include_patterns
+    used_dep_list = manager.list() # domain_exclude_prefixes
+    used_dip_list = manager.list() # domain_include_patterns
+    domain_sub_list = manager.list()
+    pruning_list = manager.list()
+    symlink_list = manager.list() # POSIX resolved path -> set of POSIX symlink paths
+
+    # Launch multiple processes iterating over the source tree
+    with Pool() as procpool:
+        procpool.starmap(
```
Eloston, line 289: I believe the chunksize is 1 by default. We should probably increase this because the IPC overhead might be impacting performance significantly. You could try playing with a few different values, but we don't have to be super rigorous about this.

We could also hardcode this value for now, unless you think it's useful enough to have as a CLI argument (I personally can't think of any use-cases where this would be useful enough. Seems like we'd be micro-optimizing for the particular system we're on at that point.)

devutils/update_lists.py
```diff
@@ -205,84 +205,106 @@ def _check_regex_match(file_path, search_regex):
...
+    used_pep_list = manager.list() # pruning_exclude_patterns
+    used_pip_list = manager.list() # pruning_include_patterns
+    used_dep_list = manager.list() # domain_exclude_prefixes
+    used_dip_list = manager.list() # domain_include_patterns
+    domain_sub_list = manager.list()
+    pruning_list = manager.list()
+    symlink_list = manager.list() # POSIX resolved path -> set of POSIX symlink paths
+
+    # Launch multiple processes iterating over the source tree
+    with Pool() as procpool:
```
Eloston, line 288: This will create a process pool that uses all available threads on the system. We should probably make the number of workers customizable via CLI args, to follow the convention for multi-process tools.

Eloston (changes requested):
Thanks for investing your time into this. You've made some good progress; let's make it better.

Ahrotahn:
The multiprocessing Managers aren't the most efficient.  A separate process needs to be created just to manage the data for the proxies that are passed to the subprocesses.  I ended up choosing to use them due to a mixup.  The Pool map functions *are* able return the data returned from the the spawned processes.  I had tested a few different implementations and kept a log of the run-time for each, and I had 1m34s listed for iterating over the returned data.  I revisited it again after reading your comment and was averaging about 53s, so I must've made a mistake previously.  Once I get some time I'll update the commit with those changes.  I'll remove the globals and add an argument for limiting the number of processes.  That could be useful if you have something else running at the same time that you want to keep responsive.  

I thought the chunksize was 1 by default as well.  I had played around with different numbers and couldn't get better times than when using the default.  I dug a bit deeper to see what was going on under the hood and apparently the default chunksize is calculated by dividing the length of the iterator by four times the size of the pool ([link](https://github.com/python/cpython/blob/master/Lib/multiprocessing/pool.py#L477)).  That explains a lot!  It would be nice if that were mentioned in the python docs...

Ahrotahn:
I just updated the commit with the new changes.  Let me know if you'd prefer to have the new argument worded or ordered differently.  The same goes for the variable names, I try to keep them descriptive but I abbreviated the used sets since the names were getting a bit too long. 

This gets my run-time average down to 53s!  There is likely still room for improvement, but I'm pretty pleased to get that down to under a minute.

Eloston (approved):
LGTM! I love how thorough you've been in testing and implementing these changes; thanks for sharing some of your passion here :)

PR #1282, opened 2020-12-12
Patch distilled page to have a useable menu

toluschr:
### What does this do?
The gruvbox-dark colorscheme was missing from the distilled-page themes. I added it as 'gruv'.
I also noticed that the topmenu was completely glitched on my system (Probably due to Ungoogled Chromium substituting Google-Api-Urls).

The first patch / commit transforms it to look like this:
image

The second patch adds the gruvbox-dark colorscheme to the options: (https://android.stackexchange.com/questions/218970/enable-chromes-reader-view-manually-when-it-isnt-offered)
image

I might also do a followup PR that adds both an additional slider (width) and a flag to always show the distilled icon in the urlbar, since chromiums detection mechanism doesn't work well.

### DIFF
--- a/patches/extra/ungoogled-chromium/fix-distilled-icons.patch
+++ b/patches/extra/ungoogled-chromium/fix-distilled-icons.patch
@@ -1,0 +1,14 @@
+--- a/components/dom_distiller/core/css/distilledpage.css
++++ b/components/dom_distiller/core/css/distilledpage.css
+@@ -28,6 +28,11 @@
+   box-sizing: border-box;
+ }
+ 
++.material-icons {
++  font-family: Material Icons;
++  font-style: normal !important;
++}
++
+ /* Remove all margins and padding from certain element and add word wrap. */
+ 
+ blockquote,
--- a/patches/series
+++ b/patches/series
@@ -84,6 +84,7 @@ extra/ungoogled-chromium/add-flag-to-close-window-with-last-tab.patch
 extra/ungoogled-chromium/add-flag-to-convert-popups-to-tabs.patch
 extra/ungoogled-chromium/add-extra-channel-info.patch
 extra/ungoogled-chromium/prepopulated-search-engines.patch
+extra/ungoogled-chromium/fix-distilled-icons.patch
 extra/bromite/fingerprinting-flags-client-rects-and-measuretext.patch
 extra/bromite/flag-max-connections-per-host.patch
 extra/bromite/flag-fingerprinting-canvas-image-data-noise.patch


toluschr:
The gruvbox color doesn't really fit the contributing guidelines. Still, it's a nice feature.

patches/extra/ungoogled-chromium/add-gruvbox-distilled.patch
```diff
@@ -0,0 +1,300 @@
...
+-  if (theme == mojom::Theme::kDark)
+-    return kDarkCssClass;
++  if (theme == mojom::Theme::kLight)
++    return kLightCssClass;
+   if (theme == mojom::Theme::kSepia)
+     return kSepiaCssClass;
+-  return kLightCssClass;
++  if (theme == mojom::Theme::kGruv)
++    return kGruvCssClass;
++  return kDarkCssClass;
```
Eloston, line 246: I'm wondering: why does the Dark and Light CSS classes need to be rearranged here?

toluschr: I sorted it, so the order stays the same throughout the code. Light, Sepia, Gruv, Dark

patches/extra/ungoogled-chromium/add-gruvbox-distilled.patch
```diff
@@ -0,0 +1,300 @@
...
+       0);
++  EXPECT_EQ(
++      kSepiaJs.compare(viewer::GetDistilledPageThemeJd(mojom::Theme::kGruv)),
++      0);
++  EXPECT_EQ(
++      kDarkJs.compare(viewer::GetDistilledPageThemeJs(mojom::Theme::kDark)),
++      0);
+ }
+ 
+ TEST_F(DomDistillerViewerTest, TestGetDistilledPageFontFamilyJsOutput) {
```
toluschr, line 288: This could be removed.

toluschr:
@Eloston I deliberately made the first patch (bug fix) mergeable independently of the gruvbox patch.
The gruvbox colors were something I was working on, while fixing the bug and I threw it in, just in case there was a chance it gets merged.

Are you interested in merging the former? Else, I'll close the PR.

Eloston:
Oh yeah, we can merge the former. Thanks for reminding me.

I'm not sure about accepting the Gruvbox patch, so I'll be okay with adding it to [contrib](https://github.com/ungoogled-software/contrib) instead.

toluschr:
@Eloston Didn't have the time to edit the PR and forgot afterwards. I hope the changes are fine with you.

Eloston (approved):
LGTM, Thanks :+1:

Issue #1274, opened 2020-12-08
Android specific files


wchen342:
I have noticed this for some time but didn't have a chance to put it into a discussion. There are some binary files that are required to build Android version but are unnecessary on desktop versions, including
```
chrome/android/profiles/afdo.prof
third_party/icu/android/icudtl.dat
third_party/icu/android_small/icudtl.dat
third_party/icu/android_small/icudtl_extra.da
third_party/icu/common/icudtb.dat
third_party/jetifier_standalone/lib/jetifier-standalone.jar
```
.

For the moment I just used a patch to remove them from `pruning.list` when building Android version, but maybe it will be better just leave them in the first place if they are not a problem?

PR #1302, opened 2020-12-25
Remove android specific files from pruning list

wchen342:
Remove some files needed for Android from `pruning.list`.
Solves #1274.


Eloston (approved):
LGTM, Thanks :+1:

PR #1329, opened 2021-01-14
Exclude extra pruning file; add CentOS

wchen342:
Two minor changes:

* Exclude a file needed for Android
* Add CentOS to supported platforms


Eloston (approved):
LGTM!

Issue #1303, opened 2020-12-25
"about" (settings/help) should state this is ungoogled cromium


htgoebel:
As a user I want to verify I'm actually running the "ungoogled" version of chromium. The natruel place for this IMHO woulde be the "about" page, which in chromium is `chrome://settings/help`.

Example of possible solution:
Version 87.0.4280.88 (**Ungoogled** Build) (64-Bit)

PR #1332, opened 2021-01-16
Add extra channel info

wchen342:
Add "Ungoogled Chromium" as extra channel infomation. Solves #1303.

*nix version is tested on Fedora. Windows version is untested.

### DIFF
--- a/patches/extra/ungoogled-chromium/add-extra-channel-info.patch
+++ b/patches/extra/ungoogled-chromium/add-extra-channel-info.patch
@@ -1,0 +1,22 @@
+description: change extra channel info to "Ungoogled Chromium"
+author: Wengling Chen <feiyu2817@gmail.com>
+
+---
+ components/version_ui_strings.grdp |    4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+--- a/components/version_ui_strings.grdp
++++ b/components/version_ui_strings.grdp
+@@ -4,10 +4,10 @@
+     About Version
+   </message>
+   <message name="IDS_VERSION_UI_OFFICIAL" desc="official build on the about:version page">
+-    Official Build
++    Official Build, ungoogled-chromium
+   </message>
+   <message name="IDS_VERSION_UI_UNOFFICIAL" desc="unofficial build on the about:version page">
+-    Developer Build
++    Developer Build, ungoogled-chromium
+   </message>
+   <message name="IDS_VERSION_UI_32BIT" desc="32-bit on the chrome://version page">
+     (32-bit)
--- a/patches/series
+++ b/patches/series
@@ -84,6 +84,7 @@ extra/ungoogled-chromium/disable-dial-repeating-discovery.patch
 extra/ungoogled-chromium/remove-uneeded-ui.patch
 extra/ungoogled-chromium/add-flag-to-close-window-with-last-tab.patch
 extra/ungoogled-chromium/add-flag-to-convert-popups-to-tabs.patch
+extra/ungoogled-chromium/add-extra-channel-info.patch
 extra/bromite/fingerprinting-flags-client-rects-and-measuretext.patch
 extra/bromite/flag-max-connections-per-host.patch
 extra/bromite/flag-fingerprinting-canvas-image-data-noise.patch


patches/extra/ungoogled-chromium/add-extra-channel-info.patch
```diff
@@ -0,0 +1,30 @@
...
+ 2 files changed, 2 insertions(+), 2 deletions(-)
+
+--- a/chrome/common/channel_info_posix.cc
++++ b/chrome/common/channel_info_posix.cc
+@@ -21,7 +21,7 @@ version_info::Channel GetChannelImpl(std::string* modifier_out) {
+   version_info::Channel channel = version_info::Channel::UNKNOWN;
+   std::string modifier;
+ 
+-  char* env = getenv("CHROME_VERSION_EXTRA");
++  char* env = "Ungoogled Chromium";
```
Eloston, line 16: This would prevent reading the environment variable `CHROME_VERSION_EXTRA`.

patches/extra/ungoogled-chromium/add-extra-channel-info.patch
```diff
@@ -0,0 +1,30 @@
...
+     modifier = env;
+ 
+--- a/chrome/common/channel_info_win.cc
++++ b/chrome/common/channel_info_win.cc
+@@ -22,7 +22,7 @@ std::string GetChannelName() {
+ #endif  // defined(DCHECK_IS_CONFIGURABLE)
+   return base::UTF16ToASCII(channel);
+ #else
+-  return std::string();
++  return std::string("Ungoogled Chromium");
```
Eloston, line 27: We should probably test this on Windows to make sure it won't override the configured channel info.

Eloston (changes requested):
1. Instead of overriding extra channel info, can we simply append `(ungoogled-chromium)`?
2. Alternatively, would it be easier to patch the HTML file as suggested here? https://github.com/Eloston/ungoogled-chromium/issues/1303#issue-774748580. Or is this not cross-platform enough?

patches/extra/ungoogled-chromium/add-extra-channel-info.patch
wchen342: This is what Fedora uses in their `chromium` package so I just adapted it. Is there any actual a use case for `CHROME_VERSION_EXTRA`?

patches/extra/ungoogled-chromium/add-extra-channel-info.patch
wchen342: I cannot compile the WIndows version. Who is the maintainer of Win repo? Maybe he can help testing this.

patches/extra/ungoogled-chromium/add-extra-channel-info.patch
Eloston: > This is what Fedora uses in their chromium package so I just adapted it. Is there any actual a use case for CHROME_VERSION_EXTRA? Although appending shall be possible since the extra info is just a string.

Debian also sets it to indicate the build distribution & running distribution. I don't want to break that behavior.

Actually, could we just set `(ungoogled-chromium)` only if no extra channel info is provided at all? I don't want to force users to have this string if they don't want it.

patches/extra/ungoogled-chromium/add-extra-channel-info.patch
tangalbert919: I'm maintaining the Windows repo. I can test this for you.

Ahrotahn:
An alternative option would be to change "Official Build" in `components/version_ui_strings.grdp`.  

This wouldn't require any platform-specific changes, aligns more with the linked issue, and might clear some confusion in issues with people thinking they are running an 'official' build of ungoogled-chromium.

wchen342:
Won't that only apply when `is_official_build=true` is set? Otherwise you will see the string "Developer Build".

Ahrotahn:
That's true!  I guess "Developer Build" could also be changed.  Maybe have them named something like "ungoogled-chromium release build/ungoogled-chromium dev build"?

wchen342:
Yeah that can be a good idea. What does @Eloston think?

Eloston:
Sure, I don't mind changing the `(Developer Build)` and `(Release Build)` strings. Maybe even simply appending a string like `Developer Build, ungoogled-chromium` to not deviate too far from the existing format.

wchen342:
The patch has been updated to append `, ungoogled-chromium` to the two strings. Since it now only changes the string literals there shouldn't be any problem.

Eloston (approved):
LGTM, Thanks :+1:

PR #1333, opened 2021-01-16
Now available on the regular Flathub repo

rany2:
There is no need to enable the beta repo anymore.

Edit: also note that I will no longer update the beta repo because I don't want to build ungoogled twice. So by using the beta repo you'd be using an older chromium


Eloston (approved):
Thanks

PR #1342, opened 2021-01-20
Update to Chromium 88.0.4324.104

### LINKED ISSUE #1340: Updating to Chromium 88.0.4324.104
## Notes for the developer

## Notes for others

tangalbert919:
Resolves #1340. Managed to get a successful build on Arch Linux.


Luka-Filipovic:
That was fast, good job!
Building it on windows right now, I will let you know once it passes.

Ahrotahn:
Builds and runs fine on my end, thanks!

kramred:
Build on macOS (x86_64) also successful, build on GitHub Actions just started – let's hope that runs smoothly too

kramred:
Popups now open a new __Window__, not a __Tab__ – can someone else test&confirm?

Probably related patch: [popups-to-tabs.patch](https://github.com/Eloston/ungoogled-chromium/blob/master/patches/extra/ungoogled-chromium/popups-to-tabs.patch)

git-bruh:
@kramred doesn't seem to be happening for me, can you give an example URL?

kramred:
Yes, sorry. I should have been more specific. It seems to happen only with Javascript and the [`Window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) method.

Click the button with the first example `about:blank [ex.1]` [here](https://kramred.github.io/web-test/open_new_window.html)  
and a new Window should open instead of a tab; the other examples open tabs as expected

PF4Public:
`about:blank [ex.1]` opens new window without toolbars, others go into tab (either new, or the actual one) on Gentoo.

kramred:
You can also try it in the developer console, the minimal working example seems to be one that sets some (arbitrary) [Window features](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#window_features) like this:

`window.open('about:blank','newWindow','bla')`

Ahrotahn:
I tested using https://webbrowsertools.com/popup-blocker/ and about half of them created a popup.  I reinstalled 87 and all of them opened in new tabs.  So it seems to work for some methods but not others.  

I did some digging and it looks like they are moving a lot out of render_view_impl.cc and letting mojo handle things.  
There were two changes that both worked for me:  

Either add to the existing `popups-to-tabs.patch`:
```diff
--- a/ui/base/mojom/window_open_disposition_mojom_traits.h
+++ b/ui/base/mojom/window_open_disposition_mojom_traits.h
@@ -28,7 +28,7 @@
       case WindowOpenDisposition::NEW_BACKGROUND_TAB:
         return ui::mojom::WindowOpenDisposition::NEW_BACKGROUND_TAB;
       case WindowOpenDisposition::NEW_POPUP:
-        return ui::mojom::WindowOpenDisposition::NEW_POPUP;
+        return ui::mojom::WindowOpenDisposition::NEW_FOREGROUND_TAB;
       case WindowOpenDisposition::NEW_WINDOW:
         return ui::mojom::WindowOpenDisposition::NEW_WINDOW;
       case WindowOpenDisposition::SAVE_TO_DISK:
@@ -62,7 +62,7 @@
         *out = WindowOpenDisposition::NEW_BACKGROUND_TAB;
         return true;
       case ui::mojom::WindowOpenDisposition::NEW_POPUP:
-        *out = WindowOpenDisposition::NEW_POPUP;
+        *out = WindowOpenDisposition::NEW_FOREGROUND_TAB;
         return true;
       case ui::mojom::WindowOpenDisposition::NEW_WINDOW:
         *out = WindowOpenDisposition::NEW_WINDOW;

```

Or replace `popups-to-tabs.patch` with:
```diff
--- a/chrome/browser/ui/browser.cc
+++ b/chrome/browser/ui/browser.cc
@@ -1648,16 +1648,7 @@
                              const gfx::Rect& initial_rect,
                              bool user_gesture,
                              bool* was_blocked) {
-#if defined(OS_MAC)
-  // On the Mac, the convention is to turn popups into new tabs when in
-  // fullscreen mode. Only worry about user-initiated fullscreen as showing a
-  // popup in HTML5 fullscreen would have kicked the page out of fullscreen.
-  if (disposition == WindowOpenDisposition::NEW_POPUP &&
-      exclusive_access_manager_->fullscreen_controller()
-          ->IsFullscreenForBrowser()) {
     disposition = WindowOpenDisposition::NEW_FOREGROUND_TAB;
-  }
-#endif
 
   // At this point the |new_contents| is beyond the popup blocker, but we use
   // the same logic for determining if the popup tracker needs to be attached.

```

PF4Public:
> Or replace `popups-to-tabs.patch` with:
> 
> ```diff
> --- a/chrome/browser/ui/browser.cc
> +++ b/chrome/browser/ui/browser.cc
> @@ -1648,16 +1648,7 @@
>                               const gfx::Rect& initial_rect,
>                               bool user_gesture,
>                               bool* was_blocked) {
> -#if defined(OS_MAC)
> -  // On the Mac, the convention is to turn popups into new tabs when in
> -  // fullscreen mode. Only worry about user-initiated fullscreen as showing a
> -  // popup in HTML5 fullscreen would have kicked the page out of fullscreen.
> -  if (disposition == WindowOpenDisposition::NEW_POPUP &&
> -      exclusive_access_manager_->fullscreen_controller()
> -          ->IsFullscreenForBrowser()) {
>      disposition = WindowOpenDisposition::NEW_FOREGROUND_TAB;
> -  }
> -#endif
>  
>    // At this point the |new_contents| is beyond the popup blocker, but we use
>    // the same logic for determining if the popup tracker needs to be attached.
> ```

Will it help non-Macs?

Ahrotahn:
It works on my linux systems, I'm not able to test on other platforms right now though.  

That part of the code was changing the popup to a foreground tab only on Mac systems that are in user-initiated fullscreen mode.  
Removing that check makes it so that all popups, regardless of platform or fullscreen state, are made into new tabs.  

Granted, I haven't tested either of these changes for long enough to be sure there isn't a side-effect that I haven't encountered yet.  I'll be testing more throughout the day and will update with any new info I find.

tangalbert919:
Chromium 88.0.4324.104 was released for Windows yesterday, so expect another commit to this PR in a bit.

kramred:
@Ahrotahn – how did you test that so quickly? 

I.e. what's your/the procedure to recompile – can you just change some source files and run ninja again to incorporate it or did you re-compile the whole thing (in such a short time)?

PF4Public:
> Chromium 88.0.4324.104 was released for Windows yesterday

Could Linux and Mac also be built from 104?

kramred:
> > can you just change some source files and run ninja again to incorporate it
> 
> Isn't it exactly the way it is supposed to be?

Probably; I thought so too – but I had not tried yet and wanted to know from someone who did that successfully…
Good to know that it should work that way. Thanks!

Ahrotahn:
It probably appeared so quick because I did the work yesterday but didn't post until this morning!  
I have a 2700X in my system so clean builds are usually done within a few hours at most.  When doing dirty builds it is often done in under an hour depending on the changes.

PF4Public:
> Could Linux and Mac also be built from 104?

104 seems to be [configuring just fine on Linux](https://github.com/PF4Public/gentoo-overlay/actions/runs/503997904).

kramred:
Build on macOS x86_64 also finished successfully and with the [updated `popups-to-tabs.patch`](https://github.com/Eloston/ungoogled-chromium/pull/1342#issuecomment-765385002) from @Ahrotahn the new window issue (for `window.open('about:blank','newWindow','something')`) seems to be resolved.
@tangalbert919 – could you please include the updated patch in your PR? (I've used the second version for my build)

Ahrotahn:
I edited my prior comment to add one small change to the second diff.  
I re-included the `if (disposition == WindowOpenDisposition::NEW_POPUP) {` check.  
That section should only handle popups (and the comments there seem to imply as much), but I feel it's better to be explicit in case Google decides to change that in the future.

SugaryHull:
I'm currently testing to see if this will build with the current Debian-specific build instructions for Buster

SugaryHull:
I'm not having much luck with this. It seems to get stuck with extracting the upstream tarball and applying the patches.

screen

SugaryHull:
Portable Linux got farther along, but returned an error when trying to patch "gpu/ipc/service/gpu_watchdog_thread.cc"
portable-linux-patch-issues

braewoods:
@SugaryHull It's to be expected. I plan to deal with once this PR is merged. The patches always have to be reviewed for major bumps like this.

SugaryHull:
Good to know it's not something on my end. Keep up the good work

jstkdng:
Managed to build for arch and so far so good. Thanks.

patches/extra/ungoogled-chromium/add-flag-to-scroll-tabs.patch
```diff
@@ -38,11 +38,12 @@
...
-@@ -259,7 +274,7 @@ const char* BrowserRootView::GetClassNam
- }
+@@ -265,8 +280,7 @@ bool BrowserRootView::OnMouseWheel(const
  
- bool BrowserRootView::OnMouseWheel(const ui::MouseWheelEvent& event) {
--  if (browser_defaults::kScrollEventChangesTab) {
+   // Scroll-event-changes-tab is incompatible with scrolling tabstrip, so
+   // disable it if the latter feature is enabled.
+-  if (browser_defaults::kScrollEventChangesTab &&
+-      !base::FeatureList::IsEnabled(features::kScrollableTabStrip)) {
```
Zoraver, line 46: It makes sense that `scroll_event_changes_tab_` takes the place of `browser_defaults::kScrollEventChangesTab`, but it is less clear to me whether it should also take the place of `!base::FeatureList::IsEnabled(features::kScrollableTabStrip)`, considering the comment above.

tangalbert919: Should I just leave `!base::FeatureList::IsEnabled(features::kScrollableTabStrip)` where it was then?

Zoraver: Yes, unless you have a reason to remove it.

patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
```diff
@@ -13,21 +13,28 @@
...
+         <div class="separator"></div>
+         <a role="menuitem" id="clear-browsing-data"
+             href="chrome://settings/clearBrowserData"
+@@ -131,4 +126,4 @@
+         <iron-icon icon="cr:info-outline"></iron-icon>
+         <div>$i18nRaw{sidebarFooter}</div>
+       </div>
+-    </div>
+\ No newline at end of file
++    </div>
```
Zoraver, line 34: The above three lines just look weird to me. I don't understand why quilt is even touching this portion of the file, let alone adding `\ No newline at the end of file` *before* the end of the file... This looks like a quilt bug to me, but perhaps this is just intended behavior that I am not aware of.

Eloston (changes requested):
Looking good, just a few comments.

In addition, please also add the changes suggested by @Ahrotahn in https://github.com/Eloston/ungoogled-chromium/pull/1342#issuecomment-765385002. Let's go with the first set of changes, because I think they will be more permanent than the rewritten patch.

patches/extra/ungoogled-chromium/add-flag-for-pdf-plugin-name.patch
```diff
@@ -31,32 +31,6 @@
...
-     if (!plugins->GetString(i, &plugin))
-       continue;
--    if ((base::MatchPattern(ChromeContentClient::kPDFExtensionPluginName,
-+    if ((base::MatchPattern(ChromeContentClient::GetPDFExtensionPluginName(),
-                             plugin) ||
--         base::MatchPattern(ChromeContentClient::kPDFInternalPluginName,
-+         base::MatchPattern(ChromeContentClient::GetPDFInternalPluginName(),
-                             plugin)) &&
-         !policies.GetValue(policy::key::kAlwaysOpenPdfExternally)) {
-       prefs->RemoveValue(prefs::kPluginsAlwaysOpenPdfExternally);
```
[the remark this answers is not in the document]
tangalbert919: The file `plugin_policy_handler.cc` actually got removed in Chromium 88, and I could not find similar code in any of the other files.

patches/extra/ungoogled-chromium/add-flag-to-scroll-tabs.patch
tangalbert919: I just pushed a commit to resolve this.

tangalbert919:
I added the following to `popups-to-tabs.patch`:

```patch
--- a/ui/base/mojom/window_open_disposition_mojom_traits.h
+++ b/ui/base/mojom/window_open_disposition_mojom_traits.h
@@ -28,7 +28,7 @@
       case WindowOpenDisposition::NEW_BACKGROUND_TAB:
         return ui::mojom::WindowOpenDisposition::NEW_BACKGROUND_TAB;
       case WindowOpenDisposition::NEW_POPUP:
-        return ui::mojom::WindowOpenDisposition::NEW_POPUP;
+        return ui::mojom::WindowOpenDisposition::NEW_FOREGROUND_TAB;
       case WindowOpenDisposition::NEW_WINDOW:
         return ui::mojom::WindowOpenDisposition::NEW_WINDOW;
       case WindowOpenDisposition::SAVE_TO_DISK:
@@ -62,7 +62,7 @@
         *out = WindowOpenDisposition::NEW_BACKGROUND_TAB;
         return true;
       case ui::mojom::WindowOpenDisposition::NEW_POPUP:
-        *out = WindowOpenDisposition::NEW_POPUP;
+        *out = WindowOpenDisposition::NEW_FOREGROUND_TAB;
         return true;
       case ui::mojom::WindowOpenDisposition::NEW_WINDOW:
         *out = WindowOpenDisposition::NEW_WINDOW;
```

That should address [this comment](https://github.com/Eloston/ungoogled-chromium/pull/1342#issuecomment-765385002).

patches/extra/ungoogled-chromium/add-flag-for-pdf-plugin-name.patch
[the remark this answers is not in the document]
Ahrotahn: It was removed in [this commit](https://chromium.googlesource.com/chromium/src/+/861c8bac3925acfb511f22973ce7d7fff1ad52b0) as part of a plugin code cleanup spurred by Flash's removal.

patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
```diff
@@ -57,153 +57,169 @@
...
  }
  
  void SafeBrowsingPrivateEventRouter::OnAnalysisConnectorResult(
-@@ -433,6 +439,7 @@ void SafeBrowsingPrivateEventRouter::OnD
-     const std::string& trigger,
+@@ -420,6 +426,7 @@ void SafeBrowsingPrivateEventRouter::OnA
+     const enterprise_connectors::ContentAnalysisResponse::Result& result,
      const int64_t content_size,
      safe_browsing::EventResult event_result) {
 +#if defined(FULL_SAFE_BROWSING)
```
Zoraver, line 130: This `#if defined(FULL_SAFE_BROWSING)` got moved from the start of `SafeBrowsingPrivateEventRouter::OnDangerousDeepScanningResult()` to `SafeBrowsingPrivateEventRouter::OnAnalysisConnectorResult()`. Since the corresponding `#endif` is at the end of `SafeBrowsingPrivateEventRouter::OnDangerousDeepScanningResult()`, the net effect of this is that there is no implementation of `SafeBrowsingPrivateEventRouter::OnDangerousDeepScanningResult()`. Since this is ungoogled-chromium, this function is not likely to be called, so this might not cause issues immediately, but it might cause linker errors down the road. I would recommend correctly scoping the `#if defined ... #endif` block; note that if you do this, you will likely just need to add two lines: an `#endif` at the end of `OnAnalysisConnectorResult()` and an `#if defined(FULL_SAFE_BROWSING)` at the start of `OnDangerousDeepScanningResult()`.

Eloston: @tangalbert919 Please confirm if this change is needed. Otherwise this LGTM.

wchen342:
Works fine on CentOS and Fedora.

Eloston (approved):
LGTM :+1: 

It's really awesome to see everyone collaborating on testing. Special thanks to those who have diligently reviewed this PR and suggested fixes!

PR #1364, opened 2021-01-30
Follow-up for Chromium 88

Zoraver:
During my review of #1342, I identified two things that didn't directly effect current browser functionality but could cause issues down the road:

1. `remove-unneeded-ui.patch` has `\ No newline at end of file` *before* the end of `chrome/browser/resources/history/side_bar.html`.
2. A `#if defined(FULL_SAFE_BROWSING)` in `fix-building-without-safebrowsing.patch` got moved such that there is no implementation for `SafeBrowsingPrivateEventRouter::OnDangerousDeepScanningResult()`; this appears to not cause issues currently, but could cause linker errors in the future.

This pull request addresses both of these nitpicks; I have compiled ungoogled-chromium with these changes and it is running well.

### DIFF
--- a/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
+++ b/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
@@ -131,7 +131,23 @@
    if (!IsRealtimeReportingEnabled())
      return;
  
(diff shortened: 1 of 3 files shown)


Eloston (approved):
LGTM. Thanks for looking out for the code quality!

Issue #1372, opened 2021-02-02
Updating to Chromium 88.0.4324.146


Zoraver:
## Notes for the developer

Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:

![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

Finally, make sure to reference this issue in your PR.

Feel free to ask questions and discuss issues here along the way. Thanks!

## Notes for others

Feel free to raise issues or questions throughout the process. However, please refrain from asking ETAs unless no visible progress has been made here or on the developer's PR for a while (e.g. 2 weeks).

Zoraver:
I plan on working on this.

braewoods:
@Zoraver can you see about updating the extra/debian/disable/android.patch? The upstream version has changed to also disable building the md5sum tool and I had to temporarily include the patch in the debian repository in order to build the package. The updated version is here: https://salsa.debian.org/chromium-team/chromium/-/blob/master/debian/patches/disable/android.patch

Zoraver:
@braewoods I'm afraid I am out of time today to work on this and I don't feel like running my computer overnight just to test compilation with the the updated `android.patch`. You may want to submit a pull request updating it yourself; alternately, I will consider updating into it the next time that I update ungoogled-chromium to a new stable Chromium version.

braewoods:
Ok. I'm just not familiar with the upstream repo so I'm not sure how to refresh the patches to ensure they pass the validation scripts.

Eloston:
Actually do we still want `android.patch`? @wchen342, would it be simpler for ungoogled-chromium-android to remove `android.patch`?

@Zoraver @braewoods I can take care of it after I merge the PR

braewoods:
@Eloston I don't mind if it remains exclusive to the Debian version but I need it to be updated here or there.

Eloston:
In that case I'll just remove it. I don't see any reason why we are keeping it in this repo.

@wchen342 If this is a bad idea, please let me know. I'll revert it.

wchen342:
> In that case I'll just remove it. I don't see any reason why we are keeping it in this repo.
> 
> @wchen342 If this is a bad idea, please let me know. I'll revert it.

[...]

PR #1376, opened 2021-02-03
Update to Chromium 88.0.4324.146

Zoraver:
Addresses #1372. Ungoogled-chromium 88.0.4324.146 compiled and is running well.


Eloston (approved):
LGTM, Thanks!

Issue #1356, opened 2021-01-26
Add an option to completely disable chrome:// suggestions in the search bar


ungubajue90k:
Hello,

I use this latest ungoogled-chromium android version
https://github.com/ungoogled-software/ungoogled-chromium-android/releases/tag/87.0.4280.141-2-extensions

If I type `a`, `c` or `ch` in the search bar immediately these suggestions appear: 

- chrome://chrome-urls/
- chrome://version/

![IMG_20210126_160418](https://user-images.githubusercontent.com/77856086/105863201-eee97880-5ff0-11eb-9055-2423f8b77e9b.jpg)
![IMG_20210126_160758](https://user-images.githubusercontent.com/77856086/105863224-f446c300-5ff0-11eb-954f-5a8c8a41052a.jpg)

Please add an option to completely disable the chrome:// suggestions in the search bar. I want to decide which search suggestions I want in the search bar.

wchen342:
Adding to what the OP has said, I think this should be included in the base patchset since it is not an Android specific feature.

Ahrotahn:
I recall playing with the builtin suggestions when working on the omnibox filtering patch.  
That one is treated a bit differently since it works even with suggestions turned off.  It might be something that would need to have it's own flag instead of being rolled into the existing one.  I'll experiment with it over the week as I get time.

Luka-Filipovic:
It think [this file](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/ui/webui/chrome_web_ui_controller_factory.cc;drc=aafcff480d10ea649036833cb4052f127d733143;l=485) lists of internal pages that should appear in omnibox. Try to force it to always return `nullptr`.

Ph0rk0z:
I actually like this. Showed me all the chrome:// urls

PR #1378, opened 2021-02-04
Changes to omnibox filtering to add chrome pages

Ahrotahn:
This PR addresses #1356 by updating `add-flag-for-omnibox-autocomplete-filtering.patch` to include options for internal chrome pages.  

This flag had two options:  `search-suggestions-only` and `search-suggestions-and-bookmarks`
Those options have been changed to allow for diferent combinations:  `search`, `search-bookmarks`, `search-chrome`, and `search-bookmarks-chrome`

### Detail:

* `autocomplete_controller.cc`:  Since there's more than two options now it's better to change `provider_types` rather than conditionally branching based on the flag's value.  The enabled providers can vary, so rather than setting them directly this will use the existing state of the required providers and then remove the ones we don't need.  
Extra info:  [Default Providers](https://source.chromium.org/chromium/chromium/src/+/refs/tags/88.0.4324.146:components/omnibox/browser/autocomplete_classifier.cc;l=41), [Provider type positions](https://source.chromium.org/chromium/chromium/src/+/refs/tags/88.0.4324.146:components/omnibox/browser/autocomplete_provider.h;l=140)
Using `Contains` allows the old flags to continue to function, though chrome pages would be excluded until the user updates the flag to opt in to them.  

* `history_url_provider.cc` and `search_provider.cc`:  These have been updated to check if the flag is set rather than explicitly checking for the specific flag values.  

* `url_fixer.cc`:  Even with the chrome pages removed from suggestions there would still be a suggestion for `chrome://version/`.  This is due to the URL fixer adding the value when interpreting the input as a chrome URL.


Eloston (approved):
Pretty clever implementation. Great work detailing the changes too. LGTM!

PR #1383, opened 2021-02-08
Add flag to popups-to-tabs patch

Ahrotahn:
This PR adds a flag to the popups-to-tabs patch, addressing #1377.  

I've renamed the patch to reflect that it now adds a flag.


Eloston (approved):
Awesome, thanks!

Issue #1389, opened 2021-02-11
update_lists.py exclude pattern


cjbd:
hello, i have chromium src git repo already, i wanted to test `update_lists.py` script by:

```sh
./devutils/update_lists.py -t /path/to/chromium/src
```
this will consume all RAM and crash, with error like this:
```
ERROR: Unhandled exception while processing .git/objects/pack/pack-5582c833efb8d8fe08cc194ce33fdf372b9e495f.pack
```
obviously, it's also scanning `.git` folders, i googled a bit, trying to exclude this folder by modifying below code in `update_lists.py`
```py
    with Pool(processes) as procpool:
        returned_data = procpool.starmap(
            compute_lists_proc,
            zip(source_tree.rglob('[!.git]*'), repeat(source_tree), repeat(search_regex)))
```
it's not working, it still uses all RAM then crashes

i'm not very familiar with `rglob`, how do i scan the source tree with exclude pattern like excluding all `.git` folders?

thank you!

Ahrotahn:
Setting the `--processes` flag to a lower value should help with the memory issue.  By default it will spawn a process for each of your cores and this can take a lot of RAM.  

The glob exclusion looks like it should work though.

cjbd:
@Ahrotahn , thanks for the update, but after the change, there's still the same error: `ERROR: Unhandled exception while processing .git/objects/pack/pack-5582c833efb8d8fe08cc194ce33fdf372b9e495f.pack`, i don't think the exclusion is working  
i also tried using `--process 1`, and opened system monitor, it still uses all RAM then crash with same error

Ahrotahn:
You used `--process`, but the correct flag is <code>--process<b>es</b></code>.  

It seems the glob exclusion excludes the git directory, but not the files in it!  It might be better to change compute_lists_proc to return if the path is in the directory.  Alternatively, you could just move the directory out temporarily.  
Normally the release tarballs are used which doesn't include the git directory.  

How much RAM do you have?  I don't think you'll be able to compile chromium without at least 8GB.

cjbd:
@Ahrotahn, thanks for the update, i have 32GB RAM

cjbd:
this works:
```py
with Pool(processes) as procpool:
    returned_data = procpool.starmap(
        compute_lists_proc,
        zip(filter(lambda x: ".git" not in x.parts, source_tree.rglob('*')), repeat(source_tree), repeat(search_regex)))
```

PR #1397, opened 2021-02-17
add filter to compute_lists source_tree.rglob('*')

cjbd:
for #1389
this change updates `update_lists.py`, `compute_lists` function, to avoid scanning `.git` dirs, so that the script would work for local chromium repo cloned using `git` and `gclient` tools

tested on ubuntu 18 LTS, it's working

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -298,7 +298,7 @@ def compute_lists(source_tree, search_regex, processes):
     with Pool(processes) as procpool:
         returned_data = procpool.starmap(
             compute_lists_proc,
-            zip(source_tree.rglob('*'), repeat(source_tree), repeat(search_regex)))
+            zip(filter(lambda x: '.git' not in x.parts, source_tree.rglob('*')), repeat(source_tree), repeat(search_regex)))
 
     # Handle the returned data
     for (used_pep_set, used_pip_set, used_dep_set, used_dip_set, returned_pruning_set,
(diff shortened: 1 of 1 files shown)


Eloston (changes requested):
This is a good idea, but this implementation is too hacky.

I propose we add a new filtering scheme to `should_domain_substitute`, such as:

* `DOMAIN_EXCLUDE_PATTERNS` - Like the `INCLUDE` list of the same name, but does exclusions instead of inclusions
* `DOMAIN_EXCLUDE_PARTS` - Exclude paths containing at least one of the listed `Path.parts`

The new precedence order would be as follows:

- `DOMAIN_EXCLUDE_PREFIXES`
- *One of the new rules here*
- `DOMAIN_INCLUDE_PATTERNS`

@Ahrotahn What do you think?

Ahrotahn:
I like that idea.  The current implementation is specific to the `.git` path, but having more granular exclude patterns would allow for other paths to excluded as needs arise.  For example it may be a good idea to include quilt's `.pc` path as a precaution as well.  

@cjbd's issue seemed to be with the binary check, so similar changes would also need to be in `should_prune`.

cjbd:
@Eloston, i took another look, it seems the change to implement DOMAIN_EXCLUDE_PATTERNS and DOMAIN_EXCLUDE_PARTS is not a small one, i'm not sure my python skill is up to the task
and i don't fully understand what's the difference between DOMAIN_EXCLUDE_PATTERNS and existing DOMAIN_EXCLUDE_PREFIXES

@Ahrotahn, the filter i added can exclude pattern from both domain and prune lists

cjbd:
@Eloston , i'm closing this PR, this project already covered source tar download, and there're many differences for repo cloned using git and gclient

Eloston:
Sorry I couldn't get to implementing your change in a timely manner. If someone else wants to implement this, feel free.

PR #1400, opened 2021-02-19
Add GPC support

### LINKED ISSUE #1399: GPC (Global Privacy Control)
This is a [feature request](https://github.com/ungoogled-software/ungoogled-chromium-android/issues/63) from Android repo, but I think it is worth implemented globally (as a flag).

*** start of the original post ***
**Please explain why your feature request fall into the scope of ungoogled-chromium?**
GPC? Is Stand For Global Privacy Control If This Enabled By Default 

This Browser Will Automatic Send Signal On Website To Not Track And Sell Userdata

**Please describe the feature you would like to have**
GPC Global Privacy Control

You Add GPC And Enabled By Default 
(Make It Cannot be disabled )

No Alternative Because Many Extension Not Working Include Privacy Badger That's Why I Request To Add GPC By Default

**Additional context**
https://globalprivacycontrol.org/#download

*** end of the original post ***

wchen342:
Resolve #1399.

It was tested and proved to work on Android version.

I have removed Android specific part and left the native part here.

### DIFF
--- a/patches/extra/ungoogled-chromium/add-global-privacy-control-option.patch
+++ b/patches/extra/ungoogled-chromium/add-global-privacy-control-option.patch
@@ -1,0 +1,823 @@
+# Add Global Privacy Control support and option in settings
+
+---
+ chrome/app/settings_strings.grdp                                                             |    9 
+ chrome/browser/extensions/api/preference/preference_api.cc                                   |    2 
+ chrome/browser/extensions/api/settings_private/prefs_util.cc                                 |    2 
+ chrome/browser/renderer_preferences_util.cc                                                  |    2 
+ chrome/browser/resources/settings/BUILD.gn                                                   |    1 
+ chrome/browser/resources/settings/privacy_page/BUILD.gn                                      |   12 +
+ chrome/browser/resources/settings/privacy_page/cookies_page.html                             |    3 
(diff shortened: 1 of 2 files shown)


Eloston (changes requested):
I am inclined to close this for several reasons:

1. I think this patch is far too large just to add a settings toggle that sends one extra header. It does a great job of conforming to the Chromium coding standards, but it is not great for us to maintain.
2. As discussed in the Android issue, I think the benefit of this patch is questionable. At the very least, it seems to have a lot of politics around it.
3. If we also consider the ongoing discussion we are having about privacy enhancements in ungoogled-chromium, I am not sure if these kinds of changes are appropriate to ungoogled-chromium.
4. I am still not clear on why Privacy Badger fails to provide this feature.

In short, as the patch is now, I won't accept it. Let me know what you think.

wchen342:
> I think the benefit of this patch is questionable.
> I am not sure if these kinds of changes are appropriate to ungoogled-chromium.

Yes I realized that lately, and I am totally fine if you think it does not fit in `ungoogled-chromium`. Actually I plan to remove some extra features in Android too so I will just close this for now.

> why Privacy Badger fails to provide this feature

It was a mistake in the user report. NoScript blocked the DOM detection.

Issue #1401, opened 2021-02-20
domain_substitution.py to skip readonly files


cjbd:
hello, i'm trying `domain_substitution.py` for my linux local chromium source cloned using git and gclient tools, by
```sh
./utils/domain_substitution.py apply -r domain_regex.list -f domain_substitution.list -c build/domsubcache.tar.gz /path/to/chromium/src
```
and i'm getting error:
```
Traceback (most recent call last):
  File "./utils/domain_substitution.py", line 399, in <module>
    main()
  File "./utils/domain_substitution.py", line 395, in main
    args.callback(args)
  File "./utils/domain_substitution.py", line 349, in _callback
    apply_substitution(args.regex, args.files, args.directory, args.cache)
  File "./utils/domain_substitution.py", line 261, in apply_substitution
    crc32_hash, orig_content = _substitute_path(path, regex_pairs)
  File "./utils/domain_substitution.py", line 91, in _substitute_path
    with path.open('r+b') as input_file:
  File "/usr/lib/python3.6/pathlib.py", line 1183, in open
    opener=self._opener)
  File "/usr/lib/python3.6/pathlib.py", line 1037, in _opener
    return self._accessor.open(self, flags, mode)
  File "/usr/lib/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
PermissionError: [Errno 13] Permission denied: '/path/tochromium/src/third_party/android_sdk/public/build-tools/30.0.1/NOTICE.txt'
```
it turns out there're many readonly files, i'm guessing this is not an issue when using source downloaded from tar file

i'm thinking to update `update_lists.py` to exclude all readonly files

PR #1406, opened 2021-02-21
skip read only files

cjbd:
for #1401

this change added a new private function `_is_writable` to check file write permission, and update `should_domain_substitute` function to check `_is_writable` first before other validations, if file is read only, return False

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -14,6 +14,7 @@ the process has finished.
 import argparse
 import os
 import sys
+import stat
 
 from itertools import repeat
 from multiprocessing import Pool
@@ -156,6 +157,15 @@ def _is_binary(bytes_data):
     return bool(bytes_data.translate(None, _TEXTCHARS))
 
 
+def _is_writable(path):
+    """
+    Return True if path has write permission; False otherwise
+
+    path is a pathlib.Path or string to a directory to test.
+    """
+    st = os.stat(path)
+    return bool(st.st_mode & stat.S_IWRITE)
+
 def _dir_empty(path):
     """
     Returns True if the directory is empty; False otherwise
@@ -227,6 +237,8 @@ def should_domain_substitute(path, relative_path, search_regex, used_dep_set, us
     used_dep_set is a list of DOMAIN_EXCLUDE_PREFIXES that have been matched
     used_dip_set is a list of DOMAIN_INCLUDE_PATTERNS that have been matched
     """
+    if not _is_writable(path):
+        return False
     relative_path_posix = relative_path.as_posix().lower()
     for include_pattern in DOMAIN_INCLUDE_PATTERNS:
         if PurePosixPath(relative_path_posix).match(include_pattern):
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -166,6 +166,7 @@ def _is_writable(path):
     st = os.stat(path)
     return bool(st.st_mode & stat.S_IWRITE)
 
+
 def _dir_empty(path):
     """
     Returns True if the directory is empty; False otherwise


Eloston (changes requested):
Technically this only checks if the file owner has write permissions, not whether the user executing this Python program (`update_lists.py`) has permission to write to the file. See Python docs: https://docs.python.org/3/library/stat.html?highlight=stat#stat.S_IWRITE. A better method is to use [`os.access`](https://docs.python.org/3/library/os.html#os.access). Arguably, the best method is to attempt to domain substitute and abort or skip if an error occurs.

However, I wonder if skipping read-only files is the correct action here. Wouldn't it be safer to update the permissions of the files so domain substitution can work? What is the upside of skipping read-only files?

cjbd:
@Eloston, i guess there're many differences between using downloaded and extracted source files and git cloned source files
i haven't thought all the way through, i only looked at bunch of read only files on my drive, they seems to have nothing to do with source code, i haven't fully tested it yet
you're right, there may be read only files that need to do domain substitution too

cjbd:
i used script to pull all read only files for chromium repo, file attached, it seems ok to skip those, but i'm not 100% sure

[read_only.zip](https://github.com/Eloston/ungoogled-chromium/files/6081218/read_only.zip)

Ahrotahn:
Skipping files could possibly leave you in a situation where your build is not really ungoogled.  
I'm surprised that 43fbd1d76cda2367dbb869e4984ada276833cf5e doesn't work for you.  Does the user you're running the build under have the authority to change those file's permissions?

cjbd:
@Ahrotahn , i don't know why, i got below error message, which points to `with path.open('r+b') as input_file:`

```
  File "./utils/domain_substitution.py", line 399, in <module>
    main()
  File "./utils/domain_substitution.py", line 395, in main
    args.callback(args)
  File "./utils/domain_substitution.py", line 349, in _callback
    apply_substitution(args.regex, args.files, args.directory, args.cache)
  File "./utils/domain_substitution.py", line 261, in apply_substitution
    crc32_hash, orig_content = _substitute_path(path, regex_pairs)
  File "./utils/domain_substitution.py", line 91, in _substitute_path
    with path.open('r+b') as input_file:
  File "/usr/lib/python3.6/pathlib.py", line 1183, in open
    opener=self._opener)
  File "/usr/lib/python3.6/pathlib.py", line 1037, in _opener
    return self._accessor.open(self, flags, mode)
  File "/usr/lib/python3.6/pathlib.py", line 387, in wrapped
    return strfunc(str(pathobj), *args)
PermissionError: [Errno 13] Permission denied: '/path/tochromium/src/third_party/android_sdk/public/build-tools/30.0.1/NOTICE.txt'
```

Ahrotahn:
Are you able to run `chmod +w /path/tochromium/src/third_party/android_sdk/public/build-tools/30.0.1/NOTICE.txt`?

It's possible some part of that path is not owned by your user, or you don't have executable permission on one of the directories in the path.

cjbd:
@Ahrotahn , the file is owned by me, this is probably set by `git` or `gclient sync` to readonly
i'm thinking for chromium&chromiumos source code cloned using official method, have to go for a different way for domain substitution

Screenshot from 2021-03-06 12-21-43

Ahrotahn:
Would you mind running `namei -l /path/tochromium/src/` and posing the output?

cjbd:
@Ahrotahn 
```
f: .
drwxr-xr-x pi pi .
```

Ahrotahn:
You'll need to use the full path to your source directory to get any meaningful results.

cjbd:
@Ahrotahn , thanks for the update, i was busy with other stuff, i scanned both chromium&chromiumos source to extract domains, it's in discussion

@Eloston i don't think this PR is needed for this project, i'm closing it
as for repo cloned using git and gclient, there're many differences, this project already covered download source tar, no need for this change

Issue #1407, opened 2021-02-21
GitHub API - Old Releases Issue


v00v:
Why are there no latest package available from "ungoogled-chromium" when accessing it from GitHub API? As you can see by using this API url, it only gives very old version "67.0.3396.87-1" back from "2018-06-19T19:09:55Z". Can you please fix this issue? So those who use github API would be able to pull this data correctly? Thank you!

EDIT-2: It also happens when you try to access this url;
```https://github.com/Eloston/ungoogled-chromium/releases/latest```

Example Command:
```sh
curl -H "Accept: application/vnd.github.v3+json" \
https://api.github.com/repos/eloston/ungoogled-chromium/releases/latest
```

Example Result:
```sh
{
  "url": "https://api.github.com/repos/Eloston/ungoogled-chromium/releases/11555347",
  "assets_url": "https://api.github.com/repos/Eloston/ungoogled-chromium/releases/11555347/assets",
  "upload_url": "https://uploads.github.com/repos/Eloston/ungoogled-chromium/releases/11555347/assets{?name,label}",
  "html_url": "https://github.com/Eloston/ungoogled-chromium/releases/tag/67.0.3396.87-1",
  "id": 11555347,
  "author": {
    "login": "Eloston",
    "id": 1305560,
    "node_id": "MDQ6VXNlcjEzMDU1NjA=",
    "avatar_url": "https://avatars.githubusercontent.com/u/1305560?v=4",
    "gravatar_id": "",
    "url": "https://api.github.com/users/Eloston",
    "html_url": "https://github.com/Eloston",
    "followers_url": "https://api.github.com/users/Eloston/followers",
    "following_url": "https://api.github.com/users/Eloston/following{/other_user}",
    "gists_url": "https://api.github.com/users/Eloston/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/Eloston/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/Eloston/subscriptions",
    "organizations_url": "https://api.github.com/users/Eloston/orgs",
    "repos_url": "https://api.github.com/users/Eloston/repos",
    "events_url": "https://api.github.com/users/Eloston/events{/privacy}",
    "received_events_url": "https://api.github.com/users/Eloston/received_events",
    "type": "User",
    "site_admin": false
  },
  "node_id": "MDc6UmVsZWFzZTExNTU1MzQ3",
  "tag_name": "67.0.3396.87-1",
[...]

PR #1435, opened 2021-03-14
automatically create empty release for each tag; fixes #1414

### LINKED ISSUE #1414: "Latest" release is from 2018
Can you fix this?

image

zocker-160:
This Github workflow creates a new empty release for each (new) pushed tag and for tags only. Normal push events are getting ignored.

This solution will not create releases for past tags, only for newly pushed ones after this workflow is active.

This is an attempt to fulfill all 3 criteria given by @Eloston in https://github.com/Eloston/ungoogled-chromium/issues/1414#issuecomment-798822753

### DIFF
--- a/.github/workflows/release-on-tag.yml
+++ b/.github/workflows/release-on-tag.yml
@@ -10,9 +10,6 @@ jobs:
     name: Create Release
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@master
-        
       - name: Create Release
         id: create_release
         uses: actions/create-release@latest
--- a/.github/workflows/release-on-tag.yml
+++ b/.github/workflows/release-on-tag.yml
@@ -1,0 +1,25 @@
+name: Create new empty release
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
+        
+      - name: Create Release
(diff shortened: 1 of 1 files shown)


zocker-160:
by looking at the issues it should also fix #1407

.github/workflows/release-on-tag.yml
```diff
@@ -0,0 +1,25 @@
...
+    tags:
+      - '*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@master
```
Eloston, line 14: What is the purpose of checking out the code?

.github/workflows/release-on-tag.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build:
+    name: Create Release
+    runs-on: ubuntu-latest
```
Eloston, line 11: nit: Can we change this to use Debian instead of Ubuntu?

.github/workflows/release-on-tag.yml
zocker-160: you are right, it is not needed, I will remove it

.github/workflows/release-on-tag.yml
zocker-160: sadly no, only Ubuntu is supported right now, see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

Issue #949, opened 2020-02-24
Web Data not retaining user-specified search engines/Selection Context Search Extension results not opening


jlj2:
Using Ungoogled-chromium Portable Linux 80.0.3987.106-1.1 binary today, a pop-up appeared: “Your profile could not be opened correctly.  Some features may be unavailable."  Indeed, only the default search engines were available.

A [webpage from 2015](http://simplifyitco.com/google-chrome-profile-could-not-be-opened-correctly-error/) suggested that `~/.config/google-chrome/Default/Web Data` may be corrupt in Chrome for a similar notification;  perhaps the corresponding location for UC v80 would be  `~/.config/chromium/Default/Web Data`.

According to a recent [post](https://github.com/Eloston/ungoogled-chromium/issues/943), the Web Data file seems to store search engines.  Is it not encrypted by default in Ungoogled Chromium, along with some other config files, contrarily to some in Chromium?

Apparently, it can be opened with **Notepad++** (perhaps also **notepadqq** in Linux, a _'Notepad++-like editor for Linux'_).  However, a quick `cat` displays some unexpected '_credit_cards_' term in the second paragraph.  Perhaps it is safer to do the following check while offline?  The bash prompt has been changed below for safety's sake to '`myprompt ~$`';  it appears twice below:

```
myprompt ~ $ cat .config/chromium/Default/Web\ Data
SQLite format @  !.;ñöûö                                                                                                                   YSearxsearx.mehttps://searx.me/favicon.icohttps://searx.me/?q={searchTerms}UTF-8ch{google:pathWildcard}?ei={inputEncoding}&fr=crmas&p={searchU^øckDuckGoduckduckgo.comhttps://duckduckgo.com/favicon.icohttps://duckduckgo.co485bf7d3-0215-45af-87dc-538868000012[]https://searx.me/archTer^¾gÏ­,òÐ¢,¾óBBkc5bf7d3-0215-45af-87dc-538868000092[]https://duckduckgo.com/chrome_newtabEO]about:blank}!)Q[[]https://www.bing.com/images/detai%%tablecredit_cardscreditø°kCHROMI#enterInsightsimageBin={google:imageThumbnailBase64}https://www.bing.com/chrome/newtab~#5	
%%tablecredit_cardscredit_cards
[...]

PR #1444, opened 2021-03-19
Prepopulated search engine changes

### LINKED ISSUE #1432: Custom search engines disappear
**Describe the bug**
I use a large number of custom search engines. Unfortunately, the custom search engines disappear from time to time. It is a huge nuisance to have to enter them in each time they disappear.

**To Reproduce**
I don't have a list of steps to reproduce the behavior. But I have noticed that it seems to occur when the version is updated. For instance, the search engines disappeared when my distro automatically updated to 88.0.4324.182 from 88.0.4324.150. It also happened when 88.0.4324.150 was installed. Does Ungoogled Chromium delete any state when upgrading?

**Expected behavior**
Custom search engines should not disappear.

 - OS/Platform and version: NixOS 20.09.3436.6be212cce73
 - ungoogled-chromium version: 88.0.4324.150 and 88.0.4324.182

Ahrotahn:
Fixes #1432 (and possibly one of the issues in #949) where search engines were being reset on occasion.  

The search engines were being resynced when `kCurrentDataVersion` was updated in `components/search_engines/prepopulated_engines.json`.  
Changing the check to only do this when `resource_keyword_version` is zero makes it so that the prepopulated search engines are only added when the database hasn't been initialized yet (either a new profile or the `Web Data` database has been deleted).

This also simplifies adding search engines in `template_url_prepopulate_data.cc` by inserting them into the returned vector rather than inserting the search engines into each array individually.  DuckDuckGo exists in some arrays already, so the selected array is checked for an existing entry and only adds one if it doesn't already exist.


Eloston (approved):
Hmm, this seems like a small enough change to not really deviate too far from existing Chromium behavior.

LGTM :+1:

PR #1467, opened 2021-04-05
Keeping history saved for longer than 3 months

zoritle:
For a long time, Chrome didn't support to save browser history locally more than 90 days, they ignore the issue and just tell you to sign into google account and sync all your browser history to cloud.
https://bugs.chromium.org/p/chromium/issues/detail?id=500239
People even made up specific extensions to workaround the limit.
https://chrome.google.com/webstore/detail/history-trends-unlimited/pnmchffiealhkdloeffcdnbgdnedheme

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
@@ -1,0 +1,33 @@
+# Keep local history langer than 90 days
+
+--- a/chrome/browser/ungoogled_flag_entries.h
++++ b/chrome/browser/ungoogled_flag_entries.h
+@@ -60,4 +60,8 @@
+      "Popups to tabs",
+      "Makes popups open in new tabs.  ungoogled-chromium flag",
+      kOsAll, SINGLE_VALUE_TYPE("popups-to-tabs")},
++    {"keep-old-history",
++     "Keep old history",
++     "Keep history older than 3 months.  ungoogled-chromium flag",
++     kOsAll, SINGLE_VALUE_TYPE("keep-old-history")},
+ #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
+--- a/components/history/core/browser/history_backend.cc
++++ b/components/history/core/browser/history_backend.cc
+@@ -14,6 +14,7 @@
+ #include <vector>
+
+ #include "base/bind.h"
++#include "base/command_line.h"
+ #include "base/callback_helpers.h"
+ #include "base/compiler_specific.h"
+ #include "base/containers/flat_set.h"
+@@ -871,7 +872,8 @@ void HistoryBackend::InitImpl(
+   db_->GetStartDate(&first_recorded_time_);
+ 
+   // Start expiring old stuff.
+-  expirer_.StartExpiringOldStuff(TimeDelta::FromDays(kExpireDaysThreshold));
++  if (!base::CommandLine::ForCurrentProcess()->HasSwitch("keep-old-kistory"))
++    expirer_.StartExpiringOldStuff(TimeDelta::FromDays(kExpireDaysThreshold));
+
+ #if defined(OS_ANDROID)
+   if (backend_client_) {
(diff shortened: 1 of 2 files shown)


PF4Public:
Could it be a toggle'able flag?

Ahrotahn:
While 90 years is effectively disabled for most people, it looks like removing line 874 in the same file would truly remove the limit.  
I haven't tested it, but that's the only place `kExpireDaysThreshold` is used and the only place `StartExpiringOldStuff` is called.  

<br/>

For an example on how to set up a flag for the change you could take a look at [`add-flag-to-convert-popups-to-tabs.patch`](https://github.com/Eloston/ungoogled-chromium/blob/master/patches/extra/ungoogled-chromium/add-flag-to-convert-popups-to-tabs.patch).  It's also the last flag that has been added so the new flag entry would go right after that one.  
If you'd rather not do the flag setup then it might be a better idea to submit the patch to the [contrib repo](https://github.com/ungoogled-software/contrib) instead.

wchen342:
How much space will the history take? If they are only texts than this shouldn't  be a problem though.

Eloston:
I'd also like this to be behind a flag if we want to merge this into this repo instead of contrib.

I am not convinced that space is the only concern here: I am also worried about any potential performance regressions as a result of letting history grow very large. There may be several assumptions in the design of Chromium that rely on the assumption that the history cannot grow too large. For example, what would happen when searching the Omnibox with a very large history? How about `chrome://history`? What about other uses of history?

If someone can prove that a large history will not have a practical impact on performance or efficiency (through means of complexity analysis, benchmarking, or similarly rigorous methods), then I will agree this change does not need a flag.

---

From my perspective, a web browser is a critical piece of software in our modern society, so I don't think it is too unreasonable to demand high standards. Any issues that can arise from changes like these will take a while to manifest. Usually the symptoms are subtle, if they're noticed at all. Reports on such issues will be rare, and diagnosing rare problems are generally very challenging. I hope that clarifies my stance a little bit.

zoritle:
The history is stored in a sqlite database, include url and title, plus visit count,last visit time,etc.It shouldn't cost too much space in practice.

According to the benchmarks of sqlite, I don't think it will be a performance burden in this quntity of data(I have around 3,000 records per month)
Firefox and Vivaldi browser actually alow keeping history forerver. My Firefox history database is around 20MiB, and searching in history is instant.

Other use of history should have no problem,since this patch is just disable the background thread to automatically clean up old history.

As you wish, I tried to make up a flag,but need some help,because I don't have chromium source repo on local machine.

Ahrotahn:
It looks like only a few small changes are needed to pass validation.  The flag entry needed to be after the popups-to-tabs patch's entry and the last two context lines needed to be updated:
```diff
# Keep local history langer than 90 days

--- a/chrome/browser/ungoogled_flag_entries.h
+++ b/chrome/browser/ungoogled_flag_entries.h
@@ -60,4 +60,8 @@
      "Popups to tabs",
      "Makes popups open in new tabs.  ungoogled-chromium flag",
      kOsAll, SINGLE_VALUE_TYPE("popups-to-tabs")},
+    {"keep-old-history",
+     "Keep old history",
+     "Keep history older than 3 months.  ungoogled-chromium flag",
+     kOsAll, SINGLE_VALUE_TYPE("keep-old-history")},
 #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
--- a/components/history/core/browser/history_backend.cc
+++ b/components/history/core/browser/history_backend.cc
@@ -14,6 +14,7 @@
 #include <vector>

 #include "base/bind.h"
+#include "base/command_line.h"
 #include "base/callback_helpers.h"
 #include "base/compiler_specific.h"
 #include "base/containers/flat_set.h"
@@ -871,7 +872,8 @@ void HistoryBackend::InitImpl(
   db_->GetStartDate(&first_recorded_time_);
 
   // Start expiring old stuff.
-  expirer_.StartExpiringOldStuff(TimeDelta::FromDays(kExpireDaysThreshold));
+  if (!base::CommandLine::ForCurrentProcess()->HasSwitch("keep-old-kistory"))
+    expirer_.StartExpiringOldStuff(TimeDelta::FromDays(kExpireDaysThreshold));

 #if defined(OS_ANDROID)
   if (backend_client_) {
```

zoritle:
Thanks,it passed now.

Eloston:
I'll merge this after updating to version 90

PR #1491, opened 2021-04-16
Add macOS compatibility for set_quilt_vars.sh

fabriziobagala:


Eloston:
The macOS repo already has its own `set_quilt_vars.sh` file. Why do we need to note this here?

fabriziobagala:
> The macOS repo already has its own `set_quilt_vars.sh` file. Why do we need to note this here?

You are right, there is no need. I did not see the repository for macOS.

PR #1498, opened 2021-04-25
Docs+Patches: Add documentation for keep-old-history and fix typo

networkException:
I noticed the lack of documentation for the recently added "keep-old-history" flag.

Aditionally, both the commit introducing the flag (#1467) as well as a more recent refresh of the patch
included a typo in the flag check, making it unable to be toggled by `chrome://flags`

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
@@ -26,7 +26,7 @@
  
    // Start expiring old stuff.
 -  expirer_.StartExpiringOldStuff(TimeDelta::FromDays(kExpireDaysThreshold));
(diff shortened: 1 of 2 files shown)


patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
```diff
@@ -1,5 +1,5 @@
-# Keep local history langer than 90 days
-
+# Keep local history langer than 90 days
```
Eloston, line 1: Looks like "langer" is a typo too

Eloston (changes requested):
Good catch, thanks!

Found another typo

patches/extra/ungoogled-chromium/add-flag-to-disable-local-history-expiration.patch
networkException: oh, good catch

Eloston (approved):

PR #1518, opened 2021-05-11
Add link to flatpak version on Flathub

rugk:
Just adds a convenient link for users, who are looking for it.

We could also link [to their build scripts](https://github.com/flathub/com.github.Eloston.UngoogledChromium) additionally, but IMHO this is not really necessary.


Eloston (approved):
:+1:

Issue #1532, opened 2021-06-04
Having an offical portable Linux build of ungoogled-chromium


harry963:
Hello, I hope everyone is doing well during these hard times. This topic may have been mentioned before But I would really like an official portable build of ungoogled-chromium.
I cannot compile the browser due to low end hardware and I don't want to use some random binary that isn't official.
I think either just a regular portable build or an app image would be good maybe even add it to Guix or Nix since they work on every distro. This would allow for people with low end hardware or people who don't want to deal with the hassle of compiling to use ungoogled-chromium.
I also see many issues on GitHub of people asking to update certain binaries this would fix everyone asking for Linux binaries as they could just use this new portable binary. I think this build would benefit most ungoogled-chromium users who may be currently using a third party binary.
It would also allow for distros to add ungoogled-chromium to their repos since there is now an official binary especially smaller distros with smaller dev teams.
It wouldn't fix the issue for Windows, Mac or Linux with Musl (like Void musl or Alpine) users but it would certainly help many.
An official portable binary would be a very nice addition to have and I ask Eloston to please consider this in the future.

mdedonno1337:
One of the issues is the compute power needed to do the compilation, it would take arround 15$ per compilation on a Anazon AWS compute.
On some other build services you have limitation of time, cpu and ram which are problematic to compile.

harry963:
I could see how this could be an issue. Do you know how much it would cost per build? Maybe it would be possible for the community to fund it. 
Also, How are these binaries usually made? I would've assumed that most of these third party binaries are just complied off regular consumer hardware.
Also, it wouldn't have to be hosted on Guix. If a portable build was available, then distros could package it in their repo or people could just download and run the binary

PF4Public:
I'm building [binaries of ungoogled-chromium](https://github.com/PF4Public/ungoogled-chromium-bin/releases) for Gentoo, but they are easily useable in virtually any Linux, which provides the same versions of libraries, that I used to build against in Gentoo.

>This topic may have been mentioned before
True and the conclusion was the difficulty to define "official"? Are my binaries official enough? I see some folks use them (based on the issues filled every now and then about them).

harry963:
I think official would mean built by Eloston himself. I would very much like to use ungoogled-chromium however I am wary of the third party builds, and I feel many others feel the same way. I honestly think that ungoogled-chromium would have a much larger user base if it had a proper official build.

[...]

PR #1534, opened 2021-06-04
Official portable executable

mbakke:
Hi!

This adds a scheduled GitHub action that uploads an "official" portable Linux executable using GNU Guix.

It addresses #1532 and others.

Guix has a little-known feature to create relocatable binaries of any of its 17000+ packages with almost zero overhead.  This makes use of that feature such that the tarballs can be unpacked and run from anywhere.

It runs daily in order to get security updates, and aborts if the Guix CI system does not have an available "substitute" (build),  because Guix automatically falls back to building locally which will time out on the GitHub CI.

About the README diff: I actually did not change the third line, but Git is thorougly confused because the two sentences start similarly.

### DIFF
--- a/.github/workflows/scheduled-portable.yml
+++ b/.github/workflows/scheduled-portable.yml
@@ -1,0 +1,74 @@
+name: Create portable executable
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '40 3,15 * * *'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Out Repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Guix cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/guix
+          key: guix-cache-${{ github.sha }}
+          restore-keys: |
+            guix-cache-
+      - name: Install Guix
+        uses: PromyLOPH/guix-install-action@v1
+      - name: Assert that substitute is available
+        run: guix weather ungoogled-chromium
+      - name: Download substitutes
+        run: guix build --no-grafts --fallback ungoogled-chromium
+      - name: Install Recutils
+        run: sudo apt-get -q install recutils
+      - name: Export Guix version
+        run: >
+          echo version=$(guix show ungoogled-chromium | recsel -P version)
+          >> $GITHUB_ENV
+      - name: Create Guix pack
+        run: >
+          guix pack --fallback -RR -S /bin=bin
+          --root=ungoogled-chromium-${{ env.version }}.tar.gz
+          ungoogled-chromium
+      - name: Export Guix commit
+        run: >
+          echo shortcommit=$(echo ${{ env.version }} | cut -d- -f2 | cut -d. -f2)
+          >> $GITHUB_ENV
+      - name: Export long commit
+        run: echo commit=$(git rev-parse ${{ env.shortcommit }}) >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


mbakke:
I should mention that there will be situations where Guix lags behind official tags (rarely more than a couple of days), and also situations where Guix will be ahead of ungoogled-chromium (when @Eloston is busy and a new release is available).  So the uploaded release may not match the tag exactly.

networkException:
First of all, packages for distributing ungoogled chromium have historically been a part of the [ungoogled-software](https://github.com/ungoogled-software) organisation. As your workflow is just a cron job I don't see why it has to be in the *main* repository, mainly containing patches and scripts to build ungoogled chromium.

Then what do you mean by "ahead of ungoogled-chromium"? If the patches in this repository don't get updated to a new version, its unlikely that it would compile. As such there is no "being ahead of ungoogled-chromium".

mbakke:
With "ahead" I mean that if there is a new (minor) Chromium release, Guix will occasionally update to it before this repository has been updated, as long as the patches still apply.

Happy to move the scheduled job to a different repository, but I was not sure where.

mbakke:
Oh, and if anyone wants to test the portable Guix build, it can be downloaded from here:

https://github.com/mbakke/ungoogled-chromium/releases/download/72.0.3626.122-1/ungoogled-chromium.tar.gz

Despite the 72 version in the URL, it is actually 91.0.4472.77.

PF4Public:
If anyone asks me, I would reject this PR and redirect it into a separate platform repo.

mbakke:
Instead of uploading a "random" version to the latest tag, one alternative could be to have a `latest` or `portable` tag which the scheduled job uploads to.

mbakke:
I pushed a new commit that creates a `latest-portable` tag instead of hap-hazardly uploading the latest Guix version to the latest tag.

The version number in Guix actually includes the ungoogled-chromium commit ID it was built with, which is used to create the `latest-portable` tag.

Note: there is currently a bug that prevents the `latest-portable` tag from being updated.  I will fix that and squash these commits before removing the draft status if this approach looks sane.  Awaiting further feedback.

Eloston:
I don't really understand the use-case of this change. Aside from the glaring problem that we don't put releases on this repo, why wouldn't users just install directly from Guix? Or why couldn't they use the AppImage or Flatpak builds?

mbakke:
This is in response to #1532 and others where people have asked for a downloadable executable.

It is intended for users who can't or don't want to install Guix, Flatpak or whatever.  It has zero external dependencies and does not require root privileges to install or run. Perhaps it's better to target https://github.com/ungoogled-software/ungoogled-chromium-binaries instead?

Eloston:
[AppImage](https://en.wikipedia.org/wiki/AppImage)'s primary objective is to have zero external dependencies and does not require root privileges to install or run. This is already available on the downloads page.

Also let's be clear here: From my perspective, #1532 is fundamentally an issue of trust. It is NOT a packaging issue. The issue of trust has been discussed before, it will never be solved (i.e. as good as inspecting the code and building the binary yourself), but it can be mitigated (e.g. reproducible binaries, CI/CD builds on trusted platforms with enough visibility, signing by trusted/regular members). I think the better course-of-action is to document a more in-depth explanation of this issue and the status quo.

By the way, if your change should target something, it should be Portable Linux. Let's not forget that ungoogled-chromium also runs on macOS, Windows, and Android, which are all quite different from Desktop Linux.

PF4Public:
This PR does not fit this repository, so I'm closing it.

PR #1537, opened 2021-06-07
Update README.md

briansoule:
I helped a friend install on a fresh MacOS Big Sur instance, and it didn't seem to require `brew fetch --cask eloston-chromium` to run `brew install --cask eloston-chromium`. In fact it erred. Can anyone corroborate?


Eloston (approved):
I think you're right, I don't recall ever needing the `fetch` subcommand.

PR #1565, opened 2021-07-11
Add flag to prevent autoplay

Ahrotahn:
This PR adds a flag that, when enabled, attempts to prevent autoplaying media from starting.  
The flag can be enabled in `chrome://flags` or by adding `--enable-feature=PreventAutoplay` to the launch options.  

<br>

This is a patch that I've been iterating on for over half a year.  I've tested it on most major websites and it seems to work well for my use-case!  There's no way for me to test every possible site though, so I've submitted this as a draft until enough other people have tested this and verified it works correctly.  I'd be surprised if there aren't any edge-cases, so if you try this out please post what sites worked, didn't work, or had issues with the flag enabled.  

I'll post a followup comment in a bit that will break down what the patch is doing and my reasoning for implementing things this way.

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -51,4 +51,5 @@ These are also available on the `chrome://flags` page.
 
   <code>Feature&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;</code> | Description
   -- | --
+  `PreventAutoplay` | Prevents media from playing automatically.
   `SetIpv6ProbeFalse` | Forces the result of the browser's IPv6 probing (i.e. IPv6 connectivity test) to be unsuccessful. This causes IPv4 addresses to be prioritized over IPv6 addresses. Without this flag, the probing result is set to be successful, which causes IPv6 to be used over IPv4 when possible.
--- a/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch
@@ -1,0 +1,160 @@
+--- a/chrome/browser/ungoogled_flag_entries.h
++++ b/chrome/browser/ungoogled_flag_entries.h
+@@ -64,4 +64,8 @@
+      "Keep old history",
+      "Keep history older than 3 months.  ungoogled-chromium flag",
+      kOsAll, SINGLE_VALUE_TYPE("keep-old-history")},
++    {"prevent-autoplay",
++     "PreventAutoplay",
++     "Prevents media from playing automatically.  ungoogled-chromium flag.",
++     kOsAll, FEATURE_VALUE_TYPE(media::kPreventAutoplay)},
+ #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
+--- a/media/base/media_switches.cc
++++ b/media/base/media_switches.cc
+@@ -859,6 +859,8 @@ const base::Feature kUseFakeDeviceForMed
+ const base::Feature kBresenhamCadence{"BresenhamCadence",
+                                       base::FEATURE_DISABLED_BY_DEFAULT};
+ 
++const base::Feature kPreventAutoplay{"PreventAutoplay", base::FEATURE_DISABLED_BY_DEFAULT};
++
+ bool IsVideoCaptureAcceleratedJpegDecodingEnabled() {
+   if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+           switches::kDisableAcceleratedMjpegDecode)) {
+--- a/media/base/media_switches.h
++++ b/media/base/media_switches.h
+@@ -236,6 +236,8 @@ MEDIA_EXPORT extern const base::Feature
+ MEDIA_EXPORT extern const base::Feature kDeprecateLowUsageCodecs;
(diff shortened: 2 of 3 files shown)


Ahrotahn:
> The Programmers’ Credo: We do these things not because they are easy, but because we thought they were going to be easy.
> \- Pinboard

So preventing autoplaying media is one of those things that seems like it would be an easy problem to solve, but anyone that has attempted to work on this can vouch for the complexity of the issues that lie under the surface.  This is the simplest method I have been able to create so far that has the fewest pitfalls.  


One of the first problems is that media can be controlled through javascript.  This means that custom controls can be created for the media, and it also means you can't just check if the media element receives user input since those controls can exist outside of the element.  It's incredibly difficult to tell if a user initiated play() or if some script on the page is calling it without the user's consent.  There is a feature called UserActivation that was set up as a means to detect if the user has interacted with a page so that popups and things like autoplaying media could be blocked.  There is even a policy to prevent autoplaying unmuted videos if there is no prior user activation.  The problem is that this has been slowly been reduced to uselessness over time.  While debugging why pages would gain activation without any input I'd find activation being automatically granted in many different areas, [some even were marked with issues where the devs were aware that fake activations were being made](https://crbug.com/1082273).  I had attempted to resolve these issues, but in the end decided to simply implement our own method of detecting real user input:  RealUserActivation.  This also means that any changes made in the future to UserActivation shouldn't impact this patch.  


https://github.com/Eloston/ungoogled-chromium/blob/b5eeef268dd2edf123a76f2423e001f068adadc3/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch#L103-L160

These are points in the input handling code where real user input is taken into account.  The one in `media_session.cc` is needed to notify user input in the global media menu.  


https://github.com/Eloston/ungoogled-chromium/blob/b5eeef268dd2edf123a76f2423e001f068adadc3/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch#L34-L43

Real user activation is reset when the URL changes, which seems to be the context-switch most people expect that to happen on.  


https://github.com/Eloston/ungoogled-chromium/blob/b5eeef268dd2edf123a76f2423e001f068adadc3/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch#L44-L63

Attempting to figure out *what* is being clicked would make this patch an order of magnitude larger.  HasRealUserActivation simply checks if it was called within a second of the last recorded user input.  It's a rather dumb check, but surprisingly effective.  The downside to this method would be for situations where a media element hasn't finished loading and the user has interacted with the page a second before it attempts to play.  Luckily this seems to be a rare occurrence.  
`InSeconds() < 1` is a decent window, but could be lowered and still work fine.  I had tested with `InMilliseconds() < 500` on an old T420 laptop and it still worked well.  
Once play has been allowed, that state will persist until ClearRealUserActivation is called.  This makes it so that you won't have to click play again if you resize the player or if a stream is interrupted.  


https://github.com/Eloston/ungoogled-chromium/blob/b5eeef268dd2edf123a76f2423e001f068adadc3/patches/extra/ungoogled-chromium/add-flag-to-prevent-autoplay.patch#L64-L91

This nullifies the autoplay attribute and performs a check on play() to be sure that it is being called within a second of user input.  
Due to autoplay policies that allowed videos to play if muted, some sites will attempt to play and if not allowed will then mute the video and try again.  This meant that some sites would always be muted when clicking play.  To deal with this `rua_mute_state_` is being used as a tribool to keep track of the mute state when an autoplay attempt happens.  When the first failed play attempt happens the mute state is recorded and that state is set back when the play is allowed.  This makes it so that the video is in the same muted state it was in initially before the autoplay attempt.

csagan5:
@uazo contributed [this patch](https://github.com/bromite/bromite/blob/91.0.4472.158/build/patches/Revert-the-removal-of-an-option-to-block-autoplay.patch) to Bromite with the same purpose

Ahrotahn:
Initially I had tried a change similar to that patch and how brave handled their autoplay blocker.  The problem is that it relies on UserActivation, which while seemingly functional on android, didn't work well at all on desktop.  If I recall correctly there were even cases where extensions accessing the page would be counted as user activation.  

Another issue (for my use case) when using UserActivation is activation is counted across a site for a session.  So for example a youtube video would have it's autoplay blocked when visiting the link directly, but if you clicked through to that video from the channel page or home page then the video would be allowed to autoplay since you've interacted with the site.  This PR is more strict and will block autoplay on each page until activation.

Eloston:
@Ahrotahn Why was this closed? Is it being reworked?

Ahrotahn:
It didn't seem to garner much interest, so I figured I'd submit it to contrib later after updating it for 92.

jacob755:
this is good flag to be in ungoogled-chromium Please add this

PF4Public:
@jacob755 Please see https://github.com/ungoogled-software/contrib/blob/master/functionality/prevent-autoplay.patch

ioc2e3:
> ... there were even cases where extensions accessing the page would be counted as user activation.
> 
> Another issue (for my use case) when using UserActivation is activation is counted across a site for a session. So for example a youtube video would have it's autoplay blocked when visiting the link directly, but if you clicked through to that video from the channel page or home page then the video would be allowed to autoplay since you've interacted with the site. This PR is more strict and will block autoplay on each page until activation.

Now , what should do ?

.

PR #1590, opened 2021-08-05
Specify data in add-flag-to-clear-data-on-exit.patch

Ahrotahn:
This PR updates `add-flag-to-clear-data-on-exit.patch` to specify the data to be deleted rather than deleting everything.  

It was brought up in the original pull request that this had also deleted bookmarks.  It's reasonable to expect that enabling the flag would only delete the same data that would be removed when deleting 'Everything' from the 'Clear Browsing Data' option in the settings.


Eloston (approved):
Thanks!

Issue #1574, opened 2021-07-21
Add flag to disable sharing qr code generator


networkException:
**Is your feature request related to a problem? Please describe.**
In chromium versions prior to 92 there existed a flag to disable the then recently introduced qr code generator for sharing. Personally the addition of the corresponding context menu entry as well as omnibar icon does not add any value to the browser and I would prefer to have a choice to disable them.

**Describe the solution you'd like**
The addition of a flag to remove ui elements for the sharing qr code generator.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Ahrotahn:
I can group this in with the PR for the tabsearch button removal patch once I get them set up.  
Until then the feature can be removed by applying this patch:
```diff
--- a/chrome/browser/ui/qrcode_generator/qrcode_generator_bubble_controller.cc
+++ b/chrome/browser/ui/qrcode_generator/qrcode_generator_bubble_controller.cc
@@ -22,10 +22,7 @@
 
 // static
 bool QRCodeGeneratorBubbleController::IsGeneratorAvailable(const GURL& url) {
-  if (!url.SchemeIsHTTPOrHTTPS())
     return false;
-
-  return true;
 }
 
 // static
--- a/chrome/browser/ui/views/location_bar/location_bar_view.cc
+++ b/chrome/browser/ui/views/location_bar/location_bar_view.cc
@@ -245,7 +245,6 @@
     // the left most icon.
     params.types_enabled.push_back(PageActionIconType::kSendTabToSelf);
     params.types_enabled.push_back(PageActionIconType::kClickToCall);
-    params.types_enabled.push_back(PageActionIconType::kQRCodeGenerator);
     if (base::FeatureList::IsEnabled(kSharedClipboardUI))
       params.types_enabled.push_back(PageActionIconType::kSharedClipboard);
     if (base::FeatureList::IsEnabled(kWebOTPCrossDevice))
```

networkException:
amazing, thank you!

BachoSeven:
I think this can be closed since #1591 was merged

PR #1591, opened 2021-08-05
Add flags for tabsearch, QR generator, and grab handle

### LINKED ISSUE #1553: Remove the tab search feature
The tab search feature in chromium 91 takes useless space at the top. 
In chromium 90, the feature could be disabled in chrome://flags, which is not the case in the last version.

If possible, give a toggle in chrome://flags to remove it for users who don't want this feature.

Thanks

Ahrotahn:
This PR contains two commits, the first adds a flag that removes the tabsearch button (and it's keyboard shortcut), the second adds a flag to disable the QR generator for sharing page links.  

Fixes #1553 and #1574


PF4Public:
Awesome!
image
image

Did you avoid removing this empty padding on purpose? Could it be also incorporated into the patch?

Ahrotahn:
I have a separate patch for removing the drag handle space, but I needed to do some more testing before submitting it.  
It turns out that the flex layout compresses the tip marquee's space regardless of it's preferred size and the added padding, so the only change needed is the resizing of the handle.  

I'll add that patch to the PR soon once I get some time.  @Eloston I hope you don't mind me grouping up these patches like this!  Doing it this way makes it so that their merge order doesn't need to be considered since the flags need to be added sequentially.

PF4Public:
With the last patch it looks like this:
image

Eloston:
> @Eloston I hope you don't mind me grouping up these patches like this! Doing it this way makes it so that their merge order doesn't need to be considered since the flags need to be added sequentially.

I don't mind, thanks for letting me know!

Eloston (approved):
LGTM :+1:

PR #1592, opened 2021-08-05
Add Windows compatibility for prune_binaries.py

Nifury:
`file_path.unlink()` is not able to remove read-only files on Windows.

### DIFF
--- a/utils/prune_binaries.py
+++ b/utils/prune_binaries.py
@@ -10,7 +10,9 @@ import argparse
 from pathlib import Path
 
 from _common import ENCODING, get_logger, add_common_params
-import sys, os, stat
+import sys
+import os
+import stat
 
 
 def prune_dir(unpack_root, prune_files):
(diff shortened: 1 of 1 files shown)


utils/prune_binaries.py
```diff
@@ -10,7 +10,7 @@
 from pathlib import Path
 
 from _common import ENCODING, get_logger, add_common_params
-import sys
+import sys, os, stat
```
Eloston, line 13: nit: Can we have these imports on separate lines?

utils/prune_binaries.py
```diff
@@ -25,6 +25,9 @@ def prune_dir(unpack_root, prune_files):
         file_path = unpack_root / relative_file
         try:
             file_path.unlink()
+        except PermissionError:
+            os.chmod(file_path, stat.S_IWRITE)
```
Eloston, line 29: nit: Add a comment saying this is for Windows

Issue #1597, opened 2021-08-09
Is there a way to remove those requests rather than redirecting them?


navaneethkm004:
Everytime ungoogled-chromium just redirects those request like chrome webstore(https://chrome.9oo91e.qjz9zk/) and google services(https://support.9oo91e.qjz9zk/chrome?p=syncgoogleservices).

Is there a way to remove them rather than redirecting them to qjz9zk domain?

networkException:
Replacing the urls with that domain is basically removing them, the domain will no be resolved as it doesn't exist. Removing all functionally related to every google url in the source would be impossible to maintain, replacing the urls with something other than valid urls might cause undefined behavior.

PF4Public:
> doesn't exist

Unless Google registers it :D

PF4Public:
> 1. Can you completely remove these, cus its no use as it is of now, so why not remove it

The only reason I can think of not to do this is the added maintenance burden for no significant benefit.

navaneethkm004:
Well, I can't deny that. Umm, can you help me with the second part?

PF4Public:
> can you help me with the second part?

Unfortunately, not, not me at least.

Ahrotahn:
I can dig into this a bit and see if it's something that can be added to `remove-uneeded-ui.patch`.  There's no reason to remove every instance of substituted links in the ui, but it's nice to remove the more egregious areas.  The passwords section is something I've been aware of and put off for too long.

navaneethkm004:
> I can dig into this a bit and see if it's something that can be added to `remove-uneeded-ui.patch`. There's no reason to remove every instance of substituted links in the ui, but it's nice to remove the more egregious areas. The passwords section is something I've been aware of and put off for too long.

:) nice bro

navaneethkm004:
@Ahrotahn Hey, I need a small help. I downloaded chromium using depot_tools earlier. But when i checked the modification readme, i saw that i have no option to use this downloaded one. I'm not that experienced, so can you help me with using the downloaded one. #

Ahrotahn:
I'd recommend checking out the instructions on the repo for your platform:  https://github.com/Eloston/ungoogled-chromium/blob/master/docs/platforms.md

There's usually a writeup on the build process that's easier than attempting to set up your own.

PF4Public:
I hope your question has been answered, otherwise please let us know.
Closing this issue for now.

Issue #1514, opened 2021-05-07
Live caption don't work


Tellayi:
**Describe the bug**
Live caption don't download nor work even after enabling the setting.

**To Reproduce**
1. Go to chrome://flags and enable #enable-accessibility-live-caption
2. Relaunch the browser
3. Go to settings -> advanced settings -> accessibility and enable the live captions
4. Nothing happens, chromium doesn't download the subtitles and they don't show up.

**Expected behavior**
Chromium should download the subtitles and allow live captions to work.

**Environment (please complete the following information):**
 - OS/Platform and version: Windows 10
 - ungoogled-chromium version: 89.0.4389.114-1

**Additional context**
It's worth noting that live captions aren't a privacy threat at all, because they work fully offline.

wchen342:
Probably because of what mentioned in brave/brave-browser/issues/15864 and domain substitution. If so then not likely to work unless the download link is whitelisted in domain substitution.

Issue #605, opened 2018-11-27
Remove or replace the "Learn more" link from the "You've gone incognito" page


anchev:
**Is your feature request related to a problem? Please describe.**
The "Learn more" link in the "You've gone incognito" page leads to trk:247:https://support.9oo91e.qjz9zk/chrome/?p=incognito - obviously a non-existing address.

**Describe the solution you'd like**
Remove the "Learn more" link.

**Describe alternatives you've considered**
Replace the link and point it to a page in the documentation of ungoogled-chromium explaining better what incognito does (not).

qvint:
What about completely removing all broken links including help pages? (e.g. F1)

anchev:
> How about completely removing all broken links from Chromium including its online help pages?

I have thought about that too but it may also be confusing for someone who may have just started using UC. Perhaps it will be better to have them display a UI element saying "In UC all links pointing to Google have been removed deliberately". Then it will not create the impression of something broken. Ideally the info from each link should be available inside the browser itself, not on some external host, but that may be too much work. Also some of that info may not apply to UC itself, so it would need additional maintenance.

qvint:
I don't see much difference between _"support.9oo91e.qjz9zk Blocked by client"_ and _"Link is deliberately not working"_ messages... In both situations user will get the impression of something [deliberately] broken. If we remove the links, he won't see any error message at all--there will be no link to click.

anchev:
> "Link is deliberately not working" messages...

But I have never suggested that. Deliberaterly removed != deliberately not working.

Also a layman doesn't know what a "client" is. So any message should be clear enough for everyone, not just for experts. Currently it is not, hence the issue.

Eloston:
I could change the error message for [`BLOCKED_BY_CLIENT`](https://chromium.googlesource.com/chromium/src/+/70.0.3538.110/net/base/net_error_list.h#86) (which I don't like too much because it is conflating a standard error with an error we introduced), or I could [add a new `NET_ERROR`](https://chromium.googlesource.com/chromium/src/+/70.0.3538.110/net/base/net_error_list.h) that has its own message. I'd like to leave the links in because it's nice to know where to look for more information about certain features.

Alternatively, I could write a system that does something crazier:

1. Find all `support.google.com` help URLs in the source code
2. Download these help pages and modify them to depend on local resources only (also maybe run DOM Distiller over them?)
3. Dump all these documentation files into an extension (or embeds them into Chromium itself)
[...]

PR #1604, opened 2021-08-13
Updates to `remove-uneeded-ui.patch`

Ahrotahn:
This PR makes some small changes to `remove-uneeded-ui.patch`:

* Removes the 'Learn more' link on new incognito tabs.
* Removes the live captions entry in the settings.  It's possible that this could be enabled manually somehow, so this change only removes the entry on the settings page.  The caption settings can still be reached at `chrome://settings/captions`.  
* Removes the external link for theme entry on settings page.
* Removes the password check and managed password entries from the password page.
* Removes the (?) learn more buttons from settings pages.
* Removes the 'Larn more' link from the search engines entry on the settings page.
* Updated the safetycheck removals to disable their templates.

Addresses #1597, #1514, and #605


Eloston (approved):
Very thorough, thanks!

Issue #1624, opened 2021-09-01
Updating to Chromium 93.0.4577.82


Zoraver:
## Notes for the developer

Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:

![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)

Finally, make sure to reference this issue in your PR.

Feel free to ask questions and discuss issues here along the way. Thanks!

## Notes for others

Feel free to raise issues or questions throughout the process. However, please refrain from asking ETAs unless no visible progress has been made here or on the developer's PR for a while (e.g. 2 weeks).

wchen342:
It seems we can use `python3` to compile chromium now, but it may still need some minor fixes.
(I'm not working on it btw, just here to make a quick heads up).

Zoraver:
I plan on working on this.

Ahrotahn:
Thanks for picking this up Zoraver!  Hurricane Ida left me without electricity since last Sunday, so apologies for the radio-silence.  
I had started working on the update a bit before the storm, if you haven't started yet I could put my partial changes up on my update branch for you to work with or I could continue where I left off if you'd rather have your free time back 😅.

Zoraver:
@Ahrotahn no need for apologies. I hope you weathered the storm well and are in good health. 

Regarding ungoogled-chromium 93, I am pretty far along and am on track to have a pull request up sometime late tonight (or early tomorrow morning).

networkException:
The related pull request has been merged

Vegalord11:
Hey, I'm not a developer but just a average user. I came here when I was searching for the latest update of chromium.

PR #1635, opened 2021-09-06
Update to Chromium 93.0.4577.82

Zoraver:
Addresses #1624. I will let you know once I have a successful clean build.

Notes:
Upstream made the generation of `download_file_types.pb.h` conditional on safe browsing being enabled in some capacity, but left plenty of usage of `download_file_types.pb.h` not conditional on safe browsing being enabled. Considering that it simply contains a few type definitions and has been included in previous releases of ungoogled-chromium, I pragmatically disabled the conditional and made it so that `download_file_types.pb.h` is always generated, regardless of whether safe browsing is enabled. The alternative was to go through and disable all references to `download_file_type.pb.h` and the types it contains, which would be far more invasive and would take more effort to maintain going forward.

Speaking of invasive patches, `remove-unused-preferences-fields.patch` is now over 5,000 lines long. I personally don't think it is sustainable and have been thinking about an alternate strategy that should accomplish the same goal and be easier to maintain. I plan to elaborate further at a later time.


Ahrotahn:
> Probably `add-flag-for-omnibox-autocomplete-filtering.patch` needs to be fixed

Using the std implementation should work:
```diff
--- a/patches/extra/ungoogled-chromium/add-flag-for-omnibox-autocomplete-filtering.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-for-omnibox-autocomplete-filtering.patch
@@ -49,9 +49,9 @@
 +    const std::string flag_value = base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("omnibox-autocomplete-filtering");
 +    provider_types &= AutocompleteProvider::TYPE_KEYWORD | AutocompleteProvider::TYPE_SEARCH |
 +        AutocompleteProvider::TYPE_HISTORY_URL | AutocompleteProvider::TYPE_BOOKMARK | AutocompleteProvider::TYPE_BUILTIN;
-+    if (!base::Contains(flag_value, "bookmarks"))
++    if (flag_value.find("bookmarks") == std::string::npos)
 +      provider_types &= ~AutocompleteProvider::TYPE_BOOKMARK;
-+    if (!base::Contains(flag_value, "chrome"))
++    if (flag_value.find("chrome") == std::string::npos)
 +      provider_types &= ~AutocompleteProvider::TYPE_BUILTIN;
 +  }
    if (provider_types & AutocompleteProvider::TYPE_BOOKMARK)
@@ -104,7 +104,7 @@
      FixupHost(trimmed, parts.host, parts.scheme.is_valid(), desired_tld, &url);
      if (chrome_url && !parts.host.is_valid())
 +     if (!base::CommandLine::ForCurrentProcess()->HasSwitch("omnibox-autocomplete-filtering") ||
-+         base::Contains(base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("omnibox-autocomplete-filtering"), "chrome"))
++         base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII("omnibox-autocomplete-filtering").find("chrome") != std::string::npos)
        url.append(kChromeUIDefaultHost);
      FixupPort(trimmed, parts.port, &url);
      FixupPath(trimmed, parts.path, &url);
```

networkException:
Built and runs fine for me :+1:

Zoraver:
My build succeeded and is running well. I did run into and fix the issue that PF4Public ran into, but I mistakenly put the fix into `fix-building-without-safe-browsing.patch`. I will move it to `add-flag-for-omnibox-autocomplete-filtering.patch` shortly.

Ahrotahn:
Builds fine on my end.  Runs fine with the exception that the print preview issue is back.  

`chrome/browser/resources/print_preview/data/destination_store.js` and `chrome/browser/resources/print_preview/ui/destination_settings.js` attempt to use `forceEnablePrivetPrinting`, but it's value is never set.

Ahrotahn (approved):
Looks good to me!

PF4Public:
It requires two reviews. I wonder if Zoraver himself should also review his own PR :D

networkException (approved):
The printing issue is fixed as well, good work!

(I know this review doesn't help with merging but its intended to approve changes with a review so why not :^)

kramred:
Works as well on macOS after updating the OS specific patches

PF4Public:
Come on, guys! Are we still waiting for Eloston? What was the purpose of obtaining write permissions? Should I ask for write permissions for myself in order to at least have one more approving voice?

Zoraver:
@PF4Public pull request authors cannot approve their own pull requests ([see here](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews)). Additionally, for major changes, I would appreciate @Eloston's feedback.

wchen342:
> pull request authors cannot approve their own pull requests
Ah right, so if each PR needs two approvals we need at least three people. @PF4Public will be good.

wchen342:
@Zoraver please add the following for `chrome/browser/policy/browser_signin_policy_handler.cc` to `remove-unused-preferences-fields.patch`:

```
@@ -47,28 +47,8 @@
 #endif
       FALLTHROUGH;
     case BrowserSigninMode::kEnabled:
-      prefs->SetValue(
-#if defined(OS_ANDROID)
-          // The new kSigninAllowedOnNextStartup pref is only used on Desktop.
-          // Keep the old kSigninAllowed pref for Android until the policy is
-          // fully deprecated in M71 and can be removed.
-          prefs::kSigninAllowed,
-#else
-          prefs::kSigninAllowedOnNextStartup,
-#endif
-          base::Value(true));
       break;
-    case BrowserSigninMode::kDisabled:
-      prefs->SetValue(
-#if defined(OS_ANDROID)
-          // The new kSigninAllowedOnNextStartup pref is only used on Desktop.
-          // Keep the old kSigninAllowed pref for Android until the policy is
-          // fully deprecated in M71 and can be removed.
-          prefs::kSigninAllowed,
-#else
-          prefs::kSigninAllowedOnNextStartup,
-#endif
-          base::Value(false));
+    case BrowserSigninMode::kDisabled:;
       break;
   }
 }
```

PF4Public:
Stable updated to 93.0.4577.82

Zoraver:
I am compiling ungoogled-chromium 93.0.4577.82 now and will let you know once I have a successful build.

@wchen342 I have run into that bit of code before and decided at the time not to remove it because the `kSigninAllowedOnNextStartup` preference field is not removed. `kSigninAllowed`, on the other hand, is removed, but this is only a problem on Android. I figured that any Android-specific fixes should reside in the ungoogled-chromium-android patch set. This is simply because any changes upstream makes to Android-specific code that break ungoogled-chromium-android are unlikely to be caught during the process of updating the core patch set from one version of Chromium to the next. If you feel strongly about this issue please elaborate on your thinking and I may reconsider.

wchen342:
>  kSigninAllowed, on the other hand, is removed, but this is only a problem on Android.

In that case, why is the field removed in the first place? Can it be kept instead? If it is not used on desktop then it is better leave it untouched. I feel it is strange to have part of a field removed in one patch, and then leave the removal of its references to another patch, which is confusing.

> any changes upstream makes to Android-specific code that break ungoogled-chromium-android are unlikely to be caught during the process of updating the core patch set 

I guess the root problem is we don't have a CI. But for the time being, since we are relying on manual reports anyway I think android shouldn't be treated much different from Linux.

PF4Public:
[Builds](https://github.com/PF4Public/ungoogled-chromium-bin/releases/tag/93.0.4577.82) and runs fine:
image
Print preview is also OK.

networkException:
*The two horseman of regression testing: Playing a YouTube video and opening the print view*

Jokes aside, builds and runs fine on my end as well :+1:

Zoraver:
My build succeeded and is running well. One thing to note is that Chromium 93.0.4577.82 fixes two zero-days that are being actively exploited...

@wchen342: My understanding of `remove-unused-preferences-fields.patch` is that it exists to fix building without the `safe_browsing_prefs.*` and `signin_pref_names.*` files (see `PRUNING_INCLUDE_PATTERNS` in `update_lists.py`). The reason `kSigninAllowed` is removed while `kSigninAllowedOnNextStartup` is not is that `kSigninAllowed` is declared in `signin_pref_names.h`, which is pruned, whereas `kSigninAllowedOnNextStartup` is declared in `chrome/common/pref_names.h`, which is not pruned.

elkablo:
Why does this take so long to get merged? I understand that sometimes there are things that could be improved regarding the ungoogling, but if the code builds and runs, shouldn't it get released ASAP, since it contains security fixes from upstream Chromium? Improvements can be done in another release, if needed.

PF4Public:
> Why does this take so long to get merged?

This is an ongoing issue, @Eloston promised to come up with a solution, but is unfortunately very busy at the moment.

codeswhite:
I use this 93.*.82 build for a few days now thanks to @PF4Public who pushed it ASAP into gentoo package repository!
The build is stable.

Eloston:
Hi everyone, sorry for the long delay. I wanted to get to this last weekend, but honestly any excuse I can make won't help the fact that I am a bottleneck on this project's development speed.

Since a number of you have been around for a while and/or putting out quality work, I trust you guys will make good collaborators too. I will be adding more of you guys as collaborators so you have permissions to approve changes and get it merged. That also means future PRs will no longer require all CODEOWNERS to approve it either, since if a PR author is a code owner, I don't think it's possible for the code owner to approve their own PR. I aim to add everyone who fits the aforementioned criteria and has been recently active, but please let me know if I forget to send you an invite if you think you fit the criteria.

To those I invite, I will also be adding you to a new team in the ungoogled-software repo where we can have some internal co-ordination as needed.

Thank you all for your understanding and patience during this transition. I really hope this will help keep the project going strong.

PR #1639, opened 2021-09-13
WIP: Fix compiling with ThinLTO optimizations via the use_thin_lto=true flag

elkablo:
For some reason, building with ThinLTO (via the use_thin_lto=true flag)
fails when ungoogled-chromium patches are applied.

The failure is caused by undefined references to various crashpad
and breakpad symbols when linking.

It seems that building some of the crashpad/breakpad libraries without
LTO fixes this issues.

This adds two patches:
- the first allows for -Wl,--lto-O2 optimizations instead of just
  -Wl,--lto-O0 even if not official building
- the second separates the ThinLTO stuff from config("compiler") into
  config("thinlto"), adds this new config into default_compiler_configs,
  and then removes this config from the relevant crashpad and breakpad
  libraries, so that these libraries are build without LTO

@Zoraver @Ahrotahn @PF4Public could you please review this?

### DIFF
--- a/patches/extra/ungoogled-chromium/allow-thinlto-optimizations.patch
+++ b/patches/extra/ungoogled-chromium/allow-thinlto-optimizations.patch
@@ -1,0 +1,17 @@
+From: Marek Behún <kabel@kernel.org>
+
+Use -Wl,--lto-O2 optimizations instead of just -Wl,--lto-O0 even if not
+official build.
+
+diff -Naurp a/build/config/compiler/BUILD.gn b/build/config/compiler/BUILD.gn
+--- a/build/config/compiler/BUILD.gn
++++ b/build/config/compiler/BUILD.gn
+@@ -129,7 +129,7 @@ declare_args() {
+   # the space overhead is too great. We should use some mixture of profiles and
+   # optimization settings to better tune the size increase.
+   thin_lto_enable_optimizations =
+-      (is_chromeos_ash || is_android || is_win || is_linux) && is_official_build
++      (is_chromeos_ash || is_android || is_win || is_linux)
+ 
+   # Initialize all local variables with a pattern. This flag will fill
+   # uninitialized floating-point types (and 32-bit pointers) with 0xFF and the
--- a/patches/extra/ungoogled-chromium/fix-building-with-thinlto.patch
+++ b/patches/extra/ungoogled-chromium/fix-building-with-thinlto.patch
@@ -1,0 +1,225 @@
+From: Marek Behún <kabel@kernel.org>
+
+For some reason, building with ThinLTO (via the use_thin_lto=true flag)
+fails when ungoogled-chromium patches are applied.
+
+The failure is caused by undefined references to various crashpad
+and breakpad symbols when linking.
+
+It seems that building some of the crashpad/breakpad libraries without
+LTO fixes this issues.
+
+This patch separates the ThinLTO stuff from config("compiler") into
+config("thinlto"), adds this new config into default_compiler_configs,
+and then removes this config from the relevant crashpad and breakpad
+libraries.
+
+diff --git a/build/config/BUILDCONFIG.gn b/build/config/BUILDCONFIG.gn
+--- a/build/config/BUILDCONFIG.gn
++++ b/build/config/BUILDCONFIG.gn
+@@ -319,6 +319,7 @@ default_compiler_configs = [
+   "//build/config/compiler:afdo_optimize_size",
+   "//build/config/compiler:cet_shadow_stack",
+   "//build/config/compiler:compiler",
++  "//build/config/compiler:thinlto",
+   "//build/config/compiler:compiler_arm_fpu",
+   "//build/config/compiler:compiler_arm_thumb",
+   "//build/config/compiler:chromium_code",
+diff --git a/build/config/compiler/BUILD.gn b/build/config/compiler/BUILD.gn
+--- a/build/config/compiler/BUILD.gn
++++ b/build/config/compiler/BUILD.gn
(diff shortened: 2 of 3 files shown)


PF4Public:
Have you tried enabling official? I always build with `pgo` and `lto` and never had issues with it.

patches/extra/ungoogled-chromium/allow-thinlto-optimizations.patch
```diff
@@ -0,0 +1,17 @@
...
+official build.
+
+diff -Naurp a/build/config/compiler/BUILD.gn b/build/config/compiler/BUILD.gn
+--- a/build/config/compiler/BUILD.gn
++++ b/build/config/compiler/BUILD.gn
+@@ -129,7 +129,7 @@ declare_args() {
+   # the space overhead is too great. We should use some mixture of profiles and
+   # optimization settings to better tune the size increase.
+   thin_lto_enable_optimizations =
+-      (is_chromeos_ash || is_android || is_win || is_linux) && is_official_build
```
wchen342, line 13: This left out Debug builds with optimizations too. Chromium optimization has threee levels instead fo two, which are debug, non-debug non-official and official.

wchen342:
> the second separates the ThinLTO stuff from config("compiler") into
config("thinlto"), adds this new config into default_compiler_configs,
and then removes this config from the relevant crashpad and breakpad
libraries, so that these libraries are build without LTO

I feel like this is not the right fix:
- If none of the crashpad and breakpad related codes are actually run in UC, then it is more proper to remove the references to them instead of using a linking time hack
- If the crashpad and breakpad codes are actually run at some point in UC, then this shouldn't be done. The comments right below the section reads:
```
  # These flags enable optimizations/transformations that require whole-program visibility at link
  # time, so they need to be applied to all translation units, and we may end up
  # with miscompiles if only part of the program is compiled with LTO flags. For
  # that reason, we cannot allow targets to enable or disable these flags, for
  # example by disabling the optimize configuration.
```
.

macchrome:
I always build ungoogled chromium (Windows and Linux) with clang/llvm  thinLTO and PGO: never had a problem.

Is this proposed change really necessary? Familiarise yourself  with the relationship between LTO and PGO.

Relevant GN parameters;

clang_use_default_sample_profile = false
chrome_pgo_phase = 2
pgo_data_path="/home/acer/chromium/src/chrome-linux-master-1597082402-c5bd486642405d828e9e780a3fe4cf0c960a0b58.profdata"
use_thin_lto = true
thin_lto_enable_optimizations = true

####use_lld = true     default linker

PF4Public:
>Is this proposed change really necessary?

My thought exactly :)

elkablo:
> I always build with pgo and lto and never had issues with it.

Strange. Perhaps I am doing something wrong, I will look into this.

> Is this proposed change really necessary?

If other people don't have problems with linking, then something is probably wrong at my side. I will try to debug this. In the meantime please ignore this PR.

elkablo:
> Why are you willing to do it exactly this way?

@PF4Public, that's just a script that does the patching on the `www-client/chromium` package in Gentoo. But I will probably try to create PR into Gentoo adding a separate `www-client/ungoogled-chromium` package.

Anyway I still did not manage to make linking with LTO work without patching crashpad/breakpad to compile without LTO. I am now trying to compile the code with [one Gentoo patch](https://github.com/stha09/chromium-patches/blob/master/chromium-91-compiler.patch) reverted. But I fear it will fail even then. Afterwards I will try to compile with LLVM/clang shipped with Chromium, instead of Gentoo's, to see if maybe the problem is caused by that.

> If none of the crashpad and breakpad related codes are actually run in UC, then it is more proper to remove the references to them instead of using a linking time hack

@wchen342, yes, that would be ideal, but that change would be complicated. Though I have discovered that at least some parts of crashpad can be simply turned off by `use_crash_key_stubs=true` flag (although another fix is needed for this to work correctly). I will create a separate PR for this. Anyway removing all crashpad references would be a huge pain in the ass, more so for maintainance.

Regarding the comment below the section, I think that comment is wrong:
* [LLVM's LTO documentation](https://llvm.org/docs/LinkTimeOptimization.html) explicitly says that mixing of native object files with LTO object files is allowed.
* It doesn't make sense, since sometimes you don't have the option of LTO-only object files (assembly objects, for example).
* Somehow linking on my build fails when crashpad is compiled with LTO, but succeeds otherwise. (This may be a bug in the compiler - there were historically many such with LTO, at least for GCC)

I think it is possible that this comment was true with an older version of LLVM, because LTO was not yet fully functional.

I understand that this fix is not ideal, and will look into the issue. But currently it is the only maintainable solution (at least for my build).

PF4Public:
>But I will probably try to create PR into Gentoo adding a separate `www-client/ungoogled-chromium` package.

AFAIK there were attempts to do just that, but it was declined.

> > Why are you willing to do it exactly this way?
> 
> @PF4Public, that's just a script that does the patching on the `www-client/chromium` package in Gentoo. 

Why would you be doing that? Haven't you succeed building `ungoogled-chromium` from my overlay?

Quote from the [README.md](https://github.com/Eloston/ungoogled-chromium/blob/master/README.md):
[Gentoo: Available in ::pf4public overlay as ungoogled-chromium ebuild](https://github.com/Eloston/ungoogled-chromium/blob/master/README.md#downloads)

PF4Public:
@elkablo Do we close this PR?

elkablo:
> Do we close this PR?

@PF4Public I am going to try your ungoogled-chromium ebuild, with LTO. If it succeeds, we can.

elkablo:
@PF4Public ok it seems that problem is on my side, your ebuild builds with LTO. The problem probably is some config option. Sorry for spam.

PR #1651, opened 2021-09-21
Update to Chromium 94.0.4606.54

### LINKED ISSUE #1650: Updating to Chromium 94.0.4606.54
## Notes for the developer

## Notes for others

Ahrotahn:
This PR updates to Chromium 94.0.4606.54

Notable changes:

*  The `fieldtrial_testing_like_official_build` GN flag has been renamed to `disable_fieldtrial_testing_config` which should be less confusing!  <sup>[3034966](https://chromium-review.googlesource.com/c/chromium/src/+/3034966)</sup>
* The print preview fix is finally no longer needed! <sup>[3036801](https://chromium-review.googlesource.com/c/chromium/src/+/3036801)</sup>
* I added `chromium-upstream/missing-header-without-tflite.patch` for now which adds a header that is missing due to building with build_with_tflite_lib=false <sup>[3158501](https://chromium-review.googlesource.com/c/chromium/src/+/3158501)</sup>

There's also new error log output for debugging due to [[3024215]](https://chromium-review.googlesource.com/c/chromium/src/+/3024215).  

<br>

Fixes #1650 

[Blog post](https://chromereleases.googleblog.com/2021/09/stable-channel-update-for-desktop_21.html)
[Change log](https://chromium.googlesource.com/chromium/src/+log/93.0.4577.82..94.0.4606.54?pretty=fuller&n=10000)


patches/core/ungoogled-chromium/fix-building-without-enabling-reporting.patch
```diff
@@ -22,6 +22,20 @@
...
++++ b/content/public/common/content_switch_dependent_feature_overrides.cc
+@@ -47,9 +47,11 @@ GetSwitchDependentFeatureOverrides(const
+        std::cref(features::kDocumentPolicyNegotiation),
+        base::FeatureList::OVERRIDE_ENABLE_FEATURE},
+       {switches::kEnableExperimentalWebPlatformFeatures,
++#if BUILDFLAG(ENABLE_REPORTING)
+        std::cref(net::features::kDocumentReporting),
+        base::FeatureList::OVERRIDE_ENABLE_FEATURE},
+       {switches::kEnableExperimentalWebPlatformFeatures,
++#endif
```
Zoraver, line 35: Is there a reason that you didn't include the first line of the struct you are trying to remove in the `#if ... #endif` and instead included the first line of the next struct in the array? While this will do what you intend, it looks a bit strange to me.

In my mind, the main reason you would not want to do this is that if first line of the next struct in the array is changed in a future version of Chromium to be different than the first line of the struct you are trying to remove, refreshing this patch will be awkward. It would be more clear if no portion of any other struct is removed.

Ahrotahn: I'm not sure how I managed to get one line off, but you're right that it would cause problems as the entries get changed in the future.  I'll push an update in a bit.

networkException (approved):
Builds and runs fine on Arch :+1:

PF4Public:
Builds and runs on Gentoo

PF4Public (commented):
I'd suggest waiting a little bit for if Mac also builds fine.

As a side note, would it be beneficial in the long run to extract as much code as possible to static files (not patches)? Example would be `ungoogled_flag_entries.h`. This might help speeding updates. What you think?

networkException:
I can't reproduce this on my build

Ahrotahn:
@macchrome I'm unable to reproduce the issue as well.  Does that only happen when searching the settings, or does it also crash when searching history or chrome://flags?  

@PF4Public There might be some other patches that could benefit from doing that, but I'm not sure there'd be too many that could be made to work that way.  The flags were in a good position since those changes were all line inserts into the same file.

macchrome:
@Ahrotahn

The problem is solely related to when searching the settings.

ungoogled-chromium-94.0.4606.54-1_windows.7z (64-bit)
https://github.com/macchrome/winchrome/releases/download/v94.0.4606.54-r911515-Win64/ungoogled-chromium-94.0.4606.54-1_Win64.7z

Portable GNU/Linux cross-compile build for Windows x64 using the Chromium source tarball and the respective associated build tools, and the latest Windows SDK.

core/inox-patchset/0001-fix-building-without-safebrowsing.patch
core/debian/disable/unrar.patch
core/iridium-browser/safe_browsing-disable-incident-reporting.patch
core/iridium-browser/safe_browsing-disable-reporting-of-safebrowsing-over.patch
core/ungoogled-chromium/fix-building-without-safebrowsing.patch
core/ungoogled-chromium/remove-unused-preferences-fields.patch
extra/ungoogled-chromium/remove-uneeded-ui.patch..

have not been applied and the respective (manually) amended Windows have.  Safe browsing has been disabled by polices:  features.cc and safe_browsing_prefs.cc 

Yes, I know this is not ungoogled as such; but there must be an alternative way of removing\disabling  safe browsing\chrome cleaner, and windows users should not be deprived of a browser based on ungoogled principles.

kramred:
> I'd suggest waiting a little bit for if Mac also builds fine.

macOS [did build successfully](https://github.com/kramred/ungoogled-chromium-macos/releases) and runs well for me (x86_64) - I'll open an OS specific PR in that repo when this one is merged. Changes are already in the `release/v94*` branches of my fork

Ahrotahn:
@macchrome I did a little debugging on your build and found that the search completes and the results show just before the crash. 
 It seems to be profile related as it will crash for me when going to the profile settings as well as when trying to add a new profile from the avatar toolbar button.  You might want to try testing a build without the policy changes to see something has changed.  Alternatively, if you have the time, make a build using `symbol_level=2` without stripping and I'll be able to debug further.  

The safebrowsing patches do make things difficult and that's something that we've been trying to make better.  I've experimented with a few alternative methods in the past, but they end up being at least as much work as the current patches.  I haven't done much with that in a while though, I'll take some time soon to play around with a couple other ideas I've since then.

Nifury:
@macchrome Can't reproduce either. 
It's worth noting that `DCHECK` is enabled by default on Windows for some reason. I added `dcheck_always_on=false` and now it runs without any issue. 

screenshot
.

Zoraver (approved):
Looks good to me. Great work @Ahrotahn!

PF4Public (approved):
So are we good to merge now?

PF4Public:
Thanks @Ahrotahn!

It is 3 days now, and 2 days since last commit.
Windows confirmed building, Mac confirmed building, Linux confirmed building. I'm pushing the ~red~ green button.

PF4Public:
I'm worried for Android now 🤦‍♂️

networkException:
Don't worry, I'd have merged it as well if I were to have a little more courage. If android is facing issues those can surely be fixed by either an android specific patch or an incremental release.

wchen342:
> I'm worried for Android now man_facepalming

Well this update is so fast ~~that I didn't see it coming at all~~.
My server is building last version of cr93 right now so I cannot test the new version immediately. Also Bromite is not updating recently which also delayed my process. Anyway I will report back if there are any problems in the coming weekend.

macchrome:
@Ahrotahn

As requested:

https://github.com/macchrome/winchrome/releases/download/v94.0.4606.54-r911515-Win64/chrome-94.0.4606.54-developer-build.7z

Ahrotahn:
It seems to be caused by a string resource that isn't able to be loaded.  Removing the check [here](https://source.chromium.org/chromium/chromium/src/+/refs/tags/94.0.4606.54:ui/base/resource/resource_bundle.cc;l=1063) should let the build function, but wont solve the root issue.  I'll spend some more time this weekend to see if I can find any other info on what it's looking for specifically.

macchrome:
@Ahrotahn 
I appreciate the debug build hint, and agree with your conclusion.

Issue #1656, opened 2021-09-21
#global-media-controls flag gone.


egg4096:
Despite being reffered to by other flags on the `chrome://flags` page, the `#global-media-controls` flag, which can be used to disable this button (shows up in the top right next to the "Extensions" button while playing YouTube videos & other media):
![Screenshot from 2021-09-21 23-21-43](https://user-images.githubusercontent.com/72076491/134242022-cadfddd6-e78f-4e54-8651-7dcbc47b467b.png)
...is either hidden or removed from `chrome://flags`.
###### _I hope I'm not being an annoyance by missing something obvious._

networkException:
As far as I can tell the flag has not been present since 85: <https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/flag-metadata.json;drc=0ff877a2651bf0e328ef53620a1f01dcbb09b8f9;l=3331>

Ahrotahn:
It can still be disabled by adding `--disable-features=GlobalMediaControls` to your launch switches.  

The code for the flag still exists, so another option if you're compiling your own build would be to edit the file @networkException pointed out and set the expiry to a later version or remove the entry.

PF4Public:
I hope your question has been answered, otherwise please let us know.
Closing this issue for now.

PR #1670, opened 2021-09-26
Add flag to show a prompt when closing mutitab

Nifury:
This PR adds a flag that allows users to confirm whether they want to close the last window when multiple tabs are open in it. 

I sometimes close the browser accidentally and cannot restore the tabs because I have `ClearDataOnExit` enabled. It is exhausting trying to recall what I was browsing. A warning dialog could save the day. :smiley:

Fixes #1665

PR #1671, opened 2021-09-27
Add patch to keep expired flags

Ahrotahn:
This PR adds a patch that allows expired flags to remain in `chrome://flags` with an added warning that the flag may be removed in a future version.  

This is something that had been on my mind after commenting on #1656.  The Arch repos include a patch that unexpires the accelerated video decode flag and I've seen other issues in the past about flags disappearing.  The code and the flag itself can sometimes stick around for a long time after being expired.

### DIFF
--- a/patches/extra/ungoogled-chromium/keep-expired-flags.patch
+++ b/patches/extra/ungoogled-chromium/keep-expired-flags.patch
@@ -1,0 +1,38 @@
+--- a/components/flags_ui/flags_state.cc
++++ b/components/flags_ui/flags_state.cc
+@@ -22,6 +22,7 @@
+ #include "base/values.h"
+ #include "build/build_config.h"
+ #include "build/chromeos_buildflags.h"
++#include "chrome/browser/unexpire_flags.h"
+ #include "components/flags_ui/feature_entry.h"
+ #include "components/flags_ui/flags_storage.h"
+ #include "components/flags_ui/flags_ui_switches.h"
+@@ -577,14 +578,18 @@ void FlagsState::GetFlagFeatureEntries(
+   int current_platform = GetCurrentPlatform();
+ 
+   for (const FeatureEntry& entry : feature_entries_) {
++    std::string desc = entry.visible_description;
+     if (skip_feature_entry.Run(entry))
++      if (flags::IsFlagExpired(flags_storage, entry.internal_name))
++        desc.insert(0, "!!! NOTE: THIS FLAG IS EXPIRED AND MAY STOP FUNCTIONING OR BE REMOVED SOON !!! ");
++      else
+       continue;
+ 
+     base::Value data(base::Value::Type::DICTIONARY);
+     data.SetStringKey("internal_name", entry.internal_name);
+     data.SetStringKey("name", base::StringPiece(entry.visible_name));
+     data.SetStringKey("description",
+-                      base::StringPiece(entry.visible_description));
++                      base::StringPiece(desc));
+ 
+     base::Value supported_platforms(base::Value::Type::LIST);
+     AddOsStrings(entry.supported_platforms, &supported_platforms);
+@@ -901,6 +906,7 @@ bool FlagsState::IsSupportedFeature(cons
+     if (!entry.InternalNameMatches(name))
+       continue;
+     if (delegate_ && delegate_->ShouldExcludeFlag(storage, entry))
++      if (!flags::IsFlagExpired(storage, entry.internal_name))
+       continue;
+     return true;
+   }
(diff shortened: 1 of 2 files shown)


patches/extra/ungoogled-chromium/keep-expired-flags.patch
```diff
@@ -0,0 +1,29 @@
+--- a/components/flags_ui/flags_state.cc
++++ b/components/flags_ui/flags_state.cc
+@@ -577,14 +577,15 @@ void FlagsState::GetFlagFeatureEntries(
+   int current_platform = GetCurrentPlatform();
+ 
+   for (const FeatureEntry& entry : feature_entries_) {
++    std::string desc = entry.visible_description;
+     if (skip_feature_entry.Run(entry))
+-      continue;
++      desc.insert(0, "NOTE: THIS FLAG IS EXPIRED AND MAY BE REMOVED IN A FUTURE CHROMIUM VERSION! ");
```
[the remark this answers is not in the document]
Ahrotahn: It gets crowded in there with all those flags!  

It's actually just a means to catch attention since there's no easy way to have formatting.  The description is displayed in an inline element so linebreaks can't even be used.  I'm open to alternative wording, or even removing the note altogether.  There's no forewarning to users for when a flags will be expired, so it wouldn't change much to have no forewarning for when the flag is actually removed.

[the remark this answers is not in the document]
PF4Public: Indeed! My only concern with this patch would be the fact that those flags are like zombies ready to leave.

How about this then: "NOTE: THIS FLAG IS EXPIRED AND MAY BE REMOVED ANY TIME SOON! "

patches/extra/ungoogled-chromium/keep-expired-flags.patch
```diff
@@ -0,0 +1,29 @@
+--- a/components/flags_ui/flags_state.cc
++++ b/components/flags_ui/flags_state.cc
+@@ -577,14 +577,15 @@ void FlagsState::GetFlagFeatureEntries(
+   int current_platform = GetCurrentPlatform();
+ 
+   for (const FeatureEntry& entry : feature_entries_) {
++    std::string desc = entry.visible_description;
+     if (skip_feature_entry.Run(entry))
```
Zoraver, line 8: Based on my reading of `chrome/browser/about_flags.cc`, specifically `ShouldSkipConditionalFeatureEntry()`, I believe that `skip_feature_entry.Run(entry)` may return true when the entry has not expired, but rather is intended to be skipped for another reason. For example, upstream intends that the `enable-unsafe-webgpu` flag is only available on the dev and canary release channels; this, along with other conditions, is enforced in `ShouldSkipConditionalFeatureEntry()`.

Ahrotahn: Good point, I'll update the patch to check explicitly for expired flags along with the verbiage changes.  Interestingly the `enable-unsafe-gpu` flag is available without these changes.  I was worried that there may be an issue with some other ungoogled-chromium patch causing an issue, but I checked with a build of vanilla chromium and the flag is available there too!

Ahrotahn:
I've updated the note verbiage and checks for expired flags.  `ShouldExcludeFlag` only returns the result for `IsFlagExpired` currently, but I opted to add a check for `IsFlagExpired` instead of removing `ShouldExcludeFlag` so that this patch still will function correctly if other checks are added to `ShouldExcludeFlag` in the future.

patches/extra/ungoogled-chromium/keep-expired-flags.patch
Zoraver: Our builds likely count as `Channel::UNKNOWN`, which would explain why `enable-unsafe-gpu` is available.

Zoraver (approved):
Looks good to me! You may want to wait to merge this until after #1674.

PF4Public (approved):
Ready to merge?

PF4Public:
If no one objects, I'll merge #1671 and #1670 just after we release new version in #1680

Ahrotahn:
The text cannot be formatted with the existing code.  It should be possible to change that but would make this patch a lot more complex.  
Another option would be to include special unicode characters to create the formatting, but that runs the risk of showing up as squares for users without the font coverage.  

Currently it's using an all-caps note added to the beginning of the description for each of the kept flags.  Hopefully that is enough to catch attention.  Maybe something like this would work?:  
`NOTE: THIS FLAG IS EXPIRED AND MAY STOP FUNCTIONING OR BE REMOVED SOON! `

wchen342:
In that case yeah, we can only use ASCII texts. I personally feel all caps is a bad design though, maybe wrapping the warning with `!!!---!!!` or `###---###` is better? (I'm no UI designer so it's just my feeling).

wchen342:
Seems good to me, this looks nice.

PF4Public:
> which looks nicer

… and more prominent!

PF4Public:
Thanks @Ahrotahn!

Issue #1513, opened 2021-05-06
stop loading anything under chrome-search://local-ntp


jacob755:
**Is your feature request related to a problem? Please describe.**

the problem is when i start browser or click + button to create new tab and pasting any url 
if i wont wait for 2 second or more title bar show chrome-search://local-ntp/local-ntp.html
and pasted url wont open. Seems many have same problem when googling
i disabled all ntp flag added --disable-top-sites but still happen
i try extension that replace new tab page but still slowdown happen

**Describe the solution you'd like**

the thing i want is flag to disable any html js css that load in background in a chrome-search://local-ntp
so pure blank white page ready o paste or browse no top site or whatever was in a new tab page. And this is a Optional so if anyone want that behavior
still have it

according to here seems html js css loding 
https://chromium.googlesource.com/experimental/chromium/src/+/d07ec0748f7e29857bf61a19e1e452aa89daa7c1/chrome/browser/resources/local_ntp

**Describe alternatives you've considered**
can be disabled by patch  for everyone?!
**Additional context**

networkException:
I have personally never experienced that issue, does it occur when using a different user data directory (set using `--user-data-dir`, see https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md#overriding-the-user-data-directory)

jacob755:
there is a no difference

guest271314:
Even though on first run the option is provided to customize the page, hiding "tiles" and "cards" does not work Issue 1192882: After disabling tiles and cards they still appear on chrome://newtab https://bugs.chromium.org/p/chromium/issues/detail?id=1192882 is marked as 'WontFix'

> In general, NTPs are considered an integral part of Chrome's UI. Turning off JS will break the experience and so we don't allow this. You can also not set an arbitrary page as your NTP. This is currently by design. Closing this bug as WAI.

The workaround I use is a Chrome extension, see 
Simplest extension (unpacked, non-CWS) to disable JavaScript on new tab page https://groups.google.com/a/chromium.org/g/chromium-extensions/c/aH4kwzUqGXk/m/V2b5Af_MBAAJ.

```
{
  "name": "tab",
  "version": "1.0",
  "manifest_version": 3,
  "chrome_url_overrides" : {
    "newtab": "./blank.html"
  }
}
```

blank.html

```<!doctype html><html><body></body></html>```

jacob755:
[...]

PR #1686, opened 2021-10-11
Add flag for setting a custom new tab page

Ahrotahn:
This PR adds a flag allowing a custom new tab page to be set.  This is handy for anyone that wants a blank new tab page or wants to use a custom [startpage](https://www.reddit.com/r/startpages) for new tabs.  

The page can be internal (e.g. `about:blank`), external (e.g. `example.com`), or local (e.g. `file:///tmp/startpage.html`).  
This applies in incognito as well except when set to a `chrome://` page since most of them require access to the profile.  In those scenarios the normal incognito tab will be shown.


jacob755:
@Ahrotahn 
is this for #1513 ? if so thank you so much and much appreciated and its very good flag then 
also question can make patch to remove chrome://new-tab-page-third-party/ entirely from browser?

Ahrotahn:
If changing the new tab page indirectly solves your issue, then this would make that easier to deal with!

PR #1693, opened 2021-10-15
Add build script for easier building/debugging.

wchen342:
Because of how hard it is to fetch and build chromium and there seems to be a lot of people who want to build `ungoogled-chromium` themselves but weren't able to, I strongly feel it is necessary to write a script that runs high-level commands and strips away all the complicated details in building chromium so that new users can start building chromium without getting lost in complicated chromium-specific building systems and terms.

Since the goal is as easy to use as possible, I limit the amount of configurable options on purpose and make things as standard as possible. The script is currently only tested for `Linux` builds but I will add `Android` and `Windows` soon. For Linux, it builds `portable` with the `sysroot` coming with source. The choice is made so that the differences between different distributions are minimized, as well as based on the fact that the benefits of `ungoogled-chromium` mainly comes from its flags, patches and domain substitutions, so using `sysroot` is an acceptable compromise IMO.

There are some extra functionalities underway, like a filter for patches and pruning lists, as well as downloading source tarball directly instead of checking out chromium source which is incredibly slow on a not very fast internet connections.

If you have any suggestions on how it can be improved or what additional functionality it should have, please let me know. If you think this is a good idea to have this script, I can ask @Eloston to move it under `ungoogled-software` so it is easier to edit for the rest.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "utils/ungoogled-chromium-building-scripts"]
+	path = utils/ungoogled-chromium-building-scripts
+	url = https://github.com/wchen342/ungoogled-chromium-building-scripts.git
--- a/README.md
+++ b/README.md
@@ -146,6 +146,22 @@ List of mirrors:
 
 [See docs/building.md](docs/building.md)
 
+A quick way to build Linux binaries, run the following:
+
+```sh
+cd utils/ungoogled-chromium-building-scripts
+./build.py init --shallow
+./build.py sync --shallow --os linux --install-build-deps
+./build.py prepare --os linux
+./build.py build -a x64 --os linux
+```
+
+For debug builds, run
+```sh
+./build.py build -a x64 --os linux --debug
+```
+instead.
+
 ## Design Documentation
 
 [See docs/design.md](docs/design.md)


networkException:
LGTM :+1: 

I was a bit concerned about possible duplication with `ungoogled-software/ungoogled-chromium-*` repositories but I think with the extra functionalities you already mentioned this tool will basically only be used to build custom binaries for specific needs or for general development.

This might actually be a perfect opportunity to integrate https://github.com/ungoogled-software/contrib, not only for additional patches but also as a place to store custom domain substitution lists and so on.

Moving the repo to `ungoogled-software` would be very good as well.

mid-kid:
Dropping my 2ct, since I've been using the `-portablelinux` scripts for ages.
There's really nothing debian-specific about those scripts, it merely uses a debian image to generate an appimage if you so choose, but the base scripts are very minimal (really, it's just a copy-paste of the docs/building.md) and run anywhere chromium builds. It's just rather slow with getting updates at times, and it forces building using clang+lld.

PF4Public:
@wchen342 You did a tremendous work!

Why did you do it as a submodule? Your script is a rather independent entity. It deserves its own repository under `ungoogled-software` organization, or inclusion into `utils` directory as a part of `ungoogled-chromium` itself. Given the detachment of your script from ungoogled patches [as a whole], I'm leaning more to a separate repository. In any case, your quick-build instructions (using your script) should be present here without any doubt, but don't forget to document your script in its own readme.

Your script might serve as a common ground for debugging issues, when hunting ungoogled-chromium bugs! Or otherwise proving them unrelated to ungoogled-chromium patches!

Cannot agree more that having such script for Windows could improve situation there immensely.

Nifury:
I just went over the code, and it seems there is not much code I can reuse. I would have to create big `if` blocks as building on Windows is quite different from building on Linux. Does that sound okay to you, or would it be better if I just refactor the existing Windows build script to provide similar functionalities?

tangalbert919:
This is a good idea. It's just not clear to me if the objective of this build script is to negate the need to clone repos for specific systems (e.g. `ungoogled-chromium-windows` for Windows) and simply do all the building and debugging from this repository alone.

wchen342:
The main goal is to simplify the building process so that we don't need to manually do all the `git clone`, `gclient sync` and other stuff like setting up environment variables, etc. It currently hard-codes repository links to platform-specific ones but an option can easily be added.
The builds are always portable because it doesn't use anything system-specific (except the VS toolchain on Windows). This can be a complement to those platform-specific repositories where each system has their own way of packaging.

> do all the building and debugging from this repository 

Do you mean like merging all the platform repositories back into this main repo? I think that is a different problem. There has always been a trade-off between large mono repos and smaller, separate repos. This was discussed before my time here and I think the conclusion was to separate them so it became what we see today.

PR #1718, opened 2021-10-30
Fix close confirmation issue

Nifury:
This PR fixes https://github.com/Eloston/ungoogled-chromium/pull/1670#issuecomment-954682653 where the confirmation dialog is prompted even when the `Close Confirmation` is set to `multiple` and there is only one window open.


PF4Public (approved):
I've just rebuilt it with this change and it works as expected. Thanks!

PF4Public:
Thanks @Nifury

PR #1728, opened 2021-11-03
Add patch for DoH changes

### LINKED ISSUE #1515: No option to enable encrypted DNS & no option to disable icons on the new tab
**Describe the bug**
No option to enable encrypted DNS on chrome://settings/security.
No option to disable icons on the new tab.

**Expected behavior**
The ability to enable dns and disable icons on the new tab

**Screenshots**
Screenshot from 2021-05-09 20-02-14

Screenshot from 2021-05-09 20-03-14

 - OS/Platform and version: Fedora Linux with Gnome 40
 - ungoogled-chromium version: Version 90.0.4430.93 (Official Build, ungoogled-chromium) (64-bit) (flatpak)

Ahrotahn:
This PR adds a new core patch that makes a few changes to DNS over HTTPS (DoH):
* Removes Google DoH servers from the list of providers.  This was previously missed by the domain substitution since it doesn't match on IP addresses.  If you had manually set Google's DoH server as your provider before this patch then the entry will become a custom entry after this patch is applied.  
* Allows enabling DoH on linux.  DoH works fine on linux and many other chormium-based browsers enable this for the platform.  
* Disables DoH by default.  It's best to respect the system's DNS settings by default while allowing users to enable DoH if they need it.  
* Disables DoH auto upgrade.  The browser shouldn't be changing servers without the user's consent.  Before the Google servers were removed this had the potential to upgrade to their servers without the user's knowledge.  

Fixes #1515

### DIFF
--- a/patches/core/ungoogled-chromium/doh-changes.patch
+++ b/patches/core/ungoogled-chromium/doh-changes.patch
@@ -1,0 +1,70 @@
+--- a/chrome/common/chrome_features.cc
++++ b/chrome/common/chrome_features.cc
+@@ -309,12 +309,7 @@ const base::Feature kDesktopPWAsWebBundl
+ // Enable DNS over HTTPS (DoH).
+ const base::Feature kDnsOverHttps {
+   "DnsOverHttps",
+-#if defined(OS_WIN) || defined(OS_CHROMEOS) || defined(OS_MAC) || \
+-    defined(OS_ANDROID)
+-      base::FEATURE_ENABLED_BY_DEFAULT
+-#else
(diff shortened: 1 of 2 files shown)


Zoraver (approved):
Looks good to me!

wchen342 (approved):
LGTM:+1:

Ahrotahn:
They enabled DoH for linux for v96 in [[3198710]](https://chromium-review.googlesource.com/c/chromium/src/+/3198710), so I guess that part of the change will be short-lived!

PR #1748, opened 2021-11-26
Update stale issues and PRs workflow

PF4Public:
* Changed the default 60/7 days of inactivity to 14/14.
* Plus messages to convey the stale'ing and closing of the issues.

I'd like to hear from you all regarding the following:
1. I believe that default 60 days is way too much, but not certain if 14 is a reasonable value either. What you think?
2. How do you like the wording of the messages? Are they concise and make the point clear?

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,3 +16,7 @@ jobs:
             with:
                 exempt-all-milestones: true
                 any-of-labels: 'need info'
+                days-before-stale: 30
+                days-before-close: 7
+                stale-issue-message: "This issue has been automatically marked as stale as there has been no recent activity in response to our request for more information. Please respond so that we can proceed with this issue."
+                close-issue-message: "This issue has been automatically closed as sufficient information hasn't been provided on the issue for further actions to be taken. Feel free to add more information."


Ahrotahn:
14/14 seems reasonable.  I guess that could always be adjusted in the future if there's issues with that timeframe.  

The close message might be more clear if rearranged slightly:  "This issue has been automatically closed as sufficient information hasn't been provided on the issue for further actions to be taken."

tomasz1986:
I would just like to add that 2 weeks is nothing when people are busy with their lives (e.g. work, family, etc.). I've experienced this myself in different projects, when sometimes it may take months for someone to come back with the requested information.

Also, it's super annoying having to bump up the issue every few weeks just because it's been marked as "stale" by some bot, even though the issue itself hasn't been solved, and no developer has even had time to start working on it yet.

PF4Public:
> I would just like to add that 2 weeks is nothing

After stale'ing there are another 2 weeks before closing. This is actually why I want to gather more feedback on the timings proposed. We may make it 30/7 for example.

> Also, it's super annoying having to bump up the issue every few weeks just because it's been marked as "stale" by some bot, even though the issue itself hasn't been solved, and no developer has even had time to start working on it yet.

The intention here is to stale _only those_ issues, which are marked with `need info` label, which should signify, that not enough information is available to proceed. If enough information is gathered, the label should be removed and a new one attached, like `enhancement` or `bug` etc. You may see, that this bot acts _only_ on `need info` issues and _ignores_ the rest.

> The close message might be more clear if rearranged slightly: "This issue has been automatically closed as sufficient information hasn't been provided on the issue for further actions to be taken."

Looks even better!

wchen342:
I think 30/7 is a reasonable value. 14 is too short IMO.

PF4Public:
I've squashed everything and put 30/7.

wchen342 (approved):
Maybe change `Feel free to add more information.` to `Feel free to add more information so that we can re-open the issue.`. Otherwise LGTM.

PF4Public:
> so that we can re-open the issue

I would be very-very-very cautious about promising anything anyone, that's why I intentionally omitted this phrase ;) :D

wchen342:
The it's better left out the entire sentence if there is no intention to do anything, otherwise it is confusing I think.

PF4Public:
> there is no intention to do anything

Quite the opposite! If the information is sufficient for re-opening - why not! In case it is a mere "bump" thing — sorry, no way.

wchen342:
> > there is no intention to do anything
> 
> Quite the opposite! If the information is sufficient for re-opening - why not! In case it is a mere "bump" thing — sorry, no way.

Without explanation I think people cannot read into all these in that one sentence, but let's keep it the way you put it for now.

networkException (approved):

PR #1749, opened 2021-11-27
Automatically creating an "Updating to Chromium x.x.x.x" issue

PF4Public:
I'll leave it running in my fork for a while. Apart from that, I'd like to get some feedback on this before squashing and marking as ready for review.

TODO list before squashing:
- [x] Change variable in `if: github.repository == `

### DIFF
--- a/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
+++ b/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
@@ -1,24 +1,16 @@
 ---
 name: Create an "Updating to Chromium x.x.x.x"
 about: For letting the community track progress to a new stable Chromium
-title: Updating to Chromium [VERSION_HERE]
-labels: enhancement, help wanted
+title: Updating to Chromium {{ env.VERSION }}
+labels: update
 assignees: ''
 
 ---
 
-### Please respond if you would like to update ungoogled-chromium to the new stable Chromium version.
+Chromium stable channel for Linux has been updated to a newer version: {{ env.VERSION }}.
 
-## Notes for the developer
+If you are willing to work on updating the patches and lists, please leave a comment in this issue in order to facilitate better coordination and avoid wasted/duplicated efforts.
 
-Once you claim it, it is advisable that you create a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests) for increased visibility:
+If you'd like to increase visibility of your progress or get early feedback/advice, consider creating a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests). Finally, make sure to reference this issue in your PR. Please make sure to read [/docs/developing.md](https://github.com/Eloston/ungoogled-chromium/blob/master/docs/developing.md#updating-patches) for guidance.
 
-![GitHub Interface for creating Draft Pull Requests](https://help.github.com/assets/images/help/pull_requests/pullrequest-send.png)
-
-Finally, make sure to reference this issue in your PR.
-
-Feel free to ask questions and discuss issues here along the way. Thanks!
-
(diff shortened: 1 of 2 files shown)


networkException:
Maybe there's an action that creates an issue if it doesn't exist? Instead of having two seperate steps for that, otherwise lgtm

PF4Public:
> Maybe there's an action that creates an issue if it doesn't exist?

At the time of #1219 `JasonEtco/create-an-issue` didn't have such functionality, but it looks like it has `update_existing: true`  and `search_existing: all` nowadays, which might be good for us. I'll look into it.

PF4Public:
dblock just published fixed version of `JasonEtco/create-an-issue`, so we are all set now and there's no need for forking. I'll leave it running in my fork for a while.

networkException:
Not entirely sure but wouldn't it be cleaner to just have a seperate tag for update prs / issues?

PF4Public:
I cannot think of anything better than just 'update' for a tag name. @networkException What would _you_ call it?

networkException:
I mean thats descriptive so why not

PF4Public:
It is indeed cleaner. And if no one claims it, we can put `help wanted` manually.
image

networkException:
This might be a bit bikeshed-y but reading "Once you claim it, it is advisable that you create a Draft Pull Request for increased visibility" does not really fit with how we are handling update prs currently.
Sure drafts very much have their place for larger updates that require thorough review or for updating to non stable (but soon to be stable) branches, but there is no need to increase visibility if there are only a hand full prs open.

How about we replace that portion with a link to https://github.com/Eloston/ungoogled-chromium/blob/master/docs/developing.md#updating-patches perhaps?

PF4Public:
Yes! @networkException you read my mind! I just wanted to rewrite the template a little bit, but ~didn't have the time for this~ was very lazy recently and was postponing this ever since :(

networkException:
Oh yea thats very relatable, winter holidays kinda killed my productivity

PF4Public:
At last I did some writing. Please see [the old template](https://github.com/PF4Public/ungoogled-chromium/issues/14) and [the new one](https://github.com/PF4Public/ungoogled-chromium/issues/16). 

Is the new one better or worse? I'm especially worried about the wording in it as it will be read many-many times. It should be flawless.

networkException:
> This might be a bit bikeshed-y but reading "Once you claim it, it is advisable that you create a Draft Pull Request for increased visibility" does not really fit with how we are handling update prs currently. Sure drafts very much have their place for larger updates that require thorough review or for updating to non stable (but soon to be stable) branches, but there is no need to increase visibility if there are only a hand full prs open.

I think that point still holds, otherwise the new one looks good to me

PF4Public:
> I think that point still holds

Oh, sorry @networkException, I forgot to reply to this point of yours. I think, that by "increased visibility" Eloston meant "claim this work" or "let others see and comment from the start". At least that's how I read this and in this context it looks fine to me. In the context of "make it obvious that you are about to submit a PR"… kind of… If you think it is dubious, we can and should rephrase that! In the end we can ping Eloston himself for his input on this one :)

Eloston:
Overall I think the idea is good and the implementation LGTM.

> I think, that by "increased visibility" Eloston meant "claim this work" or "let others see and comment from the start"

It's been too long, so I don't know why I wrote this... if I had to guess, it's probably both of these reasons. From my current perspective, I think it is better to reword this because it isn't really necessary to use a draft PR to "claim this work". I find Draft PRs are best for those who prefer to work on large changes in steps and publish the progress online for others to track. The most useful scenario is if the PR author needs some high-level feedback before finalizing some changes. However, there are developers that prefer working offline until the change is completely done, so a Draft PR doesn't fit their workflow very well.

Perhaps we can rewrite this statement so we let developers choose a workflow. For example:

> If you'd like to increase visibility of your progress or get early feedback/advice, consider creating a Draft PR.

PF4Public:
Marking as ready for review

networkException (approved):

Eloston (approved):

PR #1750, opened 2021-11-28
Update mDNS patch for local IP obfuscation

### LINKED ISSUE #1725: WebRTC still leaks the local IP of users
**Describe the bug**
Apparently the `#enable-webrtc-hide-local-ips-with-mdns` flag has no effect on this version, as the browser would ignore it (whether default or enabled) and still leak the local IP address even though the Chromium fix [already rolled out
](https://bugs.chromium.org/p/chromium/issues/detail?id=878465), now I don't know if I'm doing something wrong but https://browserleaks.com/webrtc still shows my local IP when it doesn't on Chrome v95.0.4638.69 or Edge v95.0.1020.40.

**To Reproduce**
1. Go to https://browserleaks.com/webrtc or https://tenta.com/test/
2. Check local IP Address field

**Expected behavior**
Should be N/A on https://browserleaks.com/webrtc or XXXXXXX.local on https://tenta.com/test/

**Screenshots**

image
image

 - OS/Platform and version: Windows 10 21H1 x64 19043
 - ungoogled-chromium version: 95.0.4638.69

**Additional context**
While there may be ways to disable WebRTC and prevent the issue altogether, it is not the default and since Chromium has already implemented the fix, other extensions like uBlock Origin also removed their workarounds to prevent leaks (as can be seen in uBlockOrigin/uBlock-issues#1723). The combination of the user's real IP address and the local one can pose serious and dangerous security issues if exploited properly, and right now Chromium's official fix doesn't seem to be effective on this browser, nor does the flag work.

Ahrotahn:
This PR adds a change to the existing mDNS patch which allows local IP obfuscation when building with `enable_mds=false`.  

`kWebRtcHideLocalIpsWithMdns` is enabled by default in Chromium and replaces the local IP returned via WebRTC with an address that is resolvable on the local network using mDNS.  This can be disabled with `chrome://flags/#enable-webrtc-hide-local-ips-with-mdns`.  

Interestingly, the sanitation of the local IP address doesn't rely on mDNS at all!  The obfuscation can be enabled based on the flag's state rather than the existence of the mDNS responder which would be created from that flag.  The address is not replaced with the mDNS-accessible one since it's not enabled.  

<sub>Normally, with mDNS enabled:</sub>
Screenshot_20211128_095916

<sub>Without mDNS:</sub>
Screenshot_20211128_095520

This allows the browser to function similarly to vanilla Chromium and utilizes the existing flag as well.  It's a bit of misnomer since it isn't using mDNS, but most end users searching to change this behavior would end up landing on the `WebRtcHideLocalIpsWithMdns` feature flag.  

Fixes #1725

---
Notes:

I used the sites in the linked issue to test:  https://tenta.com/test/  and  https://browserleaks.com/webrtc

uBlock Origin had an option to hide the local IP returned with WebRTC and has since removed that option since Chromium added the obfuscation.  It seems that only the option itself was removed, so if you had that enabled previously then it should still be functioning and the local IP would be shown on tenta as `UNKNOWN`!

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -65,3 +65,17 @@ These are also available on the `chrome://flags` page.
   -- | --
   `ClearDataOnExit` | Clears all browsing data on exit.
   `DisableQRGenerator` | Disables the QR generator for sharing page links.
+
+
+<br>
+
+---
+
+<br>
+
+
+# Existing Chromium flags
+
+Behavior for existing Chromium flags is normally not altered, but due to build flags or other core modifications used by ungoogled-chromium some exceptions may apply:
+
+* `WebRtcHideLocalIpsWithMdns`:  This feature flag does not use mDNS to hide the local IP since ungoogled-chromium is built with `enable_mds=false`.  A randomized address will be provided instead.  
--- a/patches/extra/ungoogled-chromium/fix-building-without-mdns-and-service-discovery.patch
+++ b/patches/extra/ungoogled-chromium/fix-building-without-mdns-and-service-discovery.patch
@@ -40,3 +40,22 @@
  }
  
  }  // namespace media_router
+--- a/third_party/webrtc/p2p/client/basic_port_allocator.cc
++++ b/third_party/webrtc/p2p/client/basic_port_allocator.cc
+@@ -33,6 +33,7 @@
+ #include "rtc_base/trace_event.h"
+ #include "system_wrappers/include/field_trial.h"
+ #include "system_wrappers/include/metrics.h"
++#include "third_party/blink/public/common/features.h"
+ 
+ using rtc::CreateRandomId;
+ 
+@@ -565,7 +566,7 @@ void BasicPortAllocatorSession::GetCandi
+ }
+ 
+ bool BasicPortAllocator::MdnsObfuscationEnabled() const {
+-  return network_manager()->GetMdnsResponder() != nullptr;
++  return base::FeatureList::IsEnabled(blink::features::kWebRtcHideLocalIpsWithMdns);
+ }
+ 
+ bool BasicPortAllocatorSession::CandidatesAllocationDone() const {
--- a/patches/extra/ungoogled-chromium/fix-building-without-mdns-and-service-discovery.patch
+++ b/patches/extra/ungoogled-chromium/fix-building-without-mdns-and-service-discovery.patch
@@ -40,6 +40,25 @@
  }
  
  }  // namespace media_router
+--- a/third_party/webrtc/api/candidate.cc
++++ b/third_party/webrtc/api/candidate.cc
+@@ -10,6 +10,7 @@
+ 
+ #include "api/candidate.h"
+ 
++#include "base/guid.h"
+ #include "rtc_base/helpers.h"
+ #include "rtc_base/ip_address.h"
+ #include "rtc_base/logging.h"
+@@ -135,7 +136,7 @@ Candidate Candidate::ToSanitizedCopy(boo
+     rtc::IPAddress ip;
+     if (address().hostname().empty()) {
+       // IP needs to be redacted, but no hostname available.
+-      rtc::SocketAddress redacted_addr("redacted-ip.invalid", address().port());
++      rtc::SocketAddress redacted_addr(base::GenerateGUID() + ".local", address().port());
+       copy.set_address(redacted_addr);
+     } else if (IPFromString(address().hostname(), &ip)) {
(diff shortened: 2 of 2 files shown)


PF4Public (dismissed):
Looks pretty benigh

Zoraver (dismissed):
This looks reasonable to me. That said, something to consider is whether there are any other browsers that ever send `redacted-ip.invalid`; if there aren't, using this will identify you to websites as a user of ungoogled-chromium with `enable-webrtc-hide-local-ips-with-mdns` set to a non-default value. This is may still be better from a fingerprinting perspective than exposing your hostname or local ip, but it will put you in a tiny cohort...

You may want to document the change in function of the `enable-webrtc-hide-local-ips-with-mdns` flag.

PF4Public:
@Ahrotahn Please voice your opinion on Zoraver's comment. Merging as is, or needs to be reworked?

PS: I second the "documenting" thing.

Ahrotahn:
It's a simple change to have something different returned for the IP, but I'm not sure what we could use that wouldn't have the same problem.  We could change it to "UNKNOWN" to match what uBlock Origin does, but since they removed the option we'd eventually end up in the same position.  I'll do some experimentation later today to see what happens when returning an empty or null string.  

I could add the documentation to docs/flags.md unless you think there's a better place to put that.

Zoraver:
> I could add the documentation to docs/flags.md unless you think there's a better place to put that.

Sounds good to me.

> It's a simple change to have something different returned for the IP, but I'm not sure what we could use that wouldn't have the same problem. We could change it to "UNKNOWN" to match what uBlock Origin does, but since they removed the option we'd eventually end up in the same position. I'll do some experimentation later today to see what happens when returning an empty or null string.

This is a tricky problem. If we are the only ones returning an empty or null string, we will run into the same issue...

I think that we may want to revisit hard-disabling mdns in the future with this issue in mind. Instead of hard-disabling the mdns responder, could we only enable it when `#enable-webrtc-hide-local-ips-with-mdns` is set?

@Eloston what was the original intent behind setting `enable_mdns=false`?

Ahrotahn:
The empty string thing was just a curiosity, it ends up being returned as `n/a`.  

So I dug into how the addresses are created and `CreateName()` in `services/network/mdns_responder.cc` literally just returns `base::GenerateGUID()`.  I honestly thought there would be more involved!  It's only another two-line change to have it return a random GUID, and the interesting thing is implementing it this way causes it to return a different address for each request!  This would make that address unable to be used as a means of unique identification.  

I'll push a new commit with that change and the documentation when I get some some time today.

PF4Public (approved):
Randomized response is definitely better.

wchen342:
Random response doesn't necessarily make it unidentifiable. But in this case if the browser is returning something random anyway then I guess it shouldn't be a big problem? I am not very sure on this though.

PF4Public:
Isn't it how "flag-fingerprinting-canvas-image-data-noise.patch" works?

Zoraver:
I have been thinking about the impact of returning a randomized response *without* mdns. At first glance, this change would make ungoogled-chromium with `enable-webrtc-hide-local-ips-with-mdns` set look the same as chromium with the same flag set (at least as far as webrtc fingerprinting goes). However, the address returned will not actually resolve on the local network since the mdns responder is not enabled. In the best case, this would just break certain applications of webrtc that use the local network for communication between peers. However, in the worst case, this *could* enable the browser to be fingerprinted as ungoogled-chromium (the only browser that would present an unresolvable mdns local address). Either way, I think this approach is not ideal.

Since my previous comment on this thread, I have spent some time looking into how other privacy-focused chromium-derived browsers approach this issue. One approach, which seems promising, used by Brave and Hexavalent, is described at the end of [this page](https://support.brave.com/hc/en-us/articles/360017989132-How-do-I-change-my-Privacy-Settings-). Hexavalent has two patches that [change the default policy to DefaultPublicInterfaceOnly](https://github.com/Hexavalent-Browser/Hexavalent/blob/15a66c953f7d189221a4ead5bf46adde08b07bb9/patches/0033-Hexavalent-set-webrtc-to-private-but-usable-default.patch) and [add a GUI chooser for the policy to the privacy page](https://github.com/Hexavalent-Browser/Hexavalent/blob/15244d557d9f8a6db7b61afad1f71d0de8f149f6/patches/0039-Add-WebRTC-policy-toggle-to-settings.patch). I think the cleanest approach to solving this problem would simply be to adopt these patches.

Ahrotahn:
That's definitely the better way to handle this then!

PR #1687, opened 2021-10-11
Add flags for existing switches

Ahrotahn:
This PR creates `chrome://flags` entries for existing Chromium switches and features.  
There have been many previous requests for flags that already exist, albeit only as command-line switches.  The most prominent ones are for starting in incognito and removing the tiles on the new tab page.  This also helps in situations where adding switches would be difficult, such as on mobile.  

The initial entries:  
* Disable Top Sites:  Removes the top sites and most visited tiles on the new tab page.
* Disable WebGL:  Disable all versions of WebGL.
* Force Dark Mode:  Forces dark mode in UI for platforms that support it.
* Incognito:  Starts the browser in incognito.  
* No Pings:  Disables ping attribute functionalty on \<a\> tags which can be a privacy concern.  
* Start Maximized:  Always starts the browser maximized.  
* Web UI Dark Mode:  Forces dark mode for internal chrome pages.  

This isn't meant to be an exhaustive list, but I'd like to add the more commonly used/requested ones.  
I'll keep this as a draft for a while so everyone has time to add suggestions for other useful entries.

Issue #1078, opened 2020-06-19
Licensing between GPLv3 and BSD 3-Clause


wchen342:
So I realized this when clearing out some documentations. `ungoogled-chromium` uses a BSD 3-Clause license, but `Bromite` uses GPLv3. AFAIK if you use GPL works in a BSD work you need to convert that work into GPL. Won't that be a problem for the current license of UC? Is it no one noticed that so far, or am I missing something?
@Eloston @csagan5

csagan5:
Chromium itself uses a mix of licenses for the different software components and its source code cannot be licensed under a single license; the license of ungoogled-chromium or Bromite applies only to the the additional patches.

The best would probably be to keep a reference to the license in the patches themselves.

I did not pick BSD because it allows re-packaging and selling.

wchen342:
Does it mean each patch is considered a separate work and they are indenpendent from each other? And what about distributing binaries compiled from UC, how do those count?

csagan5:
I do not see them as separate work, but you can modify the whole work by removing patches.

I consider the patches as a bundle, same as you can have "security enhancement" patches. You are provided the patches to use them during the build process instead of a stand-alone component.

> And what about distributing binaries compiled from UC, how do those count?

Licenses do not apply to binaries; binaries are released as freeware (it's no different than Chromium).

Feel free to explore this further, I have never dived into it. My goals are simple:
* open source nature of the project, so that other can continue/fork it
* avoid commercial exploitation

There are already some projects re-using the patches, although they not always refer back to Bromite (rebranding jobs).

PF4Public:
This is actually a good question.

On one hand, every patch by itself that touches chromium codebase is a derivative/modification of the said codebase as it inevitable (by the virtue of having context) contains snipplets of chromium code. Most of chromium is BSD, so this allows any modifications and derivatives. Good question is if anyone is allowed to re-license BSD code into GPL. But that feels to be about OK. So here we have one separate object with its own license. I may be wrong, but they are more like documentation of sorts. So that if you receive those patches and want to modify/distribute them further, you must abide by GPL rules. If you do this in-house for your own purposes — you don't have to.

[...]

Issue #429, opened 2018-06-21
[Proposal/Discussion] Fingerprinting mitigations from Bromite


csagan5:
I am opening this as a "ping" and to offer a better alternative to what was described in https://github.com/gcarq/inox-patchset/issues/52 (embedding addons is a bad idea).

Please see these patches that I developed for [Bromite](https://github.com/bromite/bromite):

* https://github.com/bromite/bromite/blob/master/patches/BRM048_User-Agent-anonymize.patch (minor, Android-focused)
* https://github.com/bromite/bromite/blob/master/patches/BRM049_Battery-API-return-nothing.patch
* https://github.com/bromite/bromite/blob/master/patches/BRM050_Disable-plugins-enumeration.patch
* https://github.com/bromite/bromite/blob/master/patches/BRM051_AudioBuffer-AnalyserNode-fingerprinting-mitigations-via-IDL.patch (might break stuff)
* https://github.com/bromite/bromite/blob/master/patches/BRM052_Add-fingerprinting-mitigation-for-getClientRects-getBoundingClientRect.patch (you already have an older version of this patch here: https://github.com/Eloston/ungoogled-chromium/blob/master/resources/patches/ungoogled-chromium/add-third-party-ungoogled.patch with the flag to turn it off/on that you implemented)
* https://github.com/bromite/bromite/blob/master/patches/BRM053_Canvas-fingerprinting-mitigations-for-image-data-font-metrics-and-webGL.patch (this is where I did the bulk of the work recently)

For these you probably would want to add a flag; I am not sure about a general flag for anti-fingerprinting or one for each aspect, I would be inclined for a general one, but I leave that aspect to you.

Eloston:
Thanks for letting me know about these features.

> For these you probably would want to add a flag; I am not sure about a general flag for anti-fingerprinting or one for each aspect, I would be inclined for a general one, but I leave that aspect to you.

While creating one flag for all fingerprinting deception would be less implementation work, I still want to give users more flexibility when it comes to features that change the browser's behavior.

> you already have an older version of this patch here

I don't see any new changes in the Bromite patch you linked me. Can you clarify?

> https://github.com/bromite/bromite/blob/master/patches/BRM049_Battery-API-return-nothing.patch

What purpose does this patch serve when the other battery API patch is also included?

csagan5:
> I don't see any new changes in the Bromite patch you linked me. Can you clarify?

[...]

PR #1845, opened 2022-02-21
Add WebRTC IP policy default and flag patches, add licenses to credits, remove/replace Bromite patches

### LINKED ISSUE #1725: WebRTC still leaks the local IP of users

Ahrotahn:
This is a continuation of #1750 and includes two patches from the Hexavalent project which set the default WebRTC IP handling policy and allows the user to change the policy in the settings.  A new sub directory has been created for these patches.  I tested using the same sites and everything seems to function as intended.  

Fixes #1725


networkException:
hm I'm not completely sure if we comply with the BSD-3-Clause-Clear license here:

```
Redistributions of source code must retain the above copyright notice,
this list of conditions and the following disclaimer.
```

I'm not sure if just specifying which license the patches are distributed under is actually enough, maybe we could add a new directory for full license files and reference them in the patches using the SPDX identifier?

Ahrotahn:
I've included the licence file in the directory for now.  We could do another PR to create a top-level licence directory with subdirectories for each project's licences since we don't currently include any others either.

Ahrotahn:
That last commit creates the new directory with subdirectories for each project.  I've included the license files for each, but it looks like debian and iridium both only include the upstream chromium license.  Let me know if you know if there's a better replacement.  
I'm not a lawyer so hopefully this is the best way to handle this.  

I considered moving the ungoogled-chromium license into the directory as well, but I know github and some other software defaults to the top-level license file so I left it as-is.

networkException (dismissed):
Works flawlessly

PF4Public:
Should we do anything else (in the future) with respect to licenses?

networkException:
I think this is pretty much all that we can do

Zoraver (commented):
This looks good to me. With regards to licenses, we should consider adopting [`Add-licencing-credits-to-about_credits.patch`](https://github.com/Hexavalent-Browser/Hexavalent/blob/05f2337febacc2b5e4eed5f07ae55cb3af44c9b7/patches/0041-Add-licencing-credits-to-about_credits.patch) from Hexavalent. It adds licenses to the `chrome://credits` page and would allow us to fulfill the "Redistributions in binary form" clause of the Clear BSD License.

Ahrotahn:
I've updated the patches for 99 and added a patch to include the licenses in the credits.  Everything seems to work as intended on my end.

networkException (dismissed):

PF4Public:
Just a FYI. I've asked @Eloston to glance through the changes in this PR.

Ahrotahn:
Updated for 100.0.4896.60

patches/core/ungoogled-chromium/add-licenses-to-credits.patch
```diff
@@ -0,0 +1,853 @@
...
+ )
+ 
+ 
+@@ -325,6 +330,36 @@ SPECIAL_CASES = {
+         "License Android Compatible": "yes",
+         "License File": "/third_party/swiftshader/LICENSE.txt",
+     },
++    os.path.join('bromite'): {
++      "Name": "Bromite",
++      "URL": "https://github.com/bromite/bromite",
```
Zoraver, line 823: This should probably be <https://bromite.org> instead of the repository URL.

Ahrotahn:
I've added the change for Bromite's link and rebased over the .75 update.

patches/core/ungoogled-chromium/add-licenses-to-credits.patch
```diff
@@ -0,0 +1,853 @@
+--- /dev/null
++++ b/LICENSE.bromite
+@@ -0,0 +1,674 @@
++                    GNU GENERAL PUBLIC LICENSE
++                       Version 3, 29 June 2007
```
Eloston, line 5: If I recall, the patches we got from Bromite are under our BSD 3-clause license, otherwise it would be tricky to figure out what parts of the code are under GPL v3 and what other parts are BSD 3-clause (particularly if we consider the patched code instead of the patch files; how does it work for patches that mostly remove code?)

I see two options:
1. We create a BSD 3-clause license that @csagan5 can agree with
2. We remove/rewrite (most of) the patches we incorporated from Bromite. I think we can remove the patches under `extra/bromite`, especially the fingerprinting patches since I'm not sure how useful those fingerprinting mechanisms are (people have reported websites breaking due to them).

PF4Public: Is it useful to split this PR into two then? The one with WebRTC-things and the other one fixing licenses?

qua3k: I've commented on https://github.com/bromite/bromite/issues/589 regarding that patch.

csagan5: @qua3k the patch has been tested to work as [intended](https://github.com/bromite/bromite/wiki/WebRTC#bromite-configuration), there are currently no known issues/bugs with it.

PF4Public:
@Eloston Please let us know how to go forward with this PR.

HMaker:
is it ready to be merged?

PF4Public:
Hexavalent seems to be no longer active :(

jstkdng:
it's always sad to see a project die...
Hexavalent updated its patches to v102.x so updating the patches could be required.

Ahrotahn:
I've been running with the default policy changed for most of the year and haven't yet run into any situations that required me to change the setting on the fly.  Unless there's some good reasons to keep the current patches, I think the best plan is to remove both the hexavalent and bromite patches and introduce our own that sets the default to DisableNonProxiedUdp and allows changing the setting via flag/switch.  This would make it conform to how the other changes we've added operate and reduce the maintenance burden.

Ahrotahn:
I've pushed a new set of changes rebased over the latest changes:

* Remove Bromite patches and replace the field trial patch with a new one
* Add licenses to credits
* Add a patch to set the default webrtc-ip-handling-policy to DisableNonProxiedUdp
* Add a flag for webrtc-ip-handling-policy

The webrtc-ip-handling-policy switch already exists in upstream Chromium so only the flag needed to be added. I'll probably integrate the flag into #1687 when I get some time to work on it again.

PF4Public:
> * Remove Bromite patches and replace the field trial patch with a new one

Why? did I miss something?

Ahrotahn:
The PR has been stagnant since April without a solution to the licensing issues, so I took option 2 and removed the patches.  
I'm open to suggestions for an alternative solution if you have any.

docs/flags.md
```diff
@@ -15,17 +15,14 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
...
   `--force-punycode-hostnames` | Convert all Internationalized Domain Names to punycode (ASCII representation of Unicode). See https://github.com/ungoogled-software/ungoogled-chromium/issues/370 for more details.
   `--hide-crashed-bubble` | Hides the bubble box with the message "Restore Pages? Chromium didn't shut down correctly." that shows on startup after the browser did not exit cleanly.
   `--http-accept-header` | Changes the default value of the `Accept` HTTP header sent with HTTP requests. Combined with `--disable-grease-tls` allows browser to look more like a tor-browser. See https://github.com/ungoogled-software/ungoogled-chromium/issues/783 for more details.
   `--keep-old-history` | Disables deletion of local browser history after 90 days
-  `--max-connections-per-host` | (from Bromite) Configure the maximum allowed connections per host. Valid values are `6` and `15`
```
Eloston, line 25: In addition, you may want to update some sections in the README:

1. https://github.com/ungoogled-software/ungoogled-chromium#borrowed-features
2. https://github.com/ungoogled-software/ungoogled-chromium#credits (maybe we can still credit them for inspiration?)
3. https://github.com/ungoogled-software/ungoogled-chromium#related-projects

patches/extra/ungoogled-chromium/default-webrtc-ip-handling-policy.patch
```diff
@@ -0,0 +1,11 @@
+--- a/chrome/browser/ui/browser_ui_prefs.cc
++++ b/chrome/browser/ui/browser_ui_prefs.cc
+@@ -93,7 +93,7 @@ void RegisterBrowserUserPrefs(user_prefs
+                                 false);
+ #endif
+   registry->RegisterStringPref(prefs::kWebRTCIPHandlingPolicy,
+-                               blink::kWebRTCIPHandlingDefault);
```
Eloston, line 7: I think we should clearly document all changes to default behavior (maybe in a new document) along with how to change it back to Chromium's default behavior (if possible). That way us developers and users can have a baseline to compare against vanilla Chromium.

Eloston (dismissed):
After viewing these changes with a fresh set of eyes, it LGTM. But a few optional documentation changes would help users a little.

Eloston:
Also related but I didn't want to comment on a closed issue: There was a question raised here for me that I forgot about: https://github.com/ungoogled-software/ungoogled-chromium/pull/1750#issuecomment-984305670

I don't recall the original intent, but I probably disabled this ages ago because I thought it was unnecessary to have. Web browsing has changed a bit since then, so I think it's better to re-enable it as long as we have reasonable solutions for all the privacy issues (e.g. is local IP obfuscation the only major issue with mDNS?)

docs/flags.md
Ahrotahn: I can remove that from the borrowed features section when I rebase over the new changes.  I think we should keep them in the credits and related projects though.  We could always reintroduce those patches if the licenses can be sorted out.

patches/extra/ungoogled-chromium/default-webrtc-ip-handling-policy.patch
Ahrotahn: There are occasional issues created by people running into some of the other changes to default settings, so this would be a good idea.  Though I think it'd be best to make that addition a separate PR considering how much is being done in this one already.

Ahrotahn:
I made a build with mDNS enabled and with its patch removed.  There were no issues building or running, and I don't see any new requests in wireshark or the logs.  Maybe some later changes solved the original issue.  I can make a separate PR later for that as well.

PF4Public:
Is there any possibility to keep those bromite patches? We may (probably) contact Carl and ask directly for permission to use or something like that.

>people have reported websites breaking due to them

Ain't those flags disabled by default? That is exactly why one cannot have both: perfect privacy and perfect functionality. Having the ability to choose is a worthy thing IMO. I noticed that at some point [Whats App was doing some fingerprinting in their web-version](https://github.com/ungoogled-software/ungoogled-chromium/pull/1964#issuecomment-1144678915) and indeed it was breaking the session upon each access (tab reopen, browser restart). Fortunately they abandoned this behavior very soon. Moral is… Well, there is no moral :)

PF4Public:
Since this PR has outgrown its critical mass and stalled irreversibly, I propose we merge only WebRTC-related changes (and maybe the replacement patch for field trials) and discuss the rest separately and with all the due thoroughness!

Ahrotahn:
That's fine with me, though I actually just finished updating the patchset for the latest version a little over half an hour ago.  I'll push the changes anyways once my build finishes and I get some time make sure everything is in order.  

The main blocker for this is the GPL licensing.  Every resource I have gone through states that we would have to re-license the entire project under a GPL-compatible license if we want to include those patches.  I was hoping that we could figure out an alternative, but I haven't been able to find any usable solutions online.

Ahrotahn:
Well since I had replaced some of the patches with my own they no longer need to be grouped in with the licensing stuff.  I think it would be a good idea to break some of those patches out from this PR now.  A lot has changed since this started and it might make things easier to sort out if they are separate.  

---

One idea I had to deal with the licensing would be to create a new repo for GPL-compatible patches.  We could update the scripts to check an external directory to include extra patches.  `add-ungoogled-flag-headers.patch` would make it easy to include extra flags in that situation.  This would allow each repo to have their own licensing.  The resulting binary would still be governed by the GPL if that repo is used, but we are already compliant due to having all the source publicly available under a permissive license.  It would just mean that anyone wanting to make proprietary builds wouldn't be able to include that repo.  

Granted, I'm a layman when it comes to legal stuff so I'm not sure if that is a good solution.

PF4Public:
>That's fine with me, though I actually just finished updating the patchset for the latest version a little over half an hour ago.

OK. Change of plans!

I took the initiative and did some "archeological excavations" and bored myself to near-death by reading License texts and looked through tons of PRs and issues (@Ahrotahn sorry, this took me so long to finish).

I apologise in advance for the lengthy message that I ended up with.

## Origins
In <s>a galaxy far away</s> 2018 Carl submitted a PR https://github.com/ungoogled-software/ungoogled-chromium/pull/377 with the first Bromite patch named `resources/patches/bromite/BRM045_Add-fingerprinting-mitigation-for-getClientRects-getBoundingClientRect.patch`. At that time UC was already licensed under BSD, so that means that knowingly or not, but Carl submitted that patch into the BSD licensed project without specifically stating the license, hence this patch _must_ be covered by the BSD license. Which allowed Eloston to modify this patch and adapt it for UC.

One could do the three-way comparison of the patches:

* Carl's original submission: [resources/patches/bromite/BRM045_Add-fingerprinting-mitigation-for-getClientRects-getBoundingClientRect.patch](https://github.com/ungoogled-software/ungoogled-chromium/pull/377/commits/4151259b3248f0fc5c42fa262a1d1dd43c39fb60)
* Modified version that was committed by Eloston: [/resources/patches/ungoogled-chromium/fingerprinting-flag-client-rects-noise.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/71f59c2556bb7d4fe442fa77eaee3f5227ce3cfc/resources/patches/ungoogled-chromium/fingerprinting-flag-client-rects-noise.patch)
* Present day version: [/patches/extra/bromite/fingerprinting-flags-client-rects-and-measuretext.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/patches/extra/bromite/fingerprinting-flags-client-rects-and-measuretext.patch)

With respect to this particular patch there is no license violations whatsoever.

[Every](https://github.com/ungoogled-software/ungoogled-chromium/commit/eb5aa1a0431937aede292acd2f9481d3a26ab575) [other](https://github.com/ungoogled-software/ungoogled-chromium/commit/009c8e19255a8a0255ec0be40eb4872471318463) [later](https://github.com/ungoogled-software/ungoogled-chromium/commit/d33495850fe545e2efc09cc7c3c749cb17e5e0df) [Bromite](https://github.com/ungoogled-software/ungoogled-chromium/commit/56065d81dcf1301d6e80c26e4397b754b6835e50) patch was committed by Eloston directly presumably by copying it from GPL-licensed Bromite tree into the BSD licensed UC tree. This fact deserves further investigation. I could not find any rigid proof of those patches being available under BSD license short of the following passing mention: https://github.com/ungoogled-software/ungoogled-chromium/issues/1078#issuecomment-648317691

> I think we have discussed the license for the current Bromite patches in UG already (need to find the issue/PR)

Please note that neither the issue/PR were mentioned nor I could find one. Could be they discussed that offline or out-of-band.

I could find the original proposal from Carl https://github.com/ungoogled-software/ungoogled-chromium/issues/429 (emphasis mine):

>I am opening this […] **to offer** a better alternative […].
>
>Please **see** these patches that I developed for Bromite
>
> […]

Overall in that issue Carl and Eloston were deeply involved in discussing technical details, but not even a passing mention of the licensing terms.

So to conclude I would consider `fingerprinting-flags-client-rects-and-measuretext.patch` as a modified/derivative work of a BSD-licensed (specifically for UC) patch from Carl. Other Bromite patches are taken verbatim(?) and are covered by the GPL with the possibility of Carl granting explicit permission to treat them under BSD license for UC project in the past. 

## Bromite's GPL

Carl mentioned a number of times why he choose GPL for Bromite. Also I found many occasions where he was very generous and willing to share his work for other projects to use. He was also very kind of immediately providing me with a list of patches which might be of interest for UC. He is however very strict about authorship https://github.com/ungoogled-software/ungoogled-chromium/commit/71f59c2556bb7d4fe442fa77eaee3f5227ce3cfc#comments which is very understandable. On the issue I will link further down he even hinted the possibility of re-licensing Bromite patches of his authorship into BSD specifically for UC.

## Starting point

While researching I found a good starting point: https://github.com/ungoogled-software/ungoogled-chromium/issues/1078. Turns out this issue was already discussed once. I kindly ask everyone interested to glance through that issue for a refresher before continuing.

Now, first of all I'd like to mention that I totally disagree with Carl's  statement in *"I do not see them as **separate** work"* (I'll tackle this later). Also I disagree with former me in *" […] those patches, they are **not like code**, they **never link or compile** into resulting binary, we kinda utilise them during build and they are **never included** in chromium by themselves […]"*.

So lets move on.

To begin with, lets assume there could be a project consisting of many works each being covered by its own license. Chromium being the obvious example.

## Ungoogled-Chromium

Lets handle Ungoogled-Chromium in that perspective. Assume it being licensed under BSD, but it also could have some *independent* components with own licensing terms. Indeed we can succeed in applying ungoogled-chromium modifications completely excluding Bromite or Inox patches (or both!). Which allows me to state that we have a BSD licensed work in most folders (utils, docs and so on) and we may have any number of separate works in patches.

Now lets try including GPL work as part of an *aggregate* combined with Ungoogled-Chromium.

## The GPL

Upon first inclusion of a GPL code into our *aggregate* we start *Conveying Verbatim Copies* which is covered by chapter 4 of the GPL. This allows us to distribute the *source code*, but we should also keep license notices[^1]. We should improve our license notice presentation for borrowed patches actually.

Looks good so far.

Now lets modify this GPL work as we usually do routinely update to corerspond to new Chromium sources. This is covered by the next chapter. We can do the same as previously stated[^2].

There is only one difference this time: *"You must license the entire work, as a whole, under this License to anyone who comes into possession of a copy"* and one could argue that we should relicense UC to comply, but that is not the case. Like I've mentioned, this becomes an *aggregate*[^3] of independent works, which does not require relicensing as a whole. In other words, the *whole work* is the Bromite patch in question or the collection of them[^4].

That being said it turns out that every repository under https://github.com/ungoogled-software is completely compliant with respect to GPL licensed parts.

But it gets very interesting if we consider the resulting Chromium codebase after GPL patches are applied. This is no longer an *aggregate* as they are *"combined with it such as to form a larger program"*.

So lets advance to binary distribution. This is predictably the following chapter.

Per se we cannot compile patches to object code, which means we have no way of distributing in non-binary form. That's it! So Long, and Thanks for All the Fish :)

Jokes aside, we do as a matter of fact provide binaries. This looks like this: There are BSD licensed tools (UC) which make use of GPL-licensed and modified works (Bromite patches for instance); this *aggregate* is lawfully and publicly distributed. Then someone takes this *aggregate* and with the help of it (using it as a tool of sorts) creates a modified Chromium codebase that has GPL-licensed patches applied, which makes in turn *this* modified Chromium codebase to be in need of being GPL-licensed. And this is the grey area: whether it is possible to relicense a collection of MIT/BSD/Apache sources into GPL. I would guess that this is OK since there is already a GPL code in Chromium:
image

One could argue that someone distributing this modified GPL-containing Chromium should also distribute the *"Corresponding Source"* and with that all the UC codebase and should arguably turn it into GPL[^6]. From the definition of the Corresponding Source[^5] follows that it is *"the source code needed to generate, install, and (for an executable work) run the object code"*, but that is essentially Chromium's build system + patch utility. Chromium sources are already available and the sources of the patch utility are covered by this exception: *"However, it does not include […] general-purpose tools or generally available free programs which are used **unmodified** in performing those activities but which are not part of the work"*. One could consider UC a "convenience script", not part of *"Corresponding Source"* since it is essentially patch utility that is embedding the GPL code into Chromium sources for further compilation. But all in all it is still beneficial in public interest that all the sources of this convenience script are publicly available and licensed under BSD.

## Bottom line (personal interpretation)

Considering that all sources of UC are anyway publicly available and are compatible with GPL license, there are no GPL violation be it source form, or binary. There is however one thing that we could improve: license notice presentation and license delineation between UC's BSD code and any other borrowed code covered by other licenses. I think it would be sufficient to make direct statement in borrowed patches of their license, no need to split the repo. Also I think it would be an excellent idea to use the already existing Chromium code infrastructure to make specific mention of the licenses of all the borrowed patches.

There is also a possibility (mentioned above) that Carl will allow UC to use Bromite patches of his authorship under BSD license within UC repository. Of-course this is not prohibited by the GPL[^7].

It could be that the statement *"we can succeed in applying ungoogled-chromium modifications completely excluding Bromite or Inox patches (or both!)"* is not entirely true and if that's the case, we have yet another point for improvement.

[^1]: You may convey verbatim copies of the Program's source code as you
receive it, in any medium, provided that you conspicuously and
appropriately publish on each copy an appropriate copyright notice;
keep intact all notices stating that this License and any
non-permissive terms added in accord with section 7 apply to the code;
keep intact all notices of the absence of any warranty; and give all
recipients a copy of this License along with the Program.
You may charge any price or no price for each copy that you convey,
and you may offer support or warranty protection for a fee.

[^2]:You may convey a work based on the Program, or the modifications to
produce it from the Program, in the form of source code under the
terms of section 4, provided that you also meet all of these conditions:

    a) The work must carry prominent notices stating that you modified
    it, and giving a relevant date.

    b) The work must carry prominent notices stating that it is
    released under this License and any conditions added under section
    7.  This requirement modifies the requirement in section 4 to
    "keep intact all notices".

    c) You must license the entire work, as a whole, under this
    License to anyone who comes into possession of a copy.  This
    License will therefore apply, along with any applicable section 7
    additional terms, to the whole of the work, and all its parts,
    regardless of how they are packaged.  This License gives no
    permission to license the work in any other way, but it does not
    invalidate such permission if you have separately received it.

    d) If the work has interactive user interfaces, each must display
    Appropriate Legal Notices; however, if the Program has interactive
    interfaces that do not display Appropriate Legal Notices, your
    work need not make them do so.

[^3]:A compilation of a covered work with other separate and independent
works, which are not by their nature extensions of the covered work,
and which are not combined with it such as to form a larger program,
in or on a volume of a storage or distribution medium, is called an
"aggregate" if the compilation and its resulting copyright are not
used to limit the access or legal rights of the compilation's users
beyond what the individual works permit.  Inclusion of a covered work
in an aggregate does not cause this License to apply to the other
parts of the aggregate.

[^4]:BTW this is probably the misunderstanding that led me to disagree with Carl above. If he meant a collection of Bromite patches as a *whole work*, then it is totally agreeable.

[^5]:The "Corresponding Source" for a work in object code form means all
the source code needed to generate, install, and (for an executable
work) run the object code and to modify the work, including scripts to
control those activities.  However, it does not include the work's
System Libraries, or general-purpose tools or generally available free
programs which are used unmodified in performing those activities but
which are not part of the work.  For example, Corresponding Source
includes interface definition files associated with source files for
the work, and the source code for shared libraries and dynamically
linked subprograms that the work is specifically designed to require,
such as by intimate data communication or control flow between those
subprograms and other parts of the work.

[^6]:You may convey a covered work in object code form under the terms
of sections 4 and 5, provided that you also convey the
machine-readable Corresponding Source under the terms of this License

[^7]:This License gives no permission to license the work in any other way, but it does not invalidate such permission if you have separately received it.

Ahrotahn:
Wow I didn't realize what you meant previously, I didn't mean to make things difficult.

So much of the language for the GPL is based around a code-to-binary pipeline. For example an aggregate, [as described on gnu.org](https://www.gnu.org/licenses/gpl-faq.en.html#MereAggregation), only considers the output of GPL code and not the input. The aggregate is fine as long as the binaries are separate. Chromium is able to use Speech Dispatcher since the included code is only to link with the separate system library. They wouldn't be able to statically build with that library without relicensing everything under the GPL. But our situation is different.

> Per se we cannot compile patches to object code, which means we have no way of distributing in non-binary form. That's it! So Long, and Thanks for All the Fish :)

This is the crucial component that has been making things difficult. There doesn't seem to be a consensus on how licensing *patches* works. They aren't, on their own, able to produce binaries. Their output is only modified code. There is even more ambiguity when it comes to licensing the *removal* of code.

I'll edit the link in once I find it in my history, but I had found a reference online where RMS considered patches to be a derivative work of the code it modifies. If the code were licensed under the GPL, then the patch modifying that code would also have to be licensed under a GPL-compatible one. This, to me, implies that the original codebase would retain its license. So BSD code wouldn't need to be relicensed under the GPL due to applying a GPL-licensed patch like I had thought. 

The other stumbling block that you mentioned is whether or not a patch is considered it's own, standalone 'work'. If a patch can be considered a derivative work then I think that might be the case. That would mean that we should be able to have a mixture of differently-licensed patches so long as the licenses are included. 

So we could either separate the patches into folders with a LICENSE file or include the license as a comment at the top of each patch. I'll spend some time later updating things to include one of those options instead of removing the patches.

PF4Public:
> So much of the language for the GPL is based around a code-to-binary pipeline. For example an aggregate, [as described on gnu.org](https://www.gnu.org/licenses/gpl-faq.en.html#MereAggregation), only considers the output of GPL code and not the input. The aggregate is fine as long as the binaries are separate.

You're referring to a FAQ, which might be considering one of the most common cases, just one of them, not necessarily invalidating other cases. I was referring to license text itself. The term _“aggregate”_ is defined within chapter _"5. Conveying Modified Source Versions."_, which gives me the impression of it [aggregate] being able to emerge even in source-form, which fits us perfectly!

>Chromium is able to use Speech Dispatcher since the included code is only to link with the separate system library. They wouldn't be able to statically build with that library without relicensing everything under the GPL. But our situation is different.

I also wanted to touch this thing, but somehow forgot about that, thanks for reminding! Like I've stated above, UC in its source form as an aggregate of differently-licensed works has nothing to do with GPL propagation. This issue arises to someone who uses UG and produces a GPL-containing Chromium codebase. Since we provide binaries, we have to deal with it too. Myself included: I also [_"convey"_ binaries](https://github.com/PF4Public/ungoogled-chromium-bin/releases).
In that respect we have following possibilities:

* It is impossible for anyone to distribute GPL-containing Chromium in binary form because it would be unfeasible to try and modify GPL code to be present in dynamically loading library to satisfy GPL limitations due to their intrusiveness. It would be also unenforceable to relicense all the Chromium code into GPL since anyone could use the same code under the original license (BSD/MIT and so on). This has one peculiar consequence: even Bromite would be in violation of GPL if they distribute binaries. W.r.t. UC this would simply require some mechanism to skip applying GPL patches if the resulting binary is to be distributed. This does not apply if building for oneself, GPL has a clause for that: _"To “propagate” a work means […], except executing it on a computer or modifying a private copy."_
* If there would be no disagreement between BSD-licensed Chromium and GPL code inside the binary representation, then there would be no problem distributing the resulting binary, provided that we comply with GPL, meaning we release the source code, we release instructions on how to build this code and retain license notices (more-or-less what we already do).

>That would mean that we should be able to have a mixture of differently-licensed patches so long as the licenses are included.
>So we could either separate the patches into folders with a LICENSE file or include the license as a comment at the top of each patch. I'll spend some time later updating things to include one of those options instead of removing the patches.

That's what I was trying to convey and prove, hope it worked.

>I'll edit the link in once I find it in my history, but I had found a reference online where RMS considered patches to be a derivative work of the code it modifies.

Please do! It could be insightful if it has a reasonable grounding.

uazo:
@csagan5 have you ever evaluated the GPL Linking Exception or Classpath Exception?

```
The GPL Linking Exception allows developers to use certain code, called library code, in commercial projects, 
without the application of GPL to the resulting software. Linking, in this case, is the process of
connecting the main code with a library code. If a library is released under GPL with a 
linking exception, said library can be used to create software to which GPL does not apply.
```

```
...Classpath Exception...
…the copyright holders of this library give you permission to link this library with independent modules … 
and to copy and distribute the resulting executable under terms of your choice, provided that you also meet, 
for each linked independent module, the terms and conditions of the license of that module.
```

Ahrotahn:
That at least means there shouldn't be any problems with the current patchset. Though in that case having a LICENSE file per directory would be complicated since the current patches and new patches (if they can be added) would have different licenses.  

I've recently learned about [REUSE](https://reuse.software/) where a top-level directory is made containing each of the licenses and each file would include a pair of lines with the copyright info and a link to the respective license.  That might make managing this easier. If that seems like a good solution I'll spend some time setting that up.

Ahrotahn:
The copyright holders will need to be added along with the license.  We are allowed to make the copyright for groups: "The ungoogled-chromium Authors", "The Iridium Authors", and so on.  I'd be interested if y'all prefer that implementation or if we should list everyone that worked each file individually.  

@csagan5 Were you the sole author for the Bromite patches when they were submitted?  Would you prefer to be listed individually or would "The Bromite Authors" be sufficient? 

Also the copyright year is optional.  I think we shouldn't include the year in order to reduce the number of future changes needed, but I'd be interested in your thoughts regarding that.

PF4Public:
> The copyright holders will need to be added along with the license. We are allowed to make the copyright for groups: "The ungoogled-chromium Authors", "The Iridium Authors", and so on. I'd be interested if y'all prefer that implementation or if we should list everyone that worked each file individually.

Is it a requirement of REUSE you've mentioned? Could you please elaborate on that? I take the need for noting the authors as granted, why would someone doubt that? Or am I misunderstanding something?


> Also the copyright year is optional. I think we shouldn't include the year in order to reduce the number of future changes needed, but I'd be interested in your thoughts regarding that.

Totally agree!

Ahrotahn:
Yes, the only two things that need to be added is a line for the license and a line for each copyright holder.  
For our situation the copyright holder should be the same as the author, but it's done that way because there can be situations were an author is employed to write code and the copyright holder for that code would be the employer instead of the author.

PF4Public:
Some of those patches were additionally modified by Eloston. Does it contribute to authorship?

PF4Public:
@Ahrotahn would it be possible [as proposed earlier](https://github.com/ungoogled-software/ungoogled-chromium/pull/1845#issuecomment-1349700372) to extract WebRTC-related changes into separate PR in order to <s>merge them faster</s> merge them at last? That would be a valuable addition to ungoogled-chromium.

Ahrotahn:
Sure, I'll start sorting things out this week if all goes well.

PR #1876, opened 2022-03-27
Update to Chromium 100.0.4896.60

Ahrotahn:
This PR updates Chromium to 100.0.4896.60.  

Notes:

* The jabber section in the trk patch has been removed due to the removal in [[3416696]](https://chromium-review.googlesource.com/c/chromium/src/+/3416696)

* The FLoC patch has been removed since the code has finally been removed in [[3424359]]( https://chromium-review.googlesource.com/c/chromium/src/+/3424359)

* webstore patch changes due to [[3440399]](https://chromium-review.googlesource.com/c/chromium/src/+/3440399)

* safebrowsing patch addition due to [[3312617]](https://chromium-review.googlesource.com/c/chromium/src/+/3312617)

* remove-unused-preferences-fields patch changes for [[3432769]](https://chromium-review.googlesource.com/c/chromium/src/+/3432769) and an include for safety_tips_config.cc

* Tab audio muting has returned! [[3123736]](https://chromium-review.googlesource.com/c/chromium/src/+/3123736)
  This can be enabled with `chrome://flags/#enable-tab-audio-muting` or `--enable-features=EnableTabMuting`

<br>

[Change log](https://omahaproxy.appspot.com/changelog?old_version=99.0.4844.84&new_version=100.0.4896.60)
[Blog post](https://chromereleases.googleblog.com/2022/03/stable-channel-update-for-desktop_29.html)


PF4Public:
Builds and runs :)

kramred:
Builds and runs as Linux portable and on macOS

patches/extra/ungoogled-chromium/disable-download-quarantine.patch
```diff
@@ -9,8 +9,8 @@
 -#include "components/services/quarantine/quarantine.h"
  #include "crypto/secure_hash.h"
  
- #if BUILDFLAG(IS_WIN)
-@@ -529,105 +528,6 @@ DownloadInterruptReason BaseFile::Publis
+ #if BUILDFLAG(IS_ANDROID)
```
PF4Public, line 12: Is it as intended? I mean WIN → ANDROID

PF4Public:
Quick google-ing shows (yeah, ironic) that FLoC has been rebranded into the new Topics API. @Ahrotahn have you noticed anything that needs to be removed in a similar way as was the case with FLoC?

patches/extra/ungoogled-chromium/disable-download-quarantine.patch
Ahrotahn: Yup, that is a context line rather than a change.  
They removed that section between versions:  [99.0.4844.84](https://source.chromium.org/chromium/chromium/src/+/refs/tags/99.0.4844.84:components/download/internal/common/base_file.cc;l=28) vs [100.0.4896.60](https://source.chromium.org/chromium/chromium/src/+/refs/tags/100.0.4896.60:components/download/internal/common/base_file.cc;l=28)

Ahrotahn:
The intent for Topics was only put up a few days ago:  [link](https://groups.google.com/a/chromium.org/g/blink-dev/c/oTwd6VwCwqs)
They plan to start the trial for version 101.  It should be controlled by the privacy sandbox, which we have a patch to disable already.

PF4Public (commented):
LGTM. We might better wait a little bit for if @Zoraver has time for a review.

networkException (approved):

Zoraver:
> LGTM. We might better wait a little bit for if @Zoraver has time for a review.

I don't have time to rigorously review this right now. If I get around to it I will leave my feedback here, but you shouldn't let my time constraints block the release. Once #1845 has been rebased for Chromium 100, I will make reviewing that pull request a priority.

One thing that may be of interest is that I was recently able to compile a near-vanilla copy of Chromium with `enable_reporting = false` in `args.gn` *without* `fix-building-without-safebrowsing.patch` applied. This suggests to me that `fix-building-without-safebrowsing.patch` may no longer be needed.

PF4Public:
@networkException should we merge then?

@Zoraver good find! But aren't safebrowsing and reporting separate things?

networkException:
yep I'll release it

Ahrotahn:
@PF4Public That is a uc-flavored Freudian slip 🤣.  

I can confirm that building and running without `fix-building-without-enabling-reporting.patch` works fine.  I had noticed a few commits since the beginning of the year to make building work with the flag disabled but I never took a deeper look.  
I'll make a PR to remove the patch later or include it with the next version bump if that happens before I can do that.

PR #1905, opened 2022-04-10
New issue templates

PF4Public:
I've updated issue templates and I'd like to get some feedback before squashing and marking as ready for review.

Updated templates are [Bug Report](https://github.com/PF4Public/ungoogled-chromium/issues/new?assignees=&labels=bug&template=bugreport.yml), [Feature request](https://github.com/PF4Public/ungoogled-chromium/issues/new?assignees=&labels=enhancement%2Chelp+wanted&template=feature_request.yml) and [Other](https://github.com/PF4Public/ungoogled-chromium/issues/new?assignees=&labels=&template=other.yml). Feel free to test them in my fork and post your suggestions for improvements.


nchristopher:
I think the "Other" issue template isn't necessary, and sort of defeats the point of issue templates. You could just allow blank issues instead.

.github/ISSUE_TEMPLATE/bugreport.yml
```diff
@@ -0,0 +1,102 @@
...
+  - type: checkboxes
+    attributes:
+      label: Tested vanilla Chromium?
+      options:
+        - label: I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
+  - type: checkboxes
+    attributes:
+      label: Tested Chrome?
+      options:
+        - label: I have tried reproducing this issue in Chrome and it could not be reproduced there
```
nchristopher, line 58: This should probably be one set of checkboxes; IMO it looks a bit cleaner.
```suggestion
  - type: checkboxes
    attributes:
      label: Tested upstream?
      options:
        - label: I have tried reproducing this issue in Chrome and it could not be reproduced there
        - label: I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
```

.github/ISSUE_TEMPLATE/bugreport.yml
```diff
@@ -0,0 +1,102 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        Before submitting this issue, please confirm if you should submit it to a [platform-specific issue tracker instead](//github.com/Eloston/ungoogled-chromium/blob/master/SUPPORT.md#platform-specific-support)
+
+        Otherwise, your issue may not be noticed. In addition, please read through the [SUPPORT.md](//github.com/Eloston/ungoogled-chromium/blob/master/SUPPORT.md) first
+  - type: dropdown
+    id: os
+    attributes:
+      label: OS/Platform
```
nchristopher, line 14: "OS/Platform" should just be replaced with "Platform" pretty much everywhere.

nchristopher (changes requested):

.github/ISSUE_TEMPLATE/bugreport.yml
PF4Public: Why?

.github/ISSUE_TEMPLATE/bugreport.yml
PF4Public: Reasonable.

.github/ISSUE_TEMPLATE/bugreport.yml
nchristopher: > Why?

It's shorter, and "OS/Platform" is sorta redundant. Just my opinion though, it's up to you.

networkException (approved):
Looks good to me! Besides the currently ongoing wording discussions I think we can basically ship this as is and wait for people to find ways to break it :)

.github/ISSUE_TEMPLATE/bugreport.yml
PF4Public: One could be using Ubuntu (which is an OS, not a platform) and have ungoogled-chromium there. Also one could install Flatpak (which is more like a platform, than an OS) in Ubuntu and then use ungoogled-chromium from within Flatpak. Therefore I put "OS/Platform" there.

Ahrotahn (approved):
This looks good to me, we can always alter it later if needed!

PR #1926, opened 2022-04-23
Update to Chromium 101.0.4951.41

Ahrotahn:
This PR updates to Chromium 101.0.4951.41.  

Notes:

* Small change in disable-webstore-urls.patch to handle the new task setup: [[3477059]](https://chromium-review.googlesource.com/c/chromium/src/+/3477059)

* New additions to fix-building-without-safebrowsing.patch for the changes in [[3381418]](https://chromium-review.googlesource.com/c/chromium/src/+/3381418) and [[3529377]](https://chromium-review.googlesource.com/c/chromium/src/+/3529377).  Also the change in [[3512993]](https://chromium-review.googlesource.com/c/chromium/src/+/3512993) had me realize that portion could be simplified by using the existing ifdef instead of removing it's contents.  

* The new side panel button can be removed with:
`chrome://flags/#side-panel`
or
`--disable-features=SidePanel`

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/100.0.4896.127..101.0.4951.41?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2022/04/stable-channel-update-for-desktop_26.html)


PF4Public:
In case anyone builds with pgo, they seem to be using clang-14 now:

>error: Could not read profile ../../chrome/build/pgo_profiles/chrome-linux-4951-1650389737-545882ae3e5c9e7f75e0781c8dc0d1eafb667efd.profdata: unsupported instrumentation profile format version
1 error generated.

PF4Public:
Builds and runs without PGO.

W.r.t. PGO clang-15 seems to have no objections:
image

jstkdng:
Works on my machine™
image

networkException (approved):

a61kt:
@networkException it works in my enviorment (i use obs)
https://build.opensuse.org/package/binaries/home:Area51Kacz:branches:home:ungoogled_chromium:testing/ungoogled-chromium/Arch

PF4Public (approved):
Worksforme :)

networkException:
I think we can release "early" again (as usual, really) and see if anything comes up in later reviews / issues. What do you think?

berkley4:
> I think we can release "early" again (as usual, really) and see if anything comes up in later reviews / issues. What do you think?

Given the number of CVEs fixed in each release, I would be in favour of releasing early (after basic testing). This should definitely be the rule if there are any active exploits, of which there have been a few in recent months.

PR #1949, opened 2022-05-09
Add flag for referrer header directive

### LINKED ISSUE #1918: Disable the (cross-origin) Referer header
**Is your feature request related to a problem? Please describe.**
Chromium and all Chromium-based browsers send the HTTP Referer header across different website origins. This can be a privacy leak.

A flag to disable the cross-origin Referer header would be a helpful addition. I think it could be on by default.

Extensions (I believe uMatrix) have blocked this in the past, but it would much better be done as part of the browser.

**Additional context**
This is similar but different than #1117. I think that's just reduced granularity, but a website operator can still see from what website you clicked the link on. This still allows for malicious redirects.

I've never seen a useful use of the Referer header. On the other hand, I've seen malicious uses: for example, when [clicking a link to jwz.org from news.ycombinator.com](https://news.ycombinator.com/item?id=31063136) [semi-nsfw].

Setting the `network.http.referer.XOriginPolicy` to 1 or above in Firefox fixes this, but there isn't yet an equivalent Chromium flag.

Ahrotahn:
This PR adds a patch that allows changing the amount of information passed through the referrer header.  
Using the no cross-origin option removes the referrer for cross-origin requests, while the minimal option also strips same-origin referrers to just the origin instead of the full URL.  

This can help prevent tracking, but in some instances can also cause some problems for sites that rely on the referrer to function correctly.  
The referrer header can be disabled completely by launching with the existing feature:  `--enable-features=NoReferrers`.  

Fixes #1918

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -22,6 +22,7 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
   `--max-connections-per-host` | (from Bromite) Configure the maximum allowed connections per host. Valid values are `6` and `15`
   `--omnibox-autocomplete-filtering` | Restrict omnibox autocomplete results to a combination of search suggestions (if enabled), bookmarks, and internal chrome pages.  Accepts `search`, `search-bookmarks`, `search-chrome`, and `search-bookmarks-chrome`.
   `--popups-to-tabs` | Makes popups open in new tabs.
+  `--referrer-directive` | Allows setting a custom directive for referrer headers. Accepts `nocrossorigin`, `minimal`, and `noreferrers`. The no cross-origin referrer option removes all cross-origin referrers, the minimal option removes all cross-origin referrers and strips same-origin referrers down to the origin, and the no referrers option removes all referrers.
 
 - ### Available only on desktop
 
--- a/patches/extra/ungoogled-chromium/add-flag-for-referrer-header.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-for-referrer-header.patch
@@ -30,7 +30,7 @@
  #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
 --- a/content/browser/utility_process_host.cc
 +++ b/content/browser/utility_process_host.cc
-@@ -224,6 +224,7 @@ bool UtilityProcessHost::StartProcess()
+@@ -225,6 +225,7 @@ bool UtilityProcessHost::StartProcess()
  
      // Browser command-line switches to propagate to the utility process.
      static const char* const kSwitchNames[] = {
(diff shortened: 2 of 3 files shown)


PF4Public:
> The referrer header can be disabled completely by launching with the existing feature: `--enable-features=NoReferrers`.

Is this feature not available via flags? Only commandline?

Ahrotahn:
Yes, it's a pre-existing feature in upsteam chromium and doesn't have a flag entry.  
I could include that as an option in the current flag if it makes sense.  
Maybe that should be it's own separate flag?  If so, then it might be something that could be included in #1687.  That PR has been hanging around for a while, I might need to revisit it next time I get the spare time.

Ahrotahn:
I have rebased the patch over yesterday's update and added the `noreferrers` option, which removes all referrers.

PF4Public:
On POST requests `minimal` does preserve url. Is it how it is supposed to be? All GETs don't have anything after domain, but POSTs have. It could be that XMLHttpRequests are affected.

Ahrotahn:
It should be stripping the URL down, but maybe there's something I missed?  How are you testing those POST requests?  

Using https://www.darklaunch.com/tools/test-referer to test everything except for the cross-origin referrer:
Screenshot_20220513_074951

PF4Public:
> How are you testing those POST requests?

I've inspected nginx logs on my personal website, but that's easy to reproduce: open devtools, filter `Fetch/XHR` write something and hit `Preview` (seems to work even by hovering over the button). XMLHttpRequests should fire.
image

At the same time GET is as expected:
image

Ahrotahn:
Thanks, that helped me figure out what is happening; asynchronous requests are handled differently.  I already have a working solution, but since those requests aren't run on the main thread I'll have to either change the patch to use feature flags or pass the switch through the network context.  I'll test what seems to work better this weekend as I get the time.

Ahrotahn:
It turns out that UtilityProcessHost will copy over command line switches.  This makes things easier and means the flags can stay the same as they were.

PF4Public (dismissed):
Works!

networkException (approved):
Works nicely!

I'd say we can ship this alongside 102, what do you think?

PF4Public:
Won't they conflict? Maybe after merging #1964

networkException:
I built with this branch rebased on top of the update, it didn't complain and everything works as expected

PF4Public:
🤷 I remember there were some hard-to-catch issues once, when we merged feature-PR before the version-PR.

networkException:
I mean we can merge the update first of course, though the result on disk will be the same regardless of the order in git

PF4Public:
Merging this PR first might make version-PR in need for a rebase due to failing tests, which drops reviews and so on and so forth. Or may not, Im not certain…

Ahrotahn:
The offset on line 33 will have to be incremented for the checks on github to pass, but it won't cause a problem when building.  

The PR that is merged after the first will fail the checks if it's not updated to include that offset change.  The order doesn't matter, I'll just need to add a new commit to the second PR which will invalidate the existing approvals.  

It's a shame, if there wasn't that single line added to utility_process_host.cc in 102 both PRs would have been able to be done in any order without change!

PF4Public:
Good question is: are we tagging new version after merging this PR or not? :)

networkException:
I'd say we tag after merging both

networkException:
lets see

networkException (approved):
I do, in fact, approve my own changes yes

PF4Public (approved):

PR #1970, opened 2022-05-29
Adjust argument length assertion in add-suggestions-url patch

networkException:
Chromium commit https://source.chromium.org/chromium/chromium/src/+/d30ba01286cdedafe6a77e91950854da81acb4f4
added an assertion to check that the args list has a size of 3. As the
patch adds the suggestions url as an additional argument, the browser
would crash when reaching the code.

This only affects ungoogled-chromium release 102.0.5005.61-1 as previous
chromium releases didn't include this commit.

First reported here https://github.com/ungoogled-software/ungoogled-chromium-archlinux/issues/189.

### DIFF
--- a/patches/extra/ungoogled-chromium/add-suggestions-url-field.patch
+++ b/patches/extra/ungoogled-chromium/add-suggestions-url-field.patch
@@ -379,7 +379,12 @@
    else
      NOTREACHED();
  
-@@ -373,14 +380,17 @@ void SearchEnginesHandler::HandleSearchE
+@@ -369,18 +376,21 @@ void SearchEnginesHandler::HandleSearchE
+   if (!edit_controller_.get())
+     return;
+ 
+-  CHECK_EQ(3U, args.size());
++  CHECK_EQ(4U, args.size());
    const std::string& search_engine = args[0].GetString();
    const std::string& keyword = args[1].GetString();
    const std::string& query_url = args[2].GetString();


networkException:
btw I intend to publish this as ungoogled-chromium 102.0.5005.61-2 if possible

PF4Public (approved):
I second tagging 102.0.5005.61-2

Ahrotahn (approved):
Good catch!

networkException:
I included the tor flags in the tag as well

Maybe in the future we could have a branch for each chromium release with added commits resulting in rel bumps so that feature additions would be isolated until a new chromium (perhaps even major) release

What do you think?

teeminus:
> I second tagging 102.0.5005.61-2

If the new version is "-2", then the `revision.txt` file needs to be changed as well, correct? Otherwise the new package version is still "-1".

networkException:
Aw crap

Do you know what depends on that?

teeminus:
For windows it's the zip and installer file name, I would guess for the other systems it's similar.

networkException:
Do you think that its possible to work around that for this release? I wouldn't really know a good way of fixing this retroactively

teeminus:
You can either rename the file manually afterwards, change the code That generated the file name (which needs to be reverted for the next release) or create a new "-3" release with the revision file fixed (and maybe delete the "-2" release to prevent confusion).

networkException:
We could also wait for #1974 to be resolved to tag -3

teeminus:
That's true. Then I will stop updating my windows PR to -2 and wait for -3 😃

PF4Public:
I wonder if it would be possible to get rid of `revision.txt`. Does it serve any purpose anyway?

teeminus:
Distinguish between zip/installer files compiled from the same chromium release but with different patch sets I would guess.

teeminus:
We could use the number of commits to the master branch instead of the revision file.

networkException:
I guess its needed when using git submodules, at least I think that always just references a commit

PF4Public:
OK, lets turn it the other way around:
I use [the following](https://github.com/PF4Public/gentoo-overlay/blob/22e16445ac3a17285963a9edb67f71e0bce390d9/www-client/ungoogled-chromium/ungoogled-chromium-102.0.5005.61_p3.ebuild#L14-L29) to construct the download URL:
```sh
UGC_PV="${PV/_p/-}"
UGC_PF="${PN}-${UGC_PV}"
UGC_URL="https://github.com/Eloston/${PN}/archive/"
UGC_URL="${UGC_URL}${UGC_PV}.tar.gz -> ${UGC_PF}.tar.gz"
```
With the file `ungoogled-chromium-102.0.5005.61_p3.ebuild` it constructs the following URL: `https://github.com/Eloston/ungoogled-chromium/archive/102.0.5005.61-3.tar.gz -> ungoogled-chromium-102.0.5005.61-3.tar.gz`.

I never ever reference `revision.txt`, which leads me to think that it is useless. Is it handled differently in other repositories?

Yep, I'll need to fix the repo name BTW :)

networkException:
So https://github.com/ungoogled-software/ungoogled-chromium-windows for example is using git submodules to directly import the patchset at a specific commit. To determent what chromium version is used the build scripts read `chromium_version.txt` and `revision.txt` from the "cloned" core repo

When using urls to tar archives this problem goes away of course as the version has to be specified manually

PR #1971, opened 2022-05-30
Add screenshots of the welcome page

networkException:
This pull request adds two screenshots of the browser of the welcome page to be
used by different platforms, for example as images in app stores.
See https://github.com/flathub/com.github.Eloston.UngoogledChromium/pull/59


networkException:
Even though the images contain version information I wouldn't update them unless there's a significant change to the ui to avoid increasing the repository size that much

PF4Public:
But that is essentially the same image as vanilla Chromium save for the "ungoogled-chromium" text O.o

networkException:
I guess it shows objective 2 :smile:

If you have a better aspect to show in such a screenshot I'll happily adjust

Ahrotahn:
Are there any requirements to show the version in the screenshot?  Maybe a good alternative would be `chrome://flags` since the look of that page hardly ever changes and it would still show off the light and dark themes well.

PF4Public:
> `chrome://flags` is a good idea I think

Maybe even filtered by `ungoogled-chromium`?

Ahrotahn (dismissed):
I thought I had approved this already, sorry @networkException!  
I see no problem with adding the screenshots.  We can always update them if they need to be changed.

networkException:
I quite like the idea of screenshotting the welcome page, maybe we'll just wait

PF4Public:
@networkException This PR is no longer blocked ;)

networkException:
oh right 😅

PF4Public:
Could be a nitpick, but still, perhaps make them narrower? Who needs all this empty space left and right…

networkException:
sure, how narrow though?

PF4Public:
Just narrow enough for everything to fit and have (subjectively :D) pleasant margins.
Something like this:
image

Ahrotahn (approved):
I guess it just depends on how you want the images to look on the page.  I'm fine with either version.  If you think we might ever someday include another screenshot then maybe we should make an image/screenshot subdirectory.

PR #1978, opened 2022-05-30
Add flags to hide SharingHub and SidePanel buttons.  Small cleanup.

Ahrotahn:
This PR adds two new patches that add flags to remove the sharing hub and the sidpanel buttons with `--disable-sharing-hub` and `--hide-sidepanel-button` respectively.  

I've also included a commit that reorders some entries in the flags doc and removes the non-breaking spaces from the feature flag section since they shouldn't cause line breaks due to not containing hyphens.  It also moves the bromite flags in the series file further up the list so new entries can simply be added to the end of the list.  Historically the order mattered, but since the flag headers patch was added it hasn't been necessary.


PF4Public:
@Ahrotahn How do you feel, are these changes safe enough for merging along with the SSl-thing and bumping revision to 3?

Ahrotahn:
It should be fine, but there's no rush on these changes.  I wouldn't let this hold up the other PR.

PF4Public (dismissed):
It works! They are gone!

PF4Public (approved):

networkException (approved):

PR #1991, opened 2022-06-10
Update to Chromium 102.0.5005.115

networkException:


Ahrotahn (dismissed):
The changes build and run fine for me!  Though your repo is missing the last few commits.  I don't think that it's a problem since there's no conflicts it should be able to be merged, but I'm not familiar enough with how Github would handle that.  

I've also noticed that quilt likes to order hunks differently on your system.  If you get some time, try setting `LC_ALL=C` before refreshing to see if that changes the ordering.  If so, then that might be something we could add to `set_quilt_vars.sh`.

PF4Public (approved):

PR #1993, opened 2022-06-17
Update to Chromium 103.0.5060.53

Ahrotahn:
This PR updates to Chromium 103.0.5060.53.  

Notable changes:

* A removal in 0005-disable-default-extensions.patch due to the removal of the feedback extension in [[3606948]](https://chromium-review.googlesource.com/c/chromium/src/+/3606948)

* PluginsResourceService and WebResourceService were removed in [[3584290]](https://chromium-review.googlesource.com/c/chromium/src/+/3584290), so 0007-disable-web-resource-service.patch is no longer needed and the relevant section in disable-untraceable-urls.patch has been removed.

* Small update to the trk patch due to [[3621976]](https://chromium-review.googlesource.com/c/chromium/src/+/3621976)

* A removal and some new additions to remove-unused-preferences-fields.patch due to [[3627578]](https://chromium-review.googlesource.com/c/chromium/src/+/3627578), [[3637732]](https://chromium-review.googlesource.com/c/chromium/src/+/3637732), and [[3595760]](https://chromium-review.googlesource.com/c/chromium/src/+/3595760)

* disable-formatting-in-omnibox.patch includes the new mobile prefix: [[3596512]](https://chromium-review.googlesource.com/c/chromium/src/+/3596512)

* android\_small/icudtl(\_extra).dat has been removed from update_lists.py due to [[3582190]](https://chromium-review.googlesource.com/c/chromium/deps/icu/+/3582190)

* 'false' has been changed to 'False' in remove-uneeded-ui.patch.  Apparently that's always been wrong, but has been interpreted correctly until [[3603267]](https://chromium-review.googlesource.com/c/chromium/src/+/3603267)

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/102.0.5005.115..103.0.5060.53?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2022/06/stable-channel-update-for-desktop_21.html)

Issue #1336, opened 2021-01-16
Add #extensions-toolbar-menu flag from older chromium versions


nhooyr:
So chromium used to allow you to disable the extensions icon in the toolbar.

See https://www.reddit.com/r/chrome/comments/jw322g/chrome_extension_icon_permanent_in_v87/

However this ability was recently removed. However, it works via the cli as so:

```sh
/Applications/Chromium.app/Contents/MacOS/Chromium --disable-features=ExtensionsToolbarMenu
```

I'd like to see it reintroduced in ungoogled as I prefer having a minimal toolbar.

The commit that removed it in chromium https://github.com/chromium/chromium/commit/09df13c25e1b30200b3d08531cc741be1df1ea49

nhooyr:
Also see https://bugs.chromium.org/p/chromium/issues/detail?id=1100412

AJolly:
I'd love to see this.  Chrome 91 just deprecated the cli  --disable-features=ExtensionsToolbarMenu flag, so I'm hunting for a solution.

PR #1994, opened 2022-06-19
Add flag to hide the extensions menu (the puzzle piece icon)

jwangac:
This pull request solves #1336.

The patch has been tested only on Windows. An automatic build by Github Actions can be found at
https://github.com/jwangac/ungoogled-chromium-windows/releases/tag/test

This pull request might be postponed after updating Chromium to 103.0.5060.53 (by  #1993).

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -36,6 +36,7 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
   `--custom-ntp` | Allows setting a custom URL for the new tab page.  Value can be internal (e.g. `about:blank`), external (e.g. `example.com`), or local (e.g. `file:///tmp/startpage.html`).  This applies for incognito windows as well when not set to a `chrome://` internal page.
   `--disable-sharing-hub` | Disables the sharing hub button.
   `--hide-sidepanel-button` | Hides the SidePanel Button.
+  `--hide-extensions-menu` | Hides the extensions menu (the puzzle piece icon) that allows the user to control extensions site access.
   `--hide-tab-close-buttons` | Hides the close buttons on tabs.
   `--remove-grab-handle` | Removes the reserved empty space in the tabstrip for moving the window.
   `--remove-tabsearch-button` | Removes the tabsearch button from the tabstrip.
--- a/patches/extra/ungoogled-chromium/add-flag-to-hide-extensions-menu.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-hide-extensions-menu.patch
@@ -1,0 +1,21 @@
+--- a/chrome/browser/ui/views/toolbar/toolbar_view.cc
++++ b/chrome/browser/ui/views/toolbar/toolbar_view.cc
+@@ -255,6 +255,7 @@ void ToolbarView::Init() {
+   // Do not create the extensions or browser actions container if it is a guest
+   // profile (only regular and incognito profiles host extensions).
+   if (!browser_->profile()->IsGuestSession()) {
++    if (!base::CommandLine::ForCurrentProcess()->HasSwitch("hide-extensions-menu"))
+     extensions_container =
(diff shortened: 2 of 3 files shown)


PF4Public:
Did you on purpose not reuse already existing "if" condition?

Ahrotahn (approved):
This looks good to me and works on my end.  If it gets another approval before 103 goes stable I can add a new commit to that PR to refresh the patch, otherwise I'll reapprove after this is updated for 103.  

> Did you on purpose not reuse already existing "if" condition?

I've done this in the past to reduce the surface-area of the patch, though in this case it shouldn't really matter which approach is used.

PF4Public (approved):

PF4Public:
@jwangac No worries, both approaches are fine, just wanted to know if there was any particular reason.

Ready to merge, I guess?

PF4Public:
Merged, thanks @jwangac

Rusenche:
When installing "crx" v.103.0.5060.53 crashes. Tried with several random "crx".

The other problem is if you add a random unpacked extension and then try to package it in order to create "crx" - you get an error and it is impossible to create "crx".

Rusenche:
When installing "crx" v.103.0.5060.53 crashes. Tried with several random "crx".

The other problem is if you add a random unpacked extension and then try to package it in order to create "crx" - you get an error and it is impossible to create "crx".

And let's not forget - all this is provided that chrome://flags/#hide-extensions-menu -> Enabled.

macchrome:
There is no need for this flag.

jwangac:
This PR causes Chromium to crash when installing extensions, if
the new "hide-extensions-menu" flag is enabled.

To reproduce the crash:
1. Enable the "hide-extensions-menu" flag.
2. Install an extension.
3. Chromium exits immediately.

Chromium crashes because the patch leaves the pointer `extensions_container`
uninitialized. The pointer seems to be used somewhere in Chromium without check
for null.

To prevent the crash, I try to create an extensions menu with auto-hide display
mode, if the user enable the "hide-extensions-menu" flag:
```patch
--- a/chrome/browser/ui/views/toolbar/toolbar_view.cc
+++ b/chrome/browser/ui/views/toolbar/toolbar_view.cc
@@ -255,8 +255,12 @@ void ToolbarView::Init() {
   // Do not create the extensions or browser actions container if it is a guest
   // profile (only regular and incognito profiles host extensions).
   if (!browser_->profile()->IsGuestSession()) {
-    extensions_container =
-        std::make_unique<ExtensionsToolbarContainer>(browser_);
+    if (!base::CommandLine::ForCurrentProcess()->HasSwitch("hide-extensions-menu"))
+      extensions_container =
+          std::make_unique<ExtensionsToolbarContainer>(browser_);
+    else
+      extensions_container =
+          std::make_unique<ExtensionsToolbarContainer>(browser_, ExtensionsToolbarContainer::DisplayMode::kAutoHide);
   }
   std::unique_ptr<media_router::CastToolbarButton> cast;
   if (media_router::MediaRouterEnabled(browser_->profile()))
```
`ExtensionsToolbarContainer::DisplayMode::kAutoHide` is a mode maintained by
chromium developers for web apps. However, it will not completely hide the
extensions menu. The visibility is determined by the logic in [chromium/chromium](https://github.com/chromium/chromium/blob/77818e5f1df0d268d87f9645d91a751c86b8d97c/chrome/browser/ui/views/extensions/extensions_toolbar_container.cc#L784).

So sometimes the extensions menu will show, particularly when installing a new
extension. But the menu will be hidden at most times.

Sorry for my mistake. I have updated [the patch](https://github.com/jwangac/ungoogled-chromium/commit/c953b0cffcb9e9f60afb5b550da3d0146685a065). But it is still being tested.

PF4Public:
@jwangac Feel free to open another PR when you're ready. This PR will be reverted.

PR #2006, opened 2022-07-01
Update actions

wrt54g:
I updated the github actions

### DIFF
--- a/.github/workflows/new_version_check.yml
+++ b/.github/workflows/new_version_check.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Get the latest Linux version
         id: latest-version
         run: echo "::set-output name=version::$( curl -s https://omahaproxy.appspot.com/linux )"
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


networkException:
Is there any benefit to this change?

wrt54g:
@networkException They are dependency updates. Check the releases page for these actions to view the changes.

networkException:
I know what the change is, I want to know if there's any benefit in upgrading from one stable branch to another. If there is none there's no point in making this change.

wrt54g:
Just later versions:
https://github.com/actions/stale/releases/tag/v5.0.0
https://github.com/actions/checkout/releases/tag/v3.0.2

wrt54g:
They now use nodejs v16 (or the stale action)

PF4Public:
It would have been great if you tested these changes in you fork @wrt54g for some time (a month for example) to be sure it won't bring any surprises. That's what I was about to suggest, but since you've already closed this PR, well, whatever.

networkException:
I don't really see the point of switching stable branch for the sake of updating :shrug:

PR #2014, opened 2022-07-12
Update privacy sandbox and custom ntp patches

Ahrotahn:
This PR updates the privacy sandbox patch to disable the Conversion Measurement and Aggregation Service features used for attribution reporting.  This caused a problem on windows builds that made the process hang after closing:  https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/101

I've added a new domain substitution regex which replaces the privacysandbox.com and privacysandboxservices.com addresses.  

This also updates the custom ntp patch for a change in 102 that caused the flag to no longer function in incognito.


teeminus:
Do you want to bump the revision as well?

PF4Public (approved):
Releasing as a revision bump seems reasonable enough.

networkException (approved):
Builds and runs fine

Were these features "in violation" of objective one? From some quick searches it seems like both the aggregation service and conversation measurement are reporting to some servers (or is this something site defined, I guess it would fall under objective three in that case)

Ahrotahn:
In it's current iteration this is mainly for the first one, but certainly hits on the third as well.  

I probably should have gone into a little more detail in the original post.  The reason for the new domain substitution is because it attempts to connect to `publickeyservice.aws.privacysandboxservices.com`.  
The code is being set up to handle an array or URLs, so there will be more we'll have to deal with in the future.  They have a bug report open to change the existing one/add more here:  [1295705](https://bugs.chromium.org/p/chromium/issues/detail?id=1295705)

Hopefully those will also exist on a `privacysandbox*.com` domain so we don't have to make further changes.

PF4Public:
So, are we ready to merge and tag?

PF4Public:
2 commits vs 1 commit — no big difference in my eyes :)

PR #2022, opened 2022-07-22
Per-platform version check

PF4Public:
I was thinking of something like this in https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/158#issuecomment-1172274425

Issues should look like this: https://github.com/PF4Public/ungoogled-chromium/issues/22

Android was omitted as currently not very active.

TODO-list before merging:

- [x] Change the if condition 
- [x] Squash everything

### DIFF
--- a/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
+++ b/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
@@ -1,13 +1,13 @@
 ---
 name: Create an "Updating to Chromium x.x.x.x"
 about: For letting the community track progress to a new stable Chromium
-title: Updating to Chromium {{ env.VERSION }}
+title: Updating to Chromium {{ env.VERSION }} on {{ env.PLATFORM }}
 labels: update
 assignees: ''
 
 ---
 
-Chromium stable channel for Linux has been updated to a newer version: {{ env.VERSION }}.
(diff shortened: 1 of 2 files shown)


PF4Public:
Oh by the way, how should be the Mac OS written? Is it "macOS" nowadays?

networkException:
yep "macOS" since 2016

PF4Public:
Would it be OK to create either a single "all platforms issue" or three issues each for its own platform? We still can link all of them to a single PR and close simultaneously.
Look [here](https://github.com/PF4Public/ungoogled-chromium/issues) for all of them in one place.

PF4Public:
> I think trying to further group Windows and macOS in this example would just be unnecessarily complicated, especially without a proper scripting language

My thought exactly :)

I'll prepare this draft for merging then. After we merge the version-bump PR of course.

networkException (approved):

Ahrotahn (approved):

teeminus:
How do we handle version differences between eg. windows and linux? The core repo uses the linux version and the patches are updated for that. What if we need to modify core patches for a more recent windows release?

PF4Public:
I thought there was a suggestion to have tags for all versions in this repo and then use respective ones in downstream repos with there-needed patches.

networkException:
Yes we should probably do that even without any changes to patches or lists, avoids confusion and keeps things consistent

PR #2039, opened 2022-08-06
Remove Android Build in README

georgewoodall82:
The ungoogled chromium android build that is pointed to is archived and the f-droid repo is unavailable, It should be removed until a better option is available for android.


networkException (approved):
its sad to have it removed from the list but advertising non existant releases also isn't great

PF4Public:
Excuse me, but what do you mean by "unavailable"?
S20806-235508

Instead of removing, I'd rather mention that it has older binaries.

georgewoodall82:
my bad i didn't properly check that

PR #2051, opened 2022-08-17
Update to Chromium 104.0.5112.102

Ahrotahn:
This PR updates to Chromium 104.0.5112.102.  
There weren't any notable changes on our end.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/104.0.5112.81..104.0.5112.102?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2022/08/stable-channel-update-for-desktop_16.html)


PF4Public (approved):

networkException:
Should we also have a separate commit for .101? Or just two tags for the same commit?

See https://github.com/ungoogled-software/ungoogled-chromium/pull/2022#issuecomment-1207209378

PF4Public:
> Should we also have a separate commit for .101?

That's a good question actually. I've made it possible for me to use either a tag or a plain commit for Gentoo, so for me it does not matter, but I believe it is different for other palatform repos, i.e. they accept only a tag.

networkException:
hmm windows for example relies on `chromium_version.txt` so that would always need its own commit

I guess my suggestion to keep everything consistent from now on would be to:

- Have a commit for every Linux, Windows or macOS stable version (with `chromium_version.txt` and patches updated accordingly, usually only a few lines of changes)
- Have a tag and release for every Linux, Windows or macOS stable version

Ahrotahn:
I've separated the point releases into their own commits.  We can go with whatever method works best for the platform repos.  
Historically we'd just use the latest version since the tarball jobs create them in opposite order, .102 is made first then .101 after.  This was a problem when it took half a day for each version to be made available, but with the recent updates to that process that delay is reduced significantly.

networkException (approved):
Great thank you!

PF4Public (approved):

PR #2062, opened 2022-08-26
Update to Chromium 105.0.5195.52

Ahrotahn:
This PR updates to Chromium 105.0.5195.52.

Notes:

* Most of the notable changes happened in the safebrowsing patch due to [[3689411]](https://chromium-review.googlesource.com/c/chromium/src/+/3689411), [[3652503]](https://chromium-review.googlesource.com/c/chromium/src/+/3652503), and [[3693929]](https://chromium-review.googlesource.com/c/chromium/src/+/3693929).
* A new pruning exclusion was added for a new perfetto file
* Management of wayland dependencies changed in [[3695277]](https://chromium-review.googlesource.com/c/chromium/src/+/3695277) and angle incorrectly mixes the supplied headers with system ones.  The supplied version is at 1.18, so if your system is using a newer version you might encounter undeclared identifier errors when building.  This can be worked around by applying the changes in [[3727670]](https://chromium-review.googlesource.com/c/angle/angle/+/3727670) or building with `angle_use_wayland=false`.

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/104.0.5112.102..105.0.5195.52?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2022/08/stable-channel-update-for-desktop_30.html)


PF4Public:
Builds and runs 👍🏻 
Is perfetto harmless enough so that we do not bother removing it?

Ahrotahn:
Yeah, it's a performance tracing library.  It's possible to set `enable_base_tracing` to false to disable that, but it still uses code from the library that would require a patch to deal with the missing file since it's used to generate a header it needs.

networkException (approved):

PF4Public (approved):

PF4Public:
Should we exclude `goo.gle` maybe later? #2063

networkException:
I don't mind either way, although merging later is less work as we don't have to update existing commits on this branch due to updated lists

PF4Public:
> merging later is less work

Absolutely!

PR #2065, opened 2022-08-30
CI: Notify @ungoogled-software/notify-* platform teams for new versions

networkException:
This pull request makes the github actions bot mention every relevant platform team if a new version gets released for a given platform. This does not reuse existing teams so that maintainance discussions and notifications are separate opt ins.

### DIFF
--- a/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
+++ b/.github/ISSUE_TEMPLATE/create-an--updating-to-chromium-x-x-x-x-.md
@@ -14,3 +14,5 @@ If you are willing to work on updating the patches and lists, please leave a com
 If you'd like to increase visibility of your progress or get early feedback/advice, consider creating a [Draft Pull Request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests). Finally, make sure to reference this issue in your PR. Please make sure to read [/docs/developing.md](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/developing.md#updating-patches) for guidance.
 
 Feel free to raise issues or questions throughout the process here. However, please refrain from asking for ETAs unless no visible progress has been made here or in the developer's PR for a while (e.g. 2 weeks).
+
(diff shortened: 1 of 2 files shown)


PF4Public:
Are we about to invite everyone into @ungoogled-software/notify-* teams? Anyone interested might simply subscribe on notifications on new releases or might anyone not?

networkException:
I'd have said anyone who wants to get notified for their platform specifically can just ask to get added.

Compared to releases this would be faster, they don't have to subscribe to *all* events on the project and they know which version is relevant to their platform.

PF4Public:
Your points are valid, but I have a suspicion that it is not how teams are supposed to be used 🤷

networkException:
Do you have a better idea?

PF4Public:
> Do you have a better idea?

Mention them directly via workflow maybe.

Ahrotahn:
It should work, but I'm not familiar enough with github to know if better options are available.  
This would notify only when the new version comes out and not when we tag new versions right?

networkException:
yep

Ahrotahn (approved):
Ok, seems good to me!

networkException:
@ungoogled-software/windows @ungoogled-software/macos let me know if you want to be added to the notify teams (I don't think there's a way to make them open for joining)

Nifury:
I've turned on notifications for all activities, so I guess it's not necessary to join the teams?

networkException:
In that case you don't need to join, yes

PF4Public (approved):
We may try, if you insist

networkException:
So as it turns out the github-actions bot doesn't have permission to mention teams. I still think this is a valuable to notify downstream maintainers of chromium releases and the updating process here without having to watch everything

Do you think its fine to just include everyone who wants to get notified manually? Its a bit ugly but the alternatives (like a separate account who submits those issues) are more complicated and not great either

PF4Public:
> Do you think its fine to just include everyone who wants to get notified manually?

That was in essence my suggestion, so I'm _for_ it :)

Like @Nifury says, most are already watching this repository anyway.

PR #2080, opened 2022-09-06
Add flag for disabling link and text dragging

PF4Public:
Implementation of https://github.com/ungoogled-software/ungoogled-chromium/discussions/2055#discussioncomment-3494020 through feature flag thanks to help from @Ahrotahn and @networkException.

This allows selecting text from within a link and also allows starting the selection without clearing the existing one.
[This is a long link, that I'll use for demonstration of what it is. They say a <s>picture</s> video is worth a thousand words](https://github.com/ungoogled-software/ungoogled-chromium/).
And here is some more text for demonstration purposes only :)

With the flag enabled following becomes possible:

https://user-images.githubusercontent.com/10319700/188681263-9bc37892-a980-4cf9-ac66-7f199f43a325.mp4

PS: I also reordered `--force-punycode-hostnames` to be in alphabetical order.

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-for-disabling-link-drag.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-for-disabling-link-drag.patch
@@ -1,0 +1,67 @@
+--- a/third_party/blink/common/features.cc
++++ b/third_party/blink/common/features.cc
+@@ -14,6 +14,8 @@
+ 
+ namespace blink {
+ namespace features {
++
++const base::Feature kDisableLinkDrag{"DisableLinkDrag", base::FEATURE_DISABLED_BY_DEFAULT};
+ 
+ // Apply lazy-loading to ad frames which have embeds likely impacting Core Web
+ // Vitals.
+--- a/third_party/blink/public/common/features.h
++++ b/third_party/blink/public/common/features.h
+@@ -16,6 +16,8 @@
+ 
+ namespace blink {
+ namespace features {
++
++BLINK_COMMON_EXPORT extern const base::Feature kDisableLinkDrag;
+ 
+ BLINK_COMMON_EXPORT extern const base::Feature kAutomaticLazyFrameLoadingToAds;
+ BLINK_COMMON_EXPORT extern const base::Feature
+--- a/third_party/blink/renderer/core/input/mouse_event_manager.cc
++++ b/third_party/blink/renderer/core/input/mouse_event_manager.cc
+@@ -687,8 +687,14 @@
+ 
+   bool single_click = event.Event().click_count <= 1;
+ 
++if (base::FeatureList::IsEnabled(features::kDisableLinkDrag)){
++  mouse_down_may_start_drag_ = single_click && !IsSelectionOverLink(event) &&
++                               !IsExtendingSelection(event) &&
++			        !event.GetHitTestResult().IsSelected(event.GetHitTestLocation());
++}else{
+   mouse_down_may_start_drag_ = single_click && !IsSelectionOverLink(event) &&
+                                !IsExtendingSelection(event);
++}
+ 
+   mouse_down_ = event.Event();
+ 
+--- a/third_party/blink/renderer/core/editing/selection_controller.cc
++++ b/third_party/blink/renderer/core/editing/selection_controller.cc
+@@ -1319,9 +1319,13 @@
+ }
+ 
+ bool IsSelectionOverLink(const MouseEventWithHitTestResults& event) {
++if (base::FeatureList::IsEnabled(features::kDisableLinkDrag)){
++  return event.IsOverLink();
++}else{
+   return (event.Event().GetModifiers() & WebInputEvent::Modifiers::kAltKey) !=
+              0 &&
+          event.IsOverLink();
++}
+ }
+ 
+ bool IsUserNodeDraggable(const MouseEventWithHitTestResults& event) {
+--- a/chrome/browser/ungoogled_flag_entries.h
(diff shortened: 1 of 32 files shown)


patches/extra/ungoogled-chromium/add-flag-for-disabling-link-drag.patch
```diff
@@ -0,0 +1,67 @@
...
+ bool IsUserNodeDraggable(const MouseEventWithHitTestResults& event) {
+--- a/chrome/browser/ungoogled_flag_entries.h
++++ b/chrome/browser/ungoogled_flag_entries.h
+@@ -108,4 +108,8 @@
+      "Hide SidePanel Button",
+      "Hides the SidePanel Button.  ungoogled-chromium flag.",
+      kOsDesktop, SINGLE_VALUE_TYPE("hide-sidepanel-button")},
++    {"disable-link-drag",
++     "Disable link drag",
++     "Prevents dragging of links and selected text.  ungoogled-chromium flag.",
```
networkException, line 64: I know its everywhere but is there an actual reason to have two spaces before "ungoogled-chromium flag."?

PF4Public: Hmm, good catch! I didn't even notice that! Perfectionist in me would be definitely willing to replace all double-spaces unless there is a reason for them to be there.

networkException: >Everywhere: Fix formatting issue

*breaks production*

PF4Public: I guess, it would be safe to remove all double-spaces since they are not rendered anyways:
image

tomasz1986: The double space thing comes from typewriters. Some people/organisations still recommend writing that way. There're many articles on this matter if you do a search for "double space after period" (e.g. https://smithsonianmag.com/smart-news/microsoft-word-now-considers-two-spaces-after-period-error-180974754).

PS You've missed a few double spaces in between the actual sentences, so you may want to have another look if you do intend to replace all of them 😉.

PF4Public: Indeed some of the double-spaces come from Chromium source itself. For example [here](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/prefetch/prefetch_prefs.h;l=31-38?q=%22%2F%2F+Enum+representing+possible+values+of+the+Preload+Pages+opt-in+state.++Since%22&ss=chromium). Still it is funny, how inconsistent it is: there are two spaces after the first sentence, but single spaces further down along this comment.

> PS You've missed a few double spaces in between the actual sentences, so you may want to have another look if you do intend to replace all of them .

Behold my wrath against multiple-spaces! :D Honestly saying I hate repetitive spaces because that reminds me of the times I had to work with documents whose formatting was done solely by spaces! Wild hatred!
BTW, @tomasz1986 it would be great if you could test this PR.

tomasz1986: > Still it is funny, how inconsistent it is: there are two spaces after the first sentence, but single spaces further down along this comment.

It could be due to different people making changes to the text 😉.

> BTW, @tomasz1986 it would be great if you could test this PR.

I'd love to, but I don't really compile the browser myself 🙁. I've given it a try multiple times, but my hardware isn't really sufficient for it yet.

Ahrotahn: I was taught an oldschool method of typing growing up so I still have have a habit of adding double-spaces after ending sentences.  I probably added that when doing a flag and it got copy/pasted it over to others over time.  I guess those could be considered age markers 😛.

PF4Public: So in the end I hope no one would be against me removing them, right?

jstkdng: are you satoshi nakamoto?

Ahrotahn:
I tested the changes (before the dedoublespacing) and it worked as expected.  The only thing I noticed is that you still have the commandline flag in the docs instead of adding `DisableLinkDrag` in the feature flag section.

patches/extra/ungoogled-chromium/add-flag-for-disabling-link-drag.patch
Ahrotahn: > So in the end I hope no one would be against me removing them, right?

That should be fine!

PF4Public:
> The only thing I noticed is that you still have the commandline flag in the docs instead of adding `DisableLinkDrag` in the feature flag section.

Good point! I'll force-push this whole thing once again and start building again to be sure nothing broke.

Ahrotahn (approved):
That should be fine, having them in separate commits is more important IMO.  

I've tested the new changes and everything looks good to me!

PF4Public:
Ready to merge?

networkException (approved):

PF4Public:
So after some time having this flag enabled it seems to me that some of my clicks get ignored. This never happens without this flag. 

My suspect is that I have a habit of not stopping my mouse specifically for a click, rather continue it to where I need it next and it is probable that on a hidpi screen this amounts to a "selection mode" for a couple of pixels between mouse DOWN and UP, which makes the click event consumed in "selection". Maybe this could be fixed by adding a condition on selection length or click time. 🤔

PR #2088, opened 2022-09-15
feat: add --enable-persistent-cookies switch

virchau13:
Since cookies default to session-only by default, when starting ungoogled-chromium in kiosk-mode it's impossible to persist cookie state without a lot of hassle with settings files. This PR adds a switch, `--enable-persistent-cookies`, that simply makes cookies persist across sessions.

(For context, I'm trying to display a web app in Chromium kiosk-mode. The problem is that the web app has logins, which due to the cookie setting don't get persisted across multiple sessions, meaning every time the Chromium instance is launched the web app has to be logged into again. This forces me to use normal Chromium, but it would be great if I could use ungoogled-chromium for this usecase.)

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -14,6 +14,7 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
   `--disable-beforeunload` | Disables JavaScript dialog boxes triggered by `beforeunload`
   `--disable-grease-tls` | Disables GREASE for TLS. Combined with `--http-accept-header` allows browser to look more like a tor-browser. See https://github.com/ungoogled-software/ungoogled-chromium/issues/783 for more details.
   `--disable-search-engine-collection` | Disable automatic search engine scraping from webpages.
+  `--enable-persistent-cookies` | Enables cookies to be persisted between sessions (by default cookies are session-only.)
   `--extension-mime-request-handling` | Change how extension MIME types (CRX and user scripts) are handled. Acceptable values are `download-as-regular-file` or `always-prompt-for-install`. Leave unset to use normal behavior.
   `--fingerprinting-canvas-image-data-noise` | (Added flag to Bromite feature) Implements fingerprinting deception for Canvas image data retrieved via JS APIs. In the data, at most 10 pixels are slightly modified.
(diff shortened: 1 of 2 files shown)


PF4Public:
Have you tried changing cookie settings to not remove them upon browser restart as one would with a normal Chromium profile?

virchau13:
> Have you tried changing cookie settings to not remove them upon browser restart as one would with a normal Chromium profile?

In order to get this to work with kiosk-mode, the following steps have to be taken:

1. Run Chromium normally with user-data-dir set to some directory
2. Manually(!) navigate to `chrome://settings/cookies` and untick the "Clear cookies and site data when you close all windows" checkbox
3. Close Chromium
4. Every time you want to run kiosk-mode, set user-data-dir to the same directory

The problem with this is that
1. It involves at least 1 manual step (can't be automated). This can technically be circumvented by `--disable-machine-id` and copying over the user-data-dir, but that requires full storage of the user-data-dir even when only 1 setting needs to be changed, and significantly complicates automatic setup
2. It's not reproducible for the same reason

> Have you tried just to omit this iridium patch when building?

This is for an open source project, so I'm trying to make it work with vendored ungoogled-chromium so that the user doesn't have to build all of Chromium.

(Also I just tested this and it doesn't work, so I'm marking it as draft until I fix it)

virchau13:
Okay, I found a better way to do this, so I'm closing this PR.

For anyone who stumbles across this, use the following script to launch Chromium in order to enable persistent cookies with ungoogled-chromium automatically:
```bash
# set this variable accordingly
C_USER_DATA_DIR=~/.cache/chromium-test

if [ ! -d "$C_USER_DATA_DIR/Default" ]; then
    mkdir -p "$C_USER_DATA_DIR"/Default
    echo '{ "profile": { "default_content_setting_values": { "cookies": 1 } } }' > "$C_USER_DATA_DIR"/Default/Preferences
fi

chromium --user-data-dir="$C_USER_DATA_DIR" --kiosk "$@"
```

It turns out that if you just put `{ "profile": { "default_content_setting_values": { "cookies": 1 } } }` into the `Default/Preferences` file in the user-data-dir, Chromium will pick it up when setting up the user-data-dir and make sure that setting is preserved.

PR #2125, opened 2022-10-25
Fix set-output and node.js warnings

PF4Public:


networkException (approved):
LGTM, I really gotta do that for the arch workflows as well

PF4Public:
It seems these deprecations are happening in every repo one way or another :)

Ahrotahn (approved):

PR #2169, opened 2022-12-03
Adapt some  bromite antifingerprint patches )webaudio + plugins +  UA= 

anomalocaris452:
Could Ungoogled import  Bromite patches? .. as Ungoogled adapted Bromite's clientrect + canvas patches

AudioBuffer-AnalyserNode-fp-mitigations.patch  Works against webaudio  
Disable-UA-full-version.patch  Might help  to spoof   the UserAgent
Disable-plugins-enumeration.patch Will hide your plugins  as Bromite

i've imported it by myself


PF4Public:
Have you attempted building ungoogled-chromium with these patches applied?

anomalocaris452:
@PF4Public YES it builds

https://build.opensuse.org/package/show/home:Area51Kacz:branches:home:ungoogled_chromium:testing/ungoogled-chromium22

alphabug:
Supposing these patches bring value, they should be gated behind flags.

AudioBuffer... could be useful to those who wish to harden their browser.
Disable-UA-full-version... isn't this already the standard behavior? Where is it still leaking the minor version?
Disable-plugins-enum... plugins are no longer a thing?
Clamp-time-resolution... the patch links a chromium bug that has been closed as fixed (in 109).

Issue #2109, opened 2022-10-09
Disable popup "Press F11 to exit fullscreen"


Devocub:
### Description

Disable popup "Press F11 to exit fullscreen"

### Who's implementing?

- [ ] I'm willing to implement this feature myself

### The problem

![image](https://user-images.githubusercontent.com/2735824/194773448-5a912e80-793f-4a8c-bf91-c799d45ac66c.png)

### Possible solutions

Add flag to disable the popup.

### Additional context

I was in fullscreen probably 5000+ times, I know how to exit.

PR #2176, opened 2022-12-05
Add flag for hiding the fullscreen exit UI

implicitfield:
This adds a flag that hides the "X" that appears when the mouse cursor is moved towards the top of the window in fullscreen mode. This is very similar to the now removed `enable-experimental-fullscreen-exit-ui` flag.


Ahrotahn:
`kOsWindows` would need to be changed to `kOsWin`, or alternatively `kOsDesktop` could be used even though mac builds would return false before then.  

Other than that, this looks good and worked as expected.  

If you want to hit two birds with one stone, #2109 could also be resolved with this PR by adding:
```diff
--- a/chrome/browser/ui/views/frame/browser_view.cc
+++ b/chrome/browser/ui/views/frame/browser_view.cc
@@ -1828,6 +1828,10 @@
     ExclusiveAccessBubbleHideCallback bubble_first_hide_callback,
     bool notify_download,
     bool force_update) {
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          "hide-fullscreen-exit-ui"))
+    return;
+
   DCHECK(!notify_download || exclusive_access_bubble_);
   // Trusted pinned mode does not allow to escape. So do not show the bubble.
   bool is_trusted_pinned =
```

Ahrotahn (approved):
Looks good to me!

PF4Public (approved):

PF4Public:
Thanks! Merged.

PR #2005, opened 2022-06-29
Re-add flag to hide the extensions menu (the puzzle piece icon)

jwangac:
This PR replaces my previous PR #1994. The previous PR is reverted because it
causes ungoogled-chromium to crash.

This PR re-adds the flag that allows one to hide the extensions menu (the
puzzle piece icon), by patching [the menu's visibility logic](https://github.com/chromium/chromium/blob/0f4ff69f75ce13ac45815a53000c36e7401561a5/chrome/browser/ui/views/extensions/extensions_toolbar_container.cc#L804):

```patch
 bool ExtensionsToolbarContainer::ShouldContainerBeVisible() const {
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch("hide-extensions-menu"))
+    return false;
+
   // The container (and extensions-menu button) should not be visible if we have
   // no extensions.
   if (!HasAnyExtensions())
     return false;

   // All other display modes are constantly visible.
   if (display_mode_ != DisplayMode::kAutoHide)
     return true;

   if (GetAnimatingLayoutManager()->is_animating())
     return true;

   // Is menu showing.
   if (GetExtensionsButton()->GetExtensionsMenuShowing())
     return true;

   // Is extension pop out is showing.
   if (popped_out_action_)
     return true;

   // Is extension pop up showing.
   if (popup_owner_)
     return true;

   return false;
 }
```

I test the patch on [Windows](https://github.com/jwangac/ungoogled-chromium-windows/releases/tag/103.0.5060.53-test). The patch works and avoids previous crashes

1. when installing a packed extension
2. when loading an unpacked extension
3. when packing an extension

(Thanks Rusenche for reporting these crash cases.)

But my test can be incomplete as I'm not a heavy extension user or developer.
This PR is marked as draft. I'm wondering if someone can help testing it in 
more use-cases.

This patch has some limitations.

1. The flag will hide all pinned extensions. One cannot hide only the puzzle
piece icon but keep pinned extensions.

2. Notification like `[Extension Name] has been added to Chromium` (shows after
installing an extension) and `Remove "[Extension Name]"?` (shows when removing
an extension) cannot find the hidden extension icon to attach. The notification
will show at the left-top corner.

PR #2198, opened 2022-12-28
Update documentation for hide-extensions-menu and hide-fullscreen-exit-ui flags

Ahrotahn:
This PR updates the documentation for the hide-extensions-menu and hide-fullscreen-exit-ui flags.  

There was some discussion in #2005 that the verbiage could be clarified, so that has been updated to be more clear.  
In #2176 the flag gained extra functionality.  Its description was updated but docs/flags.md didn't include the new change.

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -38,7 +38,7 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
   `--disable-sharing-hub` | Disables the sharing hub button.
   `--enable-incognito-themes` | Allows themes to change the appearance of Incognito windows.
   `--hide-extensions-menu` | Hides the extensions container. This includes the puzzle piece icon as well as any pinned extensions.
-  `--hide-fullscreen-exit-ui` | Hides the "X" that appears when the mouse cursor is moved towards the top of the window in fullscreen mode.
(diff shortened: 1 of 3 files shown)


PF4Public:
Why does it fail validation? Could be some line numbers shifted?

Ahrotahn:
I forgot to refresh the patches since I was thinking it was just a document change.  The context lines in the other patch needed to be updated due to that change.

networkException (approved):
lgtm

PF4Public:
You [all] see! That's the problem I was referring to when I was suggesting to reorganise our patches to be somewhat more separated from one another. Could we maybe store them as a plain .h file, that would be copied to chromium sources instead of patches overlaying one another?

Ahrotahn:
There might be better ways to handle them, but this is much better than when everything was placed in `about_flags.cc`.  Nearly all of the other patches needed to be updated any time an individual patch was added/updated.  
With the current setup only the context lines for the next patch need to be updated.

PF4Public:
> when everything was placed in `about_flags.cc`

I meant having only .h as a plain file, while others are patches the way they are now.

PF4Public (approved):

PF4Public:
Or better yet:
Instead of patching-in, create a folder with the files we add, which should be copied into `third-party/ungoogled` before building. There should be `ungoogled_flag_entries.h` as a plain file which `#include`'s each `feature-flag.h` snipplet also provided by plain files, which are to be copied into `third-party/ungoogled`. Is it too complicated?

PR #2214, opened 2023-01-18
(Partially?) Fix building for macOS

implicitfield:
These changes fix two build failures that currently occur when building for macOS.
I also had to revert https://github.com/chromium/chromium/commit/cb5ad37, not sure whether that commit works upstream or not.

### DIFF
--- a/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
+++ b/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
@@ -1972,6 +1972,17 @@
  }
  
  std::unique_ptr<content::WebContentsViewDelegate> CreateWebContentsViewDelegate(
+--- a/chrome/browser/ui/views/tab_contents/chrome_web_contents_view_delegate_views_mac.mm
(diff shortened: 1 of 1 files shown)


kramred (approved):
Thanks, good work. 
I probably won't have time to act as a maintainer for the macOS repo any more, would you possibly be up to take that role?

PF4Public:
@implicitfield Will these changes somehow affect other platforms?

kramred:
Alternatively these changes could be migrated to the macOS repo.

PF4Public:
> I have no problem with moving these patches to the macOS repo if that is a more appropriate place for them.

I'd like to hear from @Ahrotahn about that.

Ahrotahn (approved):
This looks good to me!  The changes built and ran without issue on my end.  

I think it's fine to have these changes in this repo.  I'm surprised the problem fixed in the first commit hadn't caused any issues before now, and the second has more to do with safebrowsing than it does with platform-specific building for mac.

PF4Public (approved):

PF4Public:
Thanks!

PR #2225, opened 2023-01-26
Get rid of "mailto:" when copying email address

### LINKED ISSUE #2224: Get rid of "mailto:" when copying email address
### Description

Please see below

### Who's implementing?

- [X] I'm willing to implement this feature myself

### The problem

When copying an email address (link), the "mailto:" scheme is also copied. It's frustrating to paste and remove the scheme every time I send an email. 

### Possible solutions

Do not change this variable 
https://github.com/ungoogled-software/ungoogled-chromium/blob/ec75c03e2555ab060abb534a9b0d34d39aabdbe9/patches/extra/ungoogled-chromium/disable-formatting-in-omnibox.patch#L16

so that the mailto scheme can be removed by `RenderViewContextMenu::FormatURLForClipboard`. 
https://github.com/chromium/chromium/blob/cec6fdae2e86af277796f89b801acc3a30698b67/chrome/browser/renderer_context_menu/render_view_context_menu.cc#L921-L935

Since the mailto link doesn't actually open a webpage, I think it's perfectly fine even if the mailto scheme is hidden in the address bar. 

### Alternatives

_No response_

### Additional context

_No response_

Nifury:
When copying an email address in vanilla Chromium, the `mailto:` prefix is removed and only the email address is copied. However, `disable-formatting-in-omnibox.patch` disables `kFormatUrlOmitMailToScheme`, so the email link won't be formatted when copied to the clipboard, and the copied email address `someone@example.com` becomes `mailto:someone@example.com`. 

This PR re-enables `kFormatUrlOmitMailToScheme` field. 

Fixes #2224

### DIFF
--- a/patches/extra/ungoogled-chromium/disable-formatting-in-omnibox.patch
+++ b/patches/extra/ungoogled-chromium/disable-formatting-in-omnibox.patch
@@ -1,4 +1,6 @@
 # Disables omission of URL elements in Omnibox
+# except for the mailto scheme so that email links can be properly 
+# formatted when copied to clipboard (#2225). 
 
 --- a/components/url_formatter/url_formatter.cc
 +++ b/components/url_formatter/url_formatter.cc
@@ -13,8 +15,6 @@
 -const FormatUrlType kFormatUrlOmitTrivialSubdomains = 1 << 5;
 -const FormatUrlType kFormatUrlTrimAfterHost = 1 << 6;
 -const FormatUrlType kFormatUrlOmitFileScheme = 1 << 7;
--const FormatUrlType kFormatUrlOmitMailToScheme = 1 << 8;
--const FormatUrlType kFormatUrlOmitMobilePrefix = 1 << 9;
 +const FormatUrlType kFormatUrlOmitUsernamePassword = 0 << 0;
 +const FormatUrlType kFormatUrlOmitHTTP = 0 << 1;
 +const FormatUrlType kFormatUrlOmitTrailingSlashOnBareHostname = 0 << 2;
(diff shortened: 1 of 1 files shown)


PF4Public:
Have you changed the order of fields or is it the diff weird?

Nifury:
I believe I only changed two or three lines. It's just the way GitHub shows the diff of a diff (patch) is confusing.

PF4Public:
Does `kFormatUrlOmitMobilePrefix` have any relation to mailto? I was thinking it removes "m." part of  a URL.

Please add your reasoning for excluding mailto into the patch header just for the record.

Nifury:
> Does kFormatUrlOmitMobilePrefix have any relation to mailto?

No it doesn't, and I didn't touch that line. I only changed the value of `kFormatUrlOmitMailToScheme` from 0 to 1. It will be more clear if you check the new file instead of the diff view. 

> Please add your reasoning for excluding mailto into the patch header just for the record

Added 😀

PF4Public:
> > Does kFormatUrlOmitMobilePrefix have any relation to mailto?
> 
> No it doesn't, and I didn't touch that line. I only changed the value of `kFormatUrlOmitMailToScheme` from 0 to 1. It will be more clear if you check the new file instead of the diff view.

Sorry I was inattentive


> > Please add your reasoning for excluding mailto into the patch header just for the record
> 
> Added

And sorry again, I should've expressed myself clearer. I meant the header of the patch itself, just below `# Disables omission of URL elements in Omnibox`. With it, there will be no need to inspect blame or git history to obtain this information.

Nifury:
Ah I thought any comment in a patch will be removed by quilt. I didn't know quilt was able to manage header comments.

PF4Public (approved):
LGTM

Ahrotahn (approved):
This looks good to me!

PR #2256, opened 2023-02-27
Introduce script to clone source tree

Ahrotahn:
This PR adds a new script to clone the Chromium source and updates the existing tasks for its use when tarballs are not available.  

`utils/clone.py` can be run without arguments but has a few available:
* `-o`, `--output`: By default the script will clone into a directory named `chromium`.  This flag allows you to set the output directory to an alternative.
* `-c`, `--custom-config`: The script will create a config file for gclient. This flag allows supplying a path to a substitute to be used instead. 
* `-p`, `--pgo`: The linux PGO profile is download by default. This flag changes the profile downloaded to one of the currently available ones: `linux`, `mac`, `mac-arm`, `win32`, `win64`

---

Google had made changes to their buildbots to only generate linux tarballs for a period then later reverted the changes.  This PR was started as a result of those changes.  This script can still be helpful even though the tarballs are now being generated again.  For example the recent 112.0.5615.121 release took nearly six hours for the tarballs to become available after the release announcement.  These changes would allow work to begin on new releases as soon as the new release's tag is made available.


teeminus:
Thanks for implementing the new process.

Do we still need the entry in the `downloads.ini` files?

I tried running the script on windows and I am currently stuck at `Running gsync`. For windows, we need to call `gclient.bat` instead of `gclient`. After fixing that, I am currently stuck with a problem in the file `depot_tools\git_cache.py`. The file expects `git.bat` as a git executable (or wrapper) for windows, which is not present (at least not on my machines). I manually fixed that, but I don't know how to fix that for automated builds.

I will report back once I got any updates on my progress.

Ahrotahn:
I'll add a check to run the .bat on windows.  We could make the clone script create the `git.bat` in the depot_tools folder or edit `git_cache.py` directly.  What change(s) worked for `git_cache.py`? Should that be `git` or `git.exe`?

Also it seems the docker image we used has updated to bullseye since we last made changes to it, so I'll have to specify buster to get the tasks to run.

teeminus:
I think the '.exe' suffix is not required, 'git' worked fine for me. But I think creating a simple 'git.bat' wrapper is the best solution as it does not require patching any files this early in the build process.

teeminus:
Which python version are you using? I get a lot of errors in the gclient scripts with python 3.10.4. It seems like some deprecated stuff has been removed in 3.10 which breaks the scripts.

teeminus:
So it turns out that we need to use a specific version of the `depot_tools` (`41a2d0f1a0173723f63ca2994e17c81eaf302b65`, see `src\DEPS`). When using that revision (and installing the packages `httplib2` and `six` using pip), gysnc finishes without any errors.

Now I am stuck with downloading the node modules, but I'll keep you updated once I figured out why it does not work.

Ahrotahn:
I had added httplib2 to the python requirements for cirrus and forgot to mention that.  I'll need to update the documentation to include it.  Maybe we should add a requirement check to the script.  I have six installed on my machine, but I didn't specify that in the requirements for the tasks.  I'll have to check if that is pulled in as a dependency for one of the other packages.  

The folder structure is like that because we need to have separate directories for depot_tools and gn.  We could instead create two directories: one for the chromium source and one that contains depot_tools and gn.  I find it to be a little tidier to have them in a single directory though since deleting that directory is all you would need to do to start fresh.  The chromium source directory is named 'src' by default but we could change that to something more convenient if needed.

teeminus:
I am totally fine with the new folder structure, I was just wondering as it differs from the tarball.

Thanks for adding my suggestions to the clone script, it works great. For the python executable, I don't think we need to use a different one for windows, just `python` should work (I will test that once my machine is stuck next time or finishes compiling).

For the `depot_tools`, I need to use a specific revision for windows as the current HEAD does not work. I wonder why this only affects windows so far, however, there is a revision listed in the DEPS file that works. For now I changed the script to do a full clone and check out the revision afterwards. The repo is not that big so it should not increase disk usage or fetch time that much.

Then I was wondering why the script clones `gn` manually, it's included in the chromium repo. Can't we use that?

Next step was to fix the pruning list. It took some time to compute the list (and fine tune it), but that worked well. I noticed, that my list contains more entries as before, but that might result from `gclient` runing on windows rather than on linux. Maybe those lists need to be platform specific now. A general question here: What is the purpose of the pruning process? I kept it for granted but now that I had some troubles with it I wondered why it's done in the first place. It deletes some executables that have to be (re-)downloaded manually after the pruning, which seems kind of counterintuitive.

The first compile attempt failed quite early as some sources were missing. But after adjusting the `gclient` config, that problem was gone too (at least for now, I am waiting for the compiling to finish right now). I think it's best here to have a config per target platform, as the platform is part of the config as well.

Again, thanks for your great work, I really appreciate it :)

PF4Public:
Will the difference in folder structer affect Linux, which could still use tarballs?

Ahrotahn:
>For the python executable, I don't think we need to use a different one for windows, just python should work (I will test that once my machine is stuck next time or finishes compiling). EDIT: Just python works fine for windows.

Great, that would simplify things.  I'll make an update to that once I get time.  

>For the depot_tools, I need to use a specific revision for windows as the current HEAD does not work. 

What fails when you try to use HEAD?  We could parse the commit, but that would complicate things a bit.  

>I noticed, that my list contains more entries as before, but that might result from gclient runing on windows rather than on linux. Maybe those lists need to be platform specific now.

Oof, I was hoping that would be something we could avoid.  But that might be due to:

>The first compile attempt failed quite early as some sources were missing. But after adjusting the gclient config, that problem was gone too

What adjustments were needed for your build?  

We *could* have the platform repos maintain their own lists, but the problem with that is non-linux contributors would need to make their changes on a linux clone which would complicate the process for them.  I think it would be better to have a single setup that would work for all desktop platforms if possible

>What is the purpose of the pruning process? I kept it for granted but now that I had some troubles with it I wondered why it's done in the first place. It deletes some executables that have to be (re-)downloaded manually after the pruning, which seems kind of counterintuitive.

It's mainly to prevent including a binary into the build that could potentially make unwanted connections or take unwanted actions.  Exceptions are sometimes made for build tools though.  So we should be able to add some exclusions for those that you are having to redownload if there's no easy way of having system versions of those tools.  

>Will the difference in folder structer affect Linux, which could still use tarballs?

It would, and so would the difference in the lists.  That is one of the trade-offs with this approach.  With the way the script is set up currently any platforms that want to build with the tarballs would need to maintain their own lists.  We could make the script replicate the tarball more closely but that requires ~4GB more data and would make the tasks take even longer.  It's still something we can experiment with if it ends up being too disruptive.  Cloning seemed like the best way to move forward considering how often Google's tarball bots have had issues, not to mention how long it takes them to get made when things are working properly.

PF4Public:
> Cloning seemed like the best way to move forward considering how often Google's tarball bots have had issues, not to mention how long it takes them to get made when things are working properly.

Please keep in mind that cloning significantly complicates packaging. There should be some kind of cache in order not to re-clone everything upon each update and that is not very easy with most packaging systems. I suggest keeping most compatibility to tars if possible.

teeminus:
Compiling succeeded :) The resulting package size differs only by a few bytes.


> What fails when you try to use HEAD? We could parse the commit, but that would complicate things a bit.

I don't know why, but now the current HEAD is working. I reverted all my changes, but couldn' reproduce the error... :/


> Oof, I was hoping that would be something we could avoid.

Well, having platform specific lists makes sense to me. I am not sure if `gclient` produces the same result on different platforms, so having a platform specific pruing list would prevent missed binaries. And as the pruning script gets the list as a parameter, changing the list is pretty easy.


> Exceptions are sometimes made for build tools though. So we should be able to add some exclusions for those that you are having to redownload if there's no easy way of having system versions of those tools.

Those exclusions could be part of the documentation on how to prepare new updates. Ideally as a list of `sed` statements, to make it even simpler.


> What adjustments were needed for your build?

<details>
<code>
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -43,7 +43,6 @@ def clone(args):
                               "src/buildtools/linux64": None,
                               "src/buildtools/mac": None,
                               "src/buildtools/reclient": None,
-                              "src/buildtools/third_party/libc++/trunk": None,
                               "src/buildtools/third_party/libc++abi/trunk": None,
                               "src/buildtools/third_party/libunwind/trunk": None,
                               "src/buildtools/win": None,
@@ -70,7 +69,7 @@ def clone(args):
                             },
                           },
                         ];
-                        target_os = ['unix'];
+                        target_os = ['win'];
                         target_os_only = True;
                         target_cpu = ['x64'];
                         target_cpu_only = True;
</code>
</details>
I am not sure how to make the `target_os` platform dependant.



> Please keep in mind that cloning significantly complicates packaging. There should be some kind of cache in order not to re-clone everything upon each update and that is not very easy with most packaging systems. I suggest keeping most compatibility to tars if possible.

I agree, cloning makes the process more complicated. But it's faster than downloading and extracting the tar. And as you get a new tar for each new release, you have to download and extract that new tar anyways. Integrating the new clone script into the windows build script was pretty easy and required just minor changes.

Ahrotahn:
>Please keep in mind that cloning significantly complicates packaging. There should be some kind of cache in order not to re-clone everything upon each update and that is not very easy with most packaging systems. I suggest keeping most compatibility to tars if possible.

Yeah that's a consideration I've been keeping in mind. The clone directory won't be removed unless the user manually removes it, and that is essentially the cache in that case.  The good news is that the time and download amount is significantly reduced on subsequent clones.  

>I don't know why, but now the current HEAD is working. I reverted all my changes, but couldn' reproduce the error... :/

Hopefully it was just the missing libraries from before that were the problem.  It would make things a lot cleaner if we can use the latest commit.  

>I am not sure how to make the `target_os` platform dependant.

Removing the `target_` entries will make gclient use the host os and cpu settings.  The thing is that the tarball bots were always using linux-specific targets, so it *should* work with the ones I have set.  I included those targets so the source would be the same across platforms.  If you get the time try working with the unix target, it may have just been the removal libc++ that was causing the issue.  I'll remove that removal along with the other changes sometime today.

teeminus:
> Hopefully it was just the missing libraries from before that were the problem. It would make things a lot cleaner if we can use the latest commit.

I was able to reproduce the problem. It's pretty strange because it only occures when the clone script is called from within the windows build script, not when I call the clone script directly. I fixed that problem by using `sys.executable` to call the clone script inside the build script. We could use the for the clone script as well: `pyexe = sys.executable`. Works just fine.

> Removing the `target_` entries will make gclient use the host os and cpu settings. The thing is that the tarball bots were always using linux-specific targets, so it _should_ work with the ones I have set. I included those targets so the source would be the same across platforms. If you get the time try working with the unix target, it may have just been the removal libc++ that was causing the issue. I'll remove that removal along with the other changes sometime today.

I compared the clone results of `target_os = ['unix'];` and `target_os = ['win'];`. Turns out, they differ. So far, compiling works with `unix` too, but it has not yet finished. I don't know if we want to set the variable per platform. On the one hand, the PDO profile can be customized too, on the other hand it would add more complexity to the clone script.

teeminus:
Compiling succeeded with `target_os=['unix']` as well.

Ahrotahn:
Awesome!  I appreciate you taking the time to test these changes.  

I still need to update the documentation and experiment with some ideas to improve the time it takes the tasks to run.  Since the unix target works I might be able to add some exclusions to match up with the tarballs as well.

teeminus:
Thanks :-)

I am totally fine with generating the pruning list for windows manually for each release. I need to update the windows documentation anyways, so adding these steps is not a big deal.

Ahrotahn:
Standardizing the targets was my attempt at making it so that you wouldn't have to do that!  If you get the time, remove your existing source directory, do a new clone, and update the lists.  I'd be interested to see what changes were needed.  I'll add exceptions for any host-specific additions so that you can use the same lists.  

---

I've updated the script to remove the test data directories.  This shaves off ~2GB from the cache archive, which is more significant than I expected!

teeminus:
Removing of unused files failes for windows, as python cannot delete readonly files. I copied the windows fix from the pruning script, now it works fine.
<details>
<code>
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -5,8 +5,9 @@ Module for cloning the source tree.
 """
</code>
<code>
 import sys
+import stat
 from argparse import ArgumentParser
-from os import chdir, environ, pathsep
+from os import chdir, chmod, environ, pathsep
 from pathlib import Path
 from shutil import copytree, copy, move
 from subprocess import run
@@ -209,11 +210,27 @@ def clone(args):
             for file in path.rglob('*'):
                 if file.is_file():
                     if file not in keep_files and file.suffix not in keep_suffix:
-                        file.unlink()
+                        try:
+                            file.unlink()
+                        # read-only files can't be deleted on Windows
+                        # so remove the flag and try again.
+                        except PermissionError:
+                            chmod(file, stat.S_IWRITE)
+                            file.unlink()
+                        except FileNotFoundError:
+                            pass
         for path in sorted(Path('src').rglob('*'), key=lambda l: len(str(l)), reverse=True):
             if not path.is_symlink() and '.git' not in path.parts:
                 if path.is_file() and ('out' in path.parts or path.name.startswith("ChangeLog")):
-                    path.unlink()
+                    try:
+                        path.unlink()
+                    # read-only files can't be deleted on Windows
+                    # so remove the flag and try again.
+                    except PermissionError:
+                        chmod(path, stat.S_IWRITE)
+                        path.unlink()
+                    except FileNotFoundError:
+                        pass
                 elif path.is_dir() and not any(path.iterdir()):
                     path.rmdir()
</code>
</details>

I updated the lists with a fresh clone, but there are differences. All of them are either windows binaries of removed folders.
image

Removed folders are:
- `third_party/android_rust_toolchain/toolchain/`
- `third_party/apache-linux/`

I have not tried (re-)compiling again (yet). Last time I had to update the list prior to compiling as pruning would fail otherwise.

Ahrotahn:
It seems like gclient is ignoring the overrides for cipd deps.  That's something else I'll have to look into.  
I had forgotten about the issues with readonly files.  I'll add a commit for that along with the other change once I get that sorted.

Ahrotahn:
So the changes to the tarball bot have been reverted:  [[4304241]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/4304241)

I'll continue to work on this so that we have an alternative for when they have issues again or if they change their minds again, but I don't see any reason to make it the default if we can still use the previous method.

PF4Public:
So, would it be possible to make source clone and tarballs be interchangeable?

Ahrotahn:
Yes, though there would need to be some compromises.  These are the main differences currently:  

<details><summary>domain_substitution.list:</summary>

```
*/__pycache__/*
buildtools/third_party/libc++abi/trunk/CREDITS.TXT
docs/website/*
native_client/*
third_party/android_rust_toolchain/*
third_party/dawn/tools/*
third_party/depot_tools/*
third_party/freetype-testing/*
third_party/llvm/*
```
</details>

<details><summary>pruning.list:</summary>

```
*/__pycache__/*
buildtools/linux64/gn
buildtools/reclient/*
docs/website/*
native_client/*
third_party/android_rust_toolchain/*
third_party/apache-linux/*
third_party/checkstyle/checkstyle-all.jar
third_party/dawn/third_party/ninja/ninja
third_party/dawn/tools/*
third_party/depot_tools/*
third_party/devtools-frontend/src/third_party/esbuild/esbuild
third_party/freetype-testing/*
third_party/llvm/*
third_party/ninja/ninja
tools/luci-go/*
tools/resultdb/result_adapter
tools/skia_goldctl/linux/goldctl
```
</details>

Some of the paths should be able to be included into the clone without causing much impact, like the docs and libc++abi.  I would have to experiment to see how much overhead nacl and the some of the third-party repos would bring.  
We could also add some of the paths as exclusions instead, but it would probably be best to keep that to a minimum.  

The tricky ones are the pycache files and the files downloaded from cipd.  
The bots have `PYTHONDONTWRITEBYTECODE` set in an attempt to prevent generation of cache files, but there are subprocesses that don't inherit the environment causing some to still be included in the tarball.  
Setting the target in `.cipd_client_platform` should allow us to download the linux binaries to match the tarball, but that would take more resources when cloning only to be pruned out afterwards.  

One idea I have would be to create a new exclusion category that would exclude matching paths from being added to the lists, but would still be attempted to pruned during the pruning step.  This would be able to easily deal with the pycache files and possibly the paths for the binaries retrieved from the cipd.

Ahrotahn:
I've been experimenting with which dependencies make the largest impact and it seems like it's worth allowing the majority for tarball compatibility with two exceptions. The llvm sources are checked out at an unnamed commit so the full history needs to be retrieved. There's also a very large GPU testing repo that isn't used for builds.  

So I've implemented a new list of paths that are always pruned if they exist. This makes the updated lists compatible between the tarball and the source cloned with this script!  

Also, it seems that cirrus' cache archiving performs much better when the sources are added to an uncompressed tar file.  It shaves a few minutes off of a fresh clone and validating with a populated cache now takes half as long as using the tarball.  It also allows for better compression.  The cache archive is now ~1.6GB which is the same size as the tarball's cache.

PF4Public:
> I've also updated the checks to attempt using the tarball by default. A clone will be made if the tarball isn't available.

Does it mean that no changes will be needed for existing building workflows that use a tarball?

Ahrotahn:
>Does it mean that no changes will be needed for existing building workflows that use a tarball?

That's correct, both methods should now be interchangeable!

Ahrotahn:
These changes have worked for everything I've thrown at it, so I think it is ready for review.  I've rebased over the latest changes and squashed the commits.  I'll update the PR in a bit.

devutils/update_lists.py
```diff
@@ -74,6 +75,7 @@
     '*.woff',
     '*.woff2',
     '*makefile',
+    '*.profdata',
```
PF4Public, line 78: So, profdata will be permanently excluded unlike how it is now?

Ahrotahn: Yes, keeping it in the pruning list would require the pgo archive to be downloaded and extracted, only to later delete the file.  Not to mention that some people want to utilize pgo profiles in their builds and have been manually exempting the file already.

PF4Public (approved):

PF4Public:
@networkException Have you had any chance to glance through the changes in this PR?

networkException (approved):

networkException:
Managed to squeeze in a more in depth look at the changes :)

Besides a rebase to fix the conflicts I have no objections, thanks for the hard work Ahrotahn!

networkException (approved):

networkException (approved):

networkException (approved):

PF4Public:
> But it's faster than downloading and extracting the tar.

> The good news is that the time and download amount is significantly reduced on subsequent clones.

By chance did anyone measure the difference in time and download amount of extracting a tar and updating working dir from one version to another?

PF4Public:
@Ahrotahn Please rebase changes over the master once more.

PF4Public (approved):

PF4Public:
So after merging this PR anyone who was excluding (myself included, no pun intended) PGO profdata, no longer needs to do so, right?

Ahrotahn:
That's correct!

PF4Public:
Merging since there was a prior approval from @networkException

rany2:
Why was `third_party/llvm` added to the prune list by this change? Could you reconsider this as it breaks the build manifest for Flatpak on ARM64; I've worked around it for the time being but I would like to remain on equal footing with vanilla Chromium.

rany2:
@Ahrotahn It's not a huge deal, the workaround I settled on isn't that bad anyway (just saving the LLVM source under a different name and renaming back when we're ready to build). It is mostly down to the Flatpak build specifically not having a prebuilt llvm we can use for chromium.

PR #2262, opened 2023-03-03
Update to Chromium 111.0.5563.64/65

Ahrotahn:
This PR updates to Chromium 111.0.5563.64.

Notes:
* The Debian parallel job patch is no longer needed after similar changes in [[4137674]](https://chromium-review.googlesource.com/c/chromium/src/+/4137674)

* A new exclusion in update_lists.py for [[4093175]](https://chromium-review.googlesource.com/c/chromium/src/+/4093175)

* Some base headers have been moved to base/functional in [[4152595]](https://chromium-review.googlesource.com/c/chromium/src/+/4152595) causing a lot of changes to header order

* A new addition to the safebrowsing patch for [[4162451]](https://chromium-review.googlesource.com/c/chromium/src/+/4162451). There's a commit in line that changes this, so hopefully it's a short-term addition. 

* The tarball bot changes have been reverted in [[4304241]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/4304241)!

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/110.0.5481.177..111.0.5563.64?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2023/03/stable-channel-update-for-desktop.html)


patches/extra/ungoogled-chromium/add-flag-to-hide-side-panel-button.patch
```diff
@@ -1,16 +1,6 @@
---- a/chrome/browser/ui/views/frame/browser_view.cc
-+++ b/chrome/browser/ui/views/frame/browser_view.cc
-@@ -955,6 +955,7 @@ BrowserView::BrowserView(std::unique_ptr
-         AddChildView(std::make_unique<ContentsSeparator>());
-     side_panel_coordinator_ = std::make_unique<SidePanelCoordinator>(this);
-   } else {
-+   if (!base::CommandLine::ForCurrentProcess()->HasSwitch("hide-sidepanel-button"))
```
PF4Public, line 7: Is this hunk no longer needed?

networkException:
Builds fine (.64), I'm crashing reproducibly when opening `chrome://settings` though

PF4Public:
Can confirm the crash at settings. Menu → about Chromium. Does not crash. Clicking the link "Google and I" immediately crashes the browser.

patches/extra/ungoogled-chromium/add-flag-to-hide-side-panel-button.patch
PF4Public: I stand corrected. The hunk I was suspecting is in fact untouched: https://github.com/ungoogled-software/ungoogled-chromium/commit/515cd084ce0889c4b84002eb9c6d1eb7c8fa77d2

@gz83 Could you elaborate on what could be broken?

Ahrotahn:
That was caused by a change in [[4128118]](https://chromium-review.googlesource.com/c/chromium/src/+/4128118), which was tough to figure out since it lazy-loaded the problem script.  

@networkException Do you prefer to have just the two version commits when tagging?  I can rebase the new change into the first one if so.

patches/extra/ungoogled-chromium/add-flag-to-hide-side-panel-button.patch
Ahrotahn: That change was for the previous iteration of the sidepanel.  Now that they've transitioned to the unified sidepanel only the change @PF4Public linked is needed, which is why the previous code was removed.

networkException:
Yes please include this in the .64 commit or in a separate commit before that

networkException (approved):

PR #2280, opened 2023-03-23
Import three Debian patches to better modularize changes

iskunk:
This PR aims to simplify building on Debian by importing three of their patches that currently conflict with later, larger u-c patches. This way, the u-c patches will apply cleanly after the Debian patch series, once the `patches/*/debian/` patches are left out.

Specifically, this change does the following:

1. Import these three patches from Debian (u-c paths shown):
   ```
   patches/core/debian/disable/signin.patch
   patches/core/debian/disable/tests.patch
   patches/core/debian/disable/third-party-cookies.patch
   ```
   (The upstream for these files can be browsed [here](https://sources.debian.org/src/chromium/111.0.5563.110-1/debian/patches/disable/), for reference)

2. Update the existing patches as follows:

   * Remove edit to `third_party/devtools-frontend/src/BUILD.gn`
     from `patches/core/ungoogled-chromium/fix-building-with-prunned-binaries.patch`
     as it is present in `patches/core/debian/disable/tests.patch`.

   * Modify edit to `components/signin/internal/identity_manager/primary_account_manager.cc`
     in `patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch`
     to account for changes made in `patches/core/debian/disable/signin.patch`.

     Note: `signin.patch` also contains a modification to `chrome/browser/ui/webui/signin/inline_login_handler_impl.cc` that is not currently part of the u-c patches. Please double-check it. I can update the PR to not include this modification, if desired.

   * Remove edit to `components/content_settings/core/browser/cookie_settings.cc`
     from `patches/extra/inox-patchset/0006-modify-default-prefs.patch`
     as it is present in `patches/core/debian/disable/third-party-cookies.patch`.

3. Add the new patches to `patches/series`, in the same order they are listed in the Debian patch series.

By the way, note that Debian has dropped two of their patches shipped by u-c, both in the chromium 105.0.5195.52-1 release. Here are the associated [changelog](https://metadata.ftp-master.debian.org/changelogs//main/c/chromium/unstable_changelog) comments:

> `disable/welcome-page.patch`: drop. Upstream fixed the original issue some time ago, and this new version finally cleaned up the workaround.

> `fixes/connection-message.patch`: drop it. I looked at sending this upstream, but the original extension doesn't exist any more, and chromium properly prints an error if a proxy is unreachable. If you can still reproduce the issue (described in http://bugs.debian.org/864539), let me know so I can get it fixed upstream.

### DIFF
--- a/patches/core/debian/disable/signin.patch
+++ b/patches/core/debian/disable/signin.patch
@@ -1,0 +1,37 @@
+description: disable browser sign-in
+origin: https://raw.githubusercontent.com/Eloston/ungoogled-chromium/master/resources/patches/ungoogled-chromium/disable-signin.patch
+
+--- a/chrome/browser/ui/webui/signin/inline_login_handler_impl.cc
++++ b/chrome/browser/ui/webui/signin/inline_login_handler_impl.cc
+@@ -401,17 +401,6 @@ void InlineSigninHelper::OnClientOAuthSu
+ 
+     signin_metrics::LogSigninReason(signin_metrics::Reason::kReauthentication);
+   } else {
+-    if (confirm_untrusted_signin_) {
+-      // Display a confirmation dialog to the user.
+-      base::RecordAction(
+-          base::UserMetricsAction("Signin_Show_UntrustedSigninPrompt"));
+-      Browser* browser = chrome::FindLastActiveWithProfile(profile_);
+-      browser->window()->ShowOneClickSigninConfirmation(
+-          base::UTF8ToUTF16(email_),
+-          base::BindOnce(&InlineSigninHelper::UntrustedSigninConfirmed,
+-                         base::Unretained(this), result.refresh_token));
+-      return;
+-    }
+     CreateSyncStarter(result.refresh_token);
+   }
+ 
+--- a/components/signin/internal/identity_manager/primary_account_manager.cc
++++ b/components/signin/internal/identity_manager/primary_account_manager.cc
+@@ -146,9 +146,9 @@ void PrimaryAccountManager::RegisterProf
+                                std::string());
+   registry->RegisterStringPref(prefs::kGoogleServicesAccountId, std::string());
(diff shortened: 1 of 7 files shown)


iskunk:
How about this:

* `tests.patch`: Instead of bringing this one in, would it be OK to split `fix-building-with-prunned-binaries.patch` into two separate patches (e.g. `...-1.patch` and `...-2.patch`) so that the second hunk can be easily excluded from the patch series?

* `third-party-cookies.patch`: Note that having all preference changes in a single patch would reduce granularity---it would become harder to enable only a subset of those changes, as well as reconcile the patch with others that conflict. The changes could be better organized, but I presume that reassembling them into new patch(es) would run afoul of authorship/licensing concerns.

  Would it help to place this patch under `patches/extra/debian/` instead?

* `signin.patch`: Removing the second hunk, and the corresponding edit to `remove-unused-preferences-fields.patch`, would cause the latter to not apply to the Debian tree. The goal here is to allow reconciling conflicts by removing certain patches from the series, instead of having to modify the patches.

* I'll prepare a separate PR to drop welcome-page and connection-message.

PF4Public:
Another concern for pulling in Debian patches is that they oftentimes lag behind Chromium releases, which means if any of the patches become broken, we need to fix them to push the next release, but when Debian catches up, patches may differ significantly (different lines, different code hunks, different approaches etc), all of which could potentially bring us undue maintenance burden, which is bad.

iskunk:
Thank you @Ahrotahn and @PF4Public for commenting. I'm closing this PR, as I've found an alternate approach to managing the conflicts between the Debian and u-c patchsets for now.

I would like to see better alignment between the u-c patches and Debian's (and possibly other distros, too), however. It's clear that there is some overlap in the modifications that these two parties make to the codebase, and I think it will ultimately be beneficial to structure the modifications in a way that takes this into account. The conflicts that exist at present are relatively minor, but the current approach will not scale to much more.

PR #2282, opened 2023-03-23
Debian updates

iskunk:
Update URL in the main README file (Debian has not had a `chromium-browser` package for years), and exclude quilt(1)'s `.pc/` directory when updating lists to allow working in a standard Debian source tree.


Ahrotahn (dismissed):
This seems fine, but I do have a slight hesitation including the quilt directory because it would imply that the source has been modified (even if the changes have been reverted).  Ideally the lists should be updated using a clean copy of the source, but I could see it being helpful when doing local development.  At any rate our code checks should fail if someone were to submit changes that don't apply cleanly to original source.

iskunk:
Thank you @Ahrotahn!

Note that the intent here is to produce a substitution list that is distinct from the official one, as it would be generated for the Debian-modified source tree (not just their patch series, but also starting from their DFSG-cleaned source tarball) rather than a pristine upstream one. I'm working to make the u-c process fit cleanly into the Debian build, and the list generated under these circumstances would only be used there.

iskunk:
My apologies, I completely missed that the `.pc/` exclusion will produce an error when the script is run in a non-Debian source tree. I've removed that change from this PR, and submitted a new one (#2290) that solves the problem in a different way. This PR is now just about updating the README.

PF4Public:
I guess these PRs are good to merge independently of the remaining one.

Thanks @iskunk.

Issue #1979, opened 2022-05-30
First-run welcome page


PF4Public:
May I write the body-text later :)

> It would be pretty easy to have a page open on first launch. I think the FAQ would be a better page than the readme since the browser is already installed and end users would be more likely to want to know how to install extensions, enable spellcheck, widevine, etc.

_Originally posted by @Ahrotahn in https://github.com/Eloston/ungoogled-chromium/discussions/1559#discussioncomment-1498350_

PF4Public:
By he way, showing this First-run page could be one of the options for #1971

jstkdng:
I mean, we've already crossed the line from just ungoogling chromium with all the custom flags we have. That being said, no complex feature has been added yet, so I think a static page linking to more information for ungoogled chromium wouldn't be bad.

Ahrotahn:
I did some [experimentation](https://github.com/Ahrotahn/ungoogled-chromium/commit/76da30e1e4d4a1d5d50f8ca372c692ed0c082c00) earlier this week and found a way to add a new internal page that shouldn't cause future headaches.  

I think the FAQ, flags, support, and contributing pages make the most sense to link.  The FAQ has answers to most of the common questions people have and the flags doc should cover the rest.  The support page might help direct people to the correct repo if their issue is platform-specific.  The contributing page probably isn't necessary but maybe it could get some other people involved with the project.  

Are there any other pages would be a good idea to include?  Is there any specific verbiage that would be useful to have on the page?

Another interesting thing to consider is that since it's an internal page we'd be able to link other ones.  So we could have a link to `chrome://flags` or we could mention that cookies are cleared by default on new profiles with a link to `chrome://settings/cookies` for example.

jstkdng:
Would it be possible to be able to change some settings in that page? Like the default cookie policy and/or the default search engine

networkException:
I think we should stick for a first launch page for now and look into further customizations down the line.

---

While this would advance the project more in the direction of a browser instead of a patchset it could also allow us to better address the concerns of different audiences:

Out of the box we could leave all settings to their default for example, provide a hardened or more privacy oriented profile directly on the welcome screen though.

If a user wishes to just use chromium with no distractions they would have to never have that welcome screen shown again.

Another user might want to get an update summary for every release with new flags / settings to choose from.

tomasz1986:
> the FAQ

[...]

PR #2289, opened 2023-03-28
Drop obsolete Debian patches

iskunk:
These patches were dropped in the 105.0.5195.52-1 release of the upstream Debian Chromium package, with the following [changelog](https://metadata.ftp-master.debian.org/changelogs//main/c/chromium/unstable_changelog) comments:

- `disable/welcome-page.patch`: drop. Upstream fixed the original issue some time ago, and this new version finally cleaned up the workaround.
- `fixes/connection-message.patch`: drop it. I looked at sending this upstream, but the original extension doesn't exist any more, and chromium properly prints an error if a proxy is unreachable. If you can still reproduce the issue (described in http://bugs.debian.org/864539), let me know so I can get it fixed upstream.


PF4Public (approved):
No objections from me. 

Besides we might have [our own welcome page… maybe… someday…](https://github.com/ungoogled-software/ungoogled-chromium/issues/1979)

Ahrotahn (approved):
No issues in my build with these changes.  

I'll see about setting up a draft for the welcome page sometime soon.

PF4Public:
I guess these PRs are good to merge independently of the remaining one.

Thanks @iskunk.

PR #2290, opened 2023-03-28
Make update_lists.py more flexible

iskunk:
This PR is spawned from #2282. The intent is for `update_lists.py` to support working with alternative Chromium source trees, like Debian's. Two new options allow

1. Specifying additional domain exclusion prefixes on the command line;

2. Downgrading unused patterns/prefixes from errors to informational messages.

#.1 is useful for Quilt's `.pc/` directory that may be present at the top of the source tree, and #.2 comes in handy when the source tree has already been pruned of offending binaries.

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -133,17 +133,18 @@ class UnusedPatterns: #pylint: disable=too-few-public-methods
         for name in self._all_names:
             setattr(self, name, set(globals()[name.upper()]))
 
-    def log_unused(self):
+    def log_unused(self, error=True):
         """
         Logs unused patterns and prefixes
 
         Returns True if there are unused patterns or prefixes; False otherwise
         """
         have_unused = False
+        log = get_logger().error if error else get_logger().info
         for name in self._all_names:
             current_set = getattr(self, name, None)
             if current_set:
-                get_logger().error('Unused from %s: %s', name.upper(), current_set)
+                log('Unused from %s: %s', name.upper(), current_set)
                 have_unused = True
         return have_unused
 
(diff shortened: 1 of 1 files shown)


Ahrotahn:
The changes seem ok to me, though I'm curious about the use for the second change.  Why would the pruning script need to be run over a source directory that has already been pruned?  Does the Debian source still contain some binaries that need to be removed?

iskunk:
When `update_lists.py` runs over the Debian source tree, both the pruning and domain-sub patterns have some misses:
```
ERROR: Unused from PRUNING_EXCLUDE_PATTERNS: {'*.ttf', '*.mk'}
ERROR: Unused from DOMAIN_INCLUDE_PATTERNS: {'*.mk'}
ERROR: Unused from DOMAIN_EXCLUDE_PREFIXES: {'tools/gn/src/gn/visual_studio_writer.cc'}
```

While I'm not interested in pruning binaries beyond what Debian already does (some shady stuff does remain, but better to update the upstream Debian build IMO), the pruning list is generated together with the domain-sub one, so it's not like I can request one without the other. (Not much point in adding an option to make that possible, either, as the pruning list can be informative even when it is not acted upon.)

PF4Public:
@braewoods Please leave your comment on this PR w.r.t. building on Debian.

braewoods:
> @braewoods Please leave your comment on this PR w.r.t. building on Debian.

And say what exactly? I'm not familiar enough with this script to know what it even does. Our current build process only uses `downloads.py` and `prune_binaries.py`.

iskunk:
Hi everyone, I think some more background on what I'm doing should be helpful here.

My goal is to build ungoogled-chromium on a Debian Chromium base, instead of on the raw upstream source. I want to take advantage of the release-engineering work performed by the Debian Chromium Team, not least since they make it possible to (1) build Chromium using the older toolchain of the Debian stable release, and (2) build Chromium at all for 32-bit x86, as that arch is no longer supported upstream.

I've already been contributing to an effort to build Debian Chromium for Ubuntu, as the latter no longer provide the package in their repos, and would like to add the option of an "ungoogled" variant as well. However, as the u-c process has to date been geared toward working with the upstream source, I've run into a few pain points using the tooling in the Debian source tree. These PRs are intended to help facilitate that work, while not making anything (overly) Debian-specific.

Ahrotahn (approved):

PR #2299, opened 2023-04-02
Update patch to fix errors in Debian-based build

iskunk:
In building a Debian Chromium-based ungoogled-chromium on the 111.0.5563.110 source, I encountered only a couple of minor compilation errors. My presumption is that these are coming up due to the additional build flags set by Debian.

(This build included the u-c GN flags. An earlier attempt was giving me numerous errors in the SafeBrowsing code, before I realized that this was necessary.)

```
In file included from ../../content/browser/file_system_access/fake_file_system_access_permission_context.cc:5:
../../content/browser/file_system_access/fake_file_system_access_permission_context.h:52:65: error: only virtual member functions can be marked 'override'
      base::OnceCallback<void(AfterWriteCheckResult)> callback) override;
                                                                ^~~~~~~~
1 error generated.
```
The `core/ungoogled-chromium/fix-building-without-safebrowsing.patch` patch removes the `PerformAfterWriteChecks()` method from the `FileSystemAccessPermissionContext` class in `content/public/browser/file_system_access_permission_context.h`, but does not remove it from the `FakeFileSystemAccessPermissionContext` subclass in `content/browser/file_system_access/fake_file_system_access_permission_context.{h,cc}`.

```
In file included from ../../content/browser/file_system_access/mock_file_system_access_permission_context.cc:5:
../../content/browser/file_system_access/mock_file_system_access_permission_context.h:60:65: error: only virtual member functions can be marked 'override'
      base::OnceCallback<void(AfterWriteCheckResult)> callback) override;
                                                                ^~~~~~~~
1 error generated.
```

The same occurs here. The `MockFileSystemAccessPermissionContext` subclass in `content/browser/file_system_access/mock_file_system_access_permission_context.{h,cc}` needs the same method removed.

This PR updates the patch in question, and yields a clean build for me.

### DIFF
--- a/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
+++ b/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
@@ -2438,3 +2438,66 @@
      "//components/safe_browsing/content/browser/web_ui",
      "//components/safe_browsing/content/common:interfaces",
      "//components/safe_browsing/content/renderer:throttles",
+--- a/content/browser/file_system_access/fake_file_system_access_permission_context.cc
++++ b/content/browser/file_system_access/fake_file_system_access_permission_context.cc
+@@ -46,13 +46,6 @@ void FakeFileSystemAccessPermissionConte
+   std::move(callback).Run(SensitiveEntryResult::kAllowed);
+ }
+ 
+-void FakeFileSystemAccessPermissionContext::PerformAfterWriteChecks(
+-    std::unique_ptr<FileSystemAccessWriteItem> item,
+-    GlobalRenderFrameHostId frame_id,
+-    base::OnceCallback<void(AfterWriteCheckResult)> callback) {
+-  std::move(callback).Run(AfterWriteCheckResult::kAllow);
+-}
+-
+ bool FakeFileSystemAccessPermissionContext::CanObtainReadPermission(
+     const url::Origin& origin) {
+   return true;
+--- a/content/browser/file_system_access/fake_file_system_access_permission_context.h
++++ b/content/browser/file_system_access/fake_file_system_access_permission_context.h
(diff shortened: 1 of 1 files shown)


PF4Public:
I wonder why aren't other Linux distros affected…

iskunk:
I'm appending the [u-c build flags](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/flags.gn) to [the long list](https://salsa.debian.org/chromium-team/chromium/-/blob/master/debian/rules#L41-107) already set by Debian, whereas I suspect most folks building u-c use only the former. I'm pretty sure that's the cause of this, though I haven't gone in to track down which exact flag is at fault.

(I did confirm that, of the subset of flags that u-c and Debian set in common, there is no disagreement on the values thereof.)

Ahrotahn:
If that's the case then this should be fine.  When you get the chance, please rebase over the new changes for 112 and refresh the changes.  

>Is there some kind of automated coverage test you guys use to make sure everything works?

Unfortunately no.  For example github has a a six-hour time limit on actions which isn't even enough time to complete a single build for a single platform.  We would probably need access to a couple dedicated threadripper servers to be able to automate that.

iskunk:
Rebased as requested, and confirmed that it builds with Debian's 112.

While stuffing a build + test suite into CI would be ideal, I'd be happy to just have something I can run locally, instead of having to manually run the browser through its paces. (I also maintain AppArmor profiles for Chromium and Firefox, and it would be useful for that too.) Is there anything like that?

Ahrotahn (approved):
This looks good to me, the changes built and ran without issue.  

---

>Is there anything like that?

There's tests for script and patch integrity, but no there hasn't been much effort to set up any tests for uc-specific changes made in the browser.

PF4Public (approved):

PR #2300, opened 2023-04-02
New utility: make_domsub_script.py

iskunk:
It is common for distribution package builds to start with a pristine (upstream) source tarball, and apply one or more patches to the source before compiling it. The u-c patches fit into this scheme nicely. However, the domain-substitution step requires not only the `domain_substitution.py` script, but also the various Python bits it depends on, a file with the list of files to operate on, and a file with the regexes. As a result, you need to include a non-trivial portion of the u-c tree with the rest of the distro build files/patches in order to perform that step.

My first thought on eliminating that dependency was to generate a patch that applies statically the same domain substitutions that the script applies dynamically. But as it happens, this patch is about 34 MB in size. No one will want to review that!

So instead, I put together this utility, which will generate a self-contained script that performs the domain substitution. (It also saves the original files in a .tar file, to allow easy reversion.) This script is only about 680 KB in size, and most of that is just the list of files. It can be easily integrated into a package build process, and reviewing it is a breeze.

The generated script uses Bourne shell and Perl, and looks like this:
```
#!/bin/sh -e
#
# This script performs domain substitution on the Chromium source files.
#
# Generated by make_domsub_script.py, part of the ungoogled-chromium project:
# https://github.com/ungoogled-software/ungoogled-chromium.git
#

# Check that we are inside the Chromium source tree
test -f build/config/compiler/BUILD.gn

# These filenames may contain spaces and/or other unusual characters
print_file_list() {
        cat <<'__END__'
BUILD.gn
PRESUBMIT.py
PRESUBMIT_test.py
ash/accelerators/accelerator_notifications.cc
ash/ambient/ambient_controller_unittest.cc
ash/ambient/backdrop/ambient_backend_controller_impl.cc
[... >10K-line-long list of files elided ...]
weblayer/browser/prefetch_browsertest.cc
weblayer/browser/weblayer_security_blocking_page_factory.cc
weblayer/browser/weblayer_variations_http_browsertest.cc
weblayer/public/prerender_controller.h
weblayer/shell/app/shell_main_params.cc
__END__
}

echo "Creating backup archive ..."

backup=domain-substitution.orig.tar
print_file_list | tar cf $backup --verbatim-files-from --files-from=-

echo "Applying ungoogled-chromium domain substitution to 11114 files ..."

print_file_list | xargs -d '\n' perl -0777 -C0 -pwi -e '
    s#fonts(\\*?)\.googleapis(\\*?)\.com#f0ntz${1}.9oo91e8p1${2}.qjz9zk#g;
    s#google([A-Za-z\-]*?\\*?)\.com(?!mon)#9oo91e${1}.qjz9zk#g;
    s#gstatic([A-Za-z\-]*?\\*?)\.com#95tat1c${1}.qjz9zk#g;
[... more regexes ...]
    s#(?<!http://schemas.)android(\\*?)\.com#8n6r01d${1}.qjz9zk#g;
    s#goo\.gl(e?)#goo.gl${1}.qjz9zk#g;
    s#privacysandbox([A-Za-z\-]*?\\*?)\.com#pr1v4cy54ndb0x${1}.qjz9zk#g;
'

# end
```


iskunk:
```
python3 -m yapf --style '.style.yapf' -e '*/third_party/*' -rpd utils
--- utils/make_domsub_script.py	(original)
+++ utils/make_domsub_script.py	(reformatted)
@@ -85,8 +85,7 @@
     parser = argparse.ArgumentParser()
     parser.set_defaults(callback=_callback)
 
-    parser.add_argument(
-        '-r', '--regex', type=Path, required=True, help='Path to domain_regex.list')
+    parser.add_argument('-r', '--regex', type=Path, required=True, help='Path to domain_regex.list')
     parser.add_argument(
         '-f', '--files', type=Path, required=True, help='Path to domain_substitution.list')
     parser.add_argument(
```

Uh... do you want me to make this edit? It'll make the line stick out like a sore thumb, IMO.

iskunk:
Tweaked the here-doc quoting, thanks to some third-party feedback.

PF4Public:
Maybe mention this utility in /docs/building.md?

iskunk:
I'm happy to add some verbiage for the new script, but since the process described in `building.md` relies heavily on the various u-c scripts, introducing one with the goal of *not* needing said scripts seems at odds with the context. Is there a particular approach you have in mind for the mention?

(What I'm doing might motivate a new "Building ungoogled-chromium for packagers" section/document, since---at least in the Debian-package scenario---the process is necessarily different.)

iskunk:
Okay, I've applied that edit.

This script is aimed at a somewhat different audience from the one directly targeted by the u-c docs---more to those who want to prepare u-c packages for others, rather than for their own consumption, and thus need to follow an existing, established workflow for package preparation. The documented process for building u-c is an awkward fit at best in that context, given the bulk of the tooling required. The script generated by this script is intended to be equivalent to a patch (which most packaging workflows already handle natively), only being a script due to practical considerations.

As long as u-c remains in principle a modification of upstream Chromium rather than a project with its own independent codebase, as long as distribution-package builds follow a "start with original source, apply patches, compile" approach to managing their upstreams, I think this script will be broadly useful to that audience. My long-term hope is that ungoogled-chromium will eventually be accepted into the package repos of popular distributions, and giving love to packagers will help bring that about.

PF4Public:
So is it acceptable for core repo or should it go into contrib? @Ahrotahn @networkException

networkException:
I don't mind it being in here

Ahrotahn (approved):
That's fine with me!

PF4Public (approved):

PF4Public:
Thanks @iskunk

iskunk:
Thanks everyone, I felt this was the right place for the new script. I have some further work that builds on this, and differs in being decidedly distro-specific, so not appropriate here. Please have a look at the [new issue](https://github.com/ungoogled-software/contrib/issues/15) I've opened in the contrib repo.

PR #2359, opened 2023-05-23
Set default WebRTC policy, add default settings documentation

Ahrotahn:
This PR sets the WebRTC IP handling policy to disable non-proxied UDP by default.  This also introduces new documentation for the default settings changed by ungoogled-chromium.  

The WebRTC IP handling policy can be changed with the existing `--webrtc-ip-handling-policy` switch and #1687 adds an entry in chrome://flags.  

This default may no longer be necessary if mDNS is re-enabled in #2116.

### DIFF
--- a/docs/default_settings.md
+++ b/docs/default_settings.md
@@ -1,0 +1,21 @@
+# Default Settings
+
+Vanilla Chromium settings are not often changed by ungoogled-chromium, however there are some exceptions.  
+Below is a list of the documented changes to the default Chromium settings.
+
+Setting | New State | Location
+-- | -- | --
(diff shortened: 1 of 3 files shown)


networkException (approved):
Minor nit, I think these could be separate commits, all good otherwise

networkException (approved):

PF4Public:
Do I get it right: it is supposed to supersede #1845 and it is a part it it?

Will we be dealing with licensing demarcation somehow someday?

Ahrotahn:
Yup!  This shouldn't run into any licensing issues as far as I'm aware.

PF4Public (approved):

PR #2380, opened 2023-06-11
Fix linking chromedriver

implicitfield:
Turns out that the fix to the referrer patch in the windows repo isn't actually windows-specific, but is actually needed on all platforms to fix linking ```chromedriver```. All credit goes to @teeminus for this patch.


PF4Public (approved):

teeminus:
> Builds and runs without issue on my end.  Good job figuring that one out @teeminus!

Nah, I wouldn't have figured that out without your help.

PF4Public:
@implicitfield Should we bump release to `-2` or you could just temporarily add this patch to mac repo?

PR #2386, opened 2023-06-13
Update to Chromium 114.0.5735.133/134

Ahrotahn:
This PR updates to Chromium 114.0.5735.133/134.
There were no notable changes on our end.  

I'll mark as ready once the tarballs are available.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/114.0.5735.110..114.0.5735.134?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2023/06/stable-channel-update-for-desktop_13.html)


Ahrotahn:
The tarballs should be available now.  
I've rebased over recent merges and added a commit to handle a v8 profile added in [[4380399]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/4380399)

PF4Public:
I think we are good now. Other PRs will be read when they are ready.

PF4Public:
@Ahrotahn Have you considered adding [a Windows tabsearch fix](https://github.com/ungoogled-software/ungoogled-chromium-windows/pull/242#issuecomment-1589663007)?

PS: @Ahrotahn You should not be apologizing! I should apologize for forcing all PRs 😇

Nifury:
That works on my end. 
It would be great if you could add it in, or we can wait for @teeminus to confirm.

networkException (approved):

PF4Public:
I noticed following warning in build log:
```
../../components/safe_browsing/core/browser/safe_browsing_metrics_collector.cc:283:1: warning: non-void function does not return a value [-Wreturn-type]
  283 | }
      | ^
1 warning generated.
```

I suspect this might be here:
https://github.com/ungoogled-software/ungoogled-chromium/blob/aa8ef97462c97419544bf0d6ee738fb5f5dca81f/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch#L4340-L4359

Maybe we should return something just in case?

PR #2403, opened 2023-07-03
Remove non-functional Live Caption checkbox from media controls

PF4Public:
Before:
livecapt
After:
image

<s>Maybe we could also add `enable_speech_service=false` to gn flags.</s> Disabling speech services also removes media controls.

### DIFF
--- a/patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
+++ b/patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
@@ -20,6 +20,7 @@
 # the (?) learn more button on many settings pages
 # unneeded elements from the profile menu
 # the 'Learn more' link on crashed tabs
+# disable LiveCaption flag by default, this also removes non-functional Live Caption checkbox from media controls
 
 --- a/chrome/browser/resources/extensions/item_list.html
 +++ b/chrome/browser/resources/extensions/item_list.html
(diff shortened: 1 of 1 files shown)


PF4Public:
I did consider disabling kLiveCaption by default, but it allows end-user to enable it and get this checkbox back.

Ahrotahn:
Then in that case it might be better to do something like this:

```diff
--- a/components/live_caption/caption_util.cc
+++ b/components/live_caption/caption_util.cc
@@ -128,7 +128,33 @@
 }
 
 bool IsLiveCaptionFeatureSupported() {
-  if (!base::FeatureList::IsEnabled(media::kLiveCaption))
-    return false;
-
-// Some Chrome OS devices do not support on-device speech.
-#if BUILDFLAG(IS_CHROMEOS_ASH)
-  if (!base::FeatureList::IsEnabled(ash::features::kOnDeviceSpeechRecognition))
-    return false;
-#elif BUILDFLAG(IS_CHROMEOS_LACROS)
-  if (!chromeos::BrowserParamsProxy::Get()->IsOndeviceSpeechSupported())
-    return false;
-#endif
-
-#if BUILDFLAG(IS_LINUX)
-  // Check if the CPU has the required instruction set to run the Speech
-  // On-Device API (SODA) library.
-  static bool has_sse41 = base::CPU().has_sse41();
-  if (!has_sse41)
-    return false;
-#endif
-
-#if BUILDFLAG(IS_WIN) && defined(ARCH_CPU_ARM64)
-  // The Speech On-Device API (SODA) component does not support Windows on
-  // arm64.
   return false;
-#else
-  return true;
-#endif
 }
 
 }  // namespace captions
```

</details>


The only reason I suggest that is because there are large changes for livecaptions that will be coming up soon (for example [[4389333]](https://chromium-review.googlesource.com/c/chromium/src/+/4389333)), so the changes in this patch will have to be redone again once those land.

PF4Public:
That's huge! Making maintenance a nightmare for us is the last thing I want, so disabling the flag should be good enough.  I'm currently rebuilding with the patch to disable livecaptions by default and if that is good enough (I remember disabling this flag via `chrome://flags` was sufficient to suppress this checkbox, so this mus work), I'll update the PR.

PF4Public:
Yep, no checkbox if no one touches `Live Caption` flag.

I've just force-pushed the branch.

Ahrotahn (approved):
Works for me!

networkException (approved):

PR #2406, opened 2023-07-08
Fully remove unused safebrowsing function and Additions to block-requests

Ahrotahn:
This PR includes some odds-and-ends changes:

* remove-unused-preferences-fields.patch was updated to fully remove an unused function, mentioned in https://github.com/ungoogled-software/ungoogled-chromium/pull/2386#issuecomment-1594981431

* Some inclusions to block-requests.patch:
  * I have re-added comments for each hunk in the patch that had been previously eaten by quilt
  * Added a removal for password manager attempts to determine affiliations ([site affiliation info](https://developer.chrome.com/blog/site-affiliation/))
  * Added a removal for devtools attempting to download extra css information

Those requests were blocked via the trk scheme and domain substitution, but these changes prevent the attempt in the first place.

### DIFF
--- a/patches/core/ungoogled-chromium/block-requests.patch
+++ b/patches/core/ungoogled-chromium/block-requests.patch
@@ -1,5 +1,11 @@
 ## Prevent request attempts
 # chrome://discards/ attempts to use d3 to display the graph
+# New tab page tries to download background images
+# New tab page attempts to download the 'One Google' bar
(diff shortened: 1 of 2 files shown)


networkException (approved):

PF4Public (approved):
Did a quick rebuild on 114 with changes from this PR. No issues.

PR #2472, opened 2023-08-24
third-third-party binaries

PF4Public:


README.md
```diff
@@ -120,6 +120,8 @@ If your operating system is not listed above, you can also try to [**Download bi
 
 These binaries are known as **contributor binaries**.
 
+Please note that no support will be provided for binaries obtained outside of [ungoogled-software](https://github.com/ungoogled-software) GitHub organisation. Usage of `chrlauncher` is also discouraged.
```
networkException, line 123: Binaries referenced in https://github.com/ungoogled-software/ungoogled-chromium-binaries are a bit of a gray area I guess (with very few, like the Windows ones built using GitHub Actions being supported). How about

```suggestion
Please note that no support will be provided for these **contributor binaries** as well as all other binaries obtained outside of the [ungoogled-software](https://github.com/ungoogled-software) GitHub organisation. Usage of `chrlauncher` is also discouraged.
```

PF4Public: Good that you mentioned that! While writing this sentence it occurred to me that the gist of a problem with third-party binaries is the lack of convenience. Indeed people want to use ungoogled-chromium and enjoy convenient updates. Agreed, this sounds as if one wants to have one's cake and eat it. That being said if we reject any support for every binary, what's left for end-user? Build from source — inconvenient! Switch to more supported [by ungoogled-chromium] Linux distribution — inconvenient! Yeah, it diverts more into philosophical realm, but I'd be happy to hear your take on this dilemma.

Rusenche: > Good that you mentioned that! While writing this sentence it occurred to me that the gist of a problem with third-party binaries is the lack of convenience. Indeed people want to use ungoogled-chromium and enjoy convenient updates. Agreed, this sounds as if one wants to have one's cake and eat it. That being said if we reject any support for every binary, what's left for end-user? Build from source — inconvenient! Switch to more supported [by ungoogled-chromium] Linux distribution — inconvenient! Yeah, it diverts more into philosophical realm, but I'd be happy to hear your take on this dilemma.

@PF4Public, I don't know if you're only looking for input from those you approve of the project - two, three, or if you're looking for input from other users.

My opinion is:
- realistically - https://github.com/ungoogled-software/ungoogled-chromium - what helps me?! It doesn't help at all. Why?! Read below.
- a ready-to-use file cannot be downloaded from here;
- binaries (whatever you call them, I don't know why, I don't understand, because I'm a normal user/Ungoogled Chromium user only)) or some other files - there are no useful files here for a normal user, because a normal user is looking for a portable option and/or exe file/s.
- ask yourself the question - how many users can (have the capabilities, knowledge, etc.) themselves ... (how is it called correctly - I don't know) - "produce" file.exe or "x64 portable"?! The answer is probably - these users can be counted on the fingers of one hand.

I'm sure there will be offended, but you want objective/critical opinions right?! Then https://github.com/ungoogled-software/ungoogled-chromium - does nothing to help any normal/ordinary user in absolutely no way.

The opposite thesis - what is not recommended is written above:
- https://github.com/ungoogled-software/ungoogled-chromium-binaries;
- "chrlauncher"...

Hmm ok then what is recommended?! No recommendation - the user is at a dead end...

Next question - how does the common/normal user get Ungoogled Chromium (current Windows version)?!

Rational thought - there you go - I'm just a regular user and want to use Ungoogled chromium (up to date). But where do I download the portable version of every current version for Windows?! Answer - no and from here: https://github.com/ungoogled-software/ungoogled-chromium. Because the final files are not available. This is the raw and objective answer.

Because https://github.com/ungoogled-software/ungoogled-chromium will be useful for many advanced users (macchrome, uazo, teeminus, Nifury); programming professionals to use the stuff/patches or whatever it's properly called, but it doesn't contribute anything to the distribution of ordinary users, and it doesn't contribute at all to the distribution of Ungoogled chromium (for Windows).

And again - ask yourself how many users can take advantage of https://github.com/ungoogled-software/ungoogled-chromium and "produce" Ungoogled chromium themselves...

And since @PF4Public wrote that he would be happy to read opinions on the subject, I am always objective and critical.

PF4Public: @Rusenche Sorry if I wasn't clear, but I didn't ask for _complains_, I was asking for _solutions_.

I'll have to hide your reply since it does not propose any solution for the dilemma I mentioned.

tomasz1986: I think another problem with the binaries listed at https://ungoogled-software.github.io/ungoogled-chromium-binaries is that many of them are extremely outdated. If a user comes for help with their Ungoogled Chromium installation stuck 30 versions behind, is there really any support they can receive (other then being recommended to use a different variant available for their OS, if it even exists)?

I think it may be beneficial to somehow clearly mark which binaries are up-to-date and supported, and which are obsolete and shouldn't really be used until they are maintained again.

PF4Public: > but I'd be happy to hear your take on this dilemma.

@networkException

networkException: I fear much of this is future work, eventually I'd like to just drop anything that isn't build by CI in one of the platform repos in the main org and let that be it. So then its everything "we" as in anyone that got "approved" by the core team to work on binaries publish is getting support, everything else not

PF4Public: So you're still in support of excluding contributor binaries from being named as supported ones, right?

networkException: Yes, although currently I guess its not clear to end users that Windows binaries are compiled by CI and wouldn't count as such.

PR #2478, opened 2023-08-29
Update to Chromium 116.0.5845.140/141

PF4Public:
No notable changes for us.


PF4Public:
Builds and runs on my Linux box.

I have no idea why cirrus CI fails, tried restarting it several times to no avail.

networkException (approved):
Builds and runs

Ahrotahn (approved):
No issues here!

Ahrotahn:
> I have no idea why cirrus CI fails, tried restarting it several times to no avail.

I dug into this and it seems we're hitting a newly implemented limit:
`Pod ephemeral local storage usage exceeds the total limit of containers 8Gi.`

This coincides with the changes [mentioned in their blog post](https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/) but I haven't been able to find any info about storage limitations anywhere.  

Maybe we can skip extracting some of the files so everything fits within the limit, but we may be forced to look at other options if that can't be done.

PF4Public:
> we may be forced to look at other options if that can't be done.

I think GitHub actions could handle it. I'll look into it (as well as automating minor updates :) ) when I have more time to spare.

Ahrotahn:
I haven't tested this yet, but adding  
`--exclude='third_party/llvm/' --exclude='third_party/angle/third_party/VK-GL-CTS/src/'`
to the tar extraction would remove ~2.7GB and should be enough to have the tarball's files fit within the limit without causing issues.  Clones might be trickier but a new flag could be added to skip generating files and downloading profiles and the node modules since they aren't needed for validation.  

Moving to Github actions seems to be the better option regardless.

PF4Public:
> I think GitHub actions could handle it.

> Moving to Github actions seems to be the better option regardless.

Just checked, free runners for GitHub actions come with 14G storage and starting with 117 all my workflows fail miserably.  :( I use docker image, which takes some space for itself, but anyway, we'd better stick to cirrus for now.

networkException:
I have some hacks to fix this on GitHub Actions https://github.com/ungoogled-software/ungoogled-chromium-archlinux/blob/04bda2976752a921f87b5aa05529081a3867f48b/.github/actions/stage/index.ts#L46C1-L53

PF4Public:
> I have some hacks to fix this on GitHub Actions https://github.com/ungoogled-software/ungoogled-chromium-archlinux/blob/04bda2976752a921f87b5aa05529081a3867f48b/.github/actions/stage/index.ts#L46C1-L53

Does it save enough for Chromium to unpack?

Ahrotahn:
It looks like that frees up ~39GB, I'm surprised they would use such heavy images!  

I have been experimenting with updating the extraction script to exclude the directories in the contingent paths list.  It doesn't have as much of an impact on the extraction time as I thought it would, but it halves the space needed:  
before  
after
I've only implemented it for tar so far, so I just need to spend some time setting that up for the other extractors as well.

PF4Public:
> I have been experimenting with updating the extraction script to exclude the directories in the contingent paths list. It doesn't have as much of an impact on the extraction time as I thought it would, but it halves the space needed:

Does it build without third_party/llvm?

> It looks like that frees up ~39GB, I'm surprised they would use such heavy images!

So that means that free runners actually have more than 14G?

Ahrotahn:
> Does it build without third_party/llvm?  

Most platforms use their system's binaries instead those sources and the pruning script already removes that directory.  All this does is skip extracting directories that are about to be pruned out anyways:  https://github.com/Ahrotahn/ungoogled-chromium/commit/b98a976a01090e54dc5168fa4b674665486cea35

> So that means that free runners actually have more than 14G?  

Seems so!  There's more info on the repo:  https://github.com/easimon/maximize-build-space#how-it-works

PR #2521, opened 2023-09-22
added chrome-webrtc-flag info on first-run

### LINKED ISSUE #2516: WebRTC connection fails
### OS/Platform

macOS

### Installed

OS/Platform's package manager

### Version

117.0.5938.62

- [X] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there

### Description

Webrtc connection fails to connect

### How to Reproduce?

1. Install Ungoogled Chromium by `brew install --cask eloston-chromium`
2. Try connecting with [this webrtc](https://webrtc.github.io/samples/src/content/peerconnection/pc1/) example 
   - click on start
   - click on call

### Actual behaviour

It didnt get connect.

### Expected behaviour

It should get connected

### Relevant log output
Screenshot 2023-09-22 at 1 16 54 PM

### Additional context

_No response_

sonigeez:
fix #2516

### DIFF
--- a/patches/extra/ungoogled-chromium/first-run-page.patch
+++ b/patches/extra/ungoogled-chromium/first-run-page.patch
@@ -126,7 +126,15 @@
 +   </li>
 +   <li>Toggle spell check in <a href="chrome://settings/languages">chrome://settings/languages</a>,
 +    or restart the browser for the dictionaries to take effect.</li>
-+  </ol></details><p>
++  </ol></details>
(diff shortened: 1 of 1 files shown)


PF4Public:
> It is probably a better idea to add this to the FAQ and/or update [the default settings doc](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/default_settings.md) with another column that includes info about each setting's importance and why one might change it. Both are linked on the first-run page already.

Agreed

sonigeez:
fair enough as you wish

PR #2548, opened 2023-10-07
Update to Chromium 118.0.5993.70

Ahrotahn:
This PR updates to Chromium 118.0.5993.70.

Notes:

* The disable-webstore-urls patch had some changes due to [[4767959]](https://chromium-review.googlesource.com/c/chromium/src/+/4767959)

* New additions to fix-building-with-prunned-binaries for [[4762904]](https://chromium-review.googlesource.com/c/chromium/src/+/4762904) to handle building without tflite

* A handful of changes to the safebrowsing patch for [[4760985]](https://chromium-review.googlesource.com/c/chromium/src/+/4760985), [[4712670]](https://chromium-review.googlesource.com/c/chromium/src/+/4712670), and [[4808423]](https://chromium-review.googlesource.com/c/chromium/src/+/4808423). I've also removed the privacy_guide_safe_browsing_fragment.js removals since those only existed due to the string removals which were removed in the last major version.

* remove-unused-preferences-fields had updates for [[4744021]](https://chromium-review.googlesource.com/c/chromium/src/+/4744021), [[4770014]](https://chromium-review.googlesource.com/c/chromium/src/+/4770014), [[4811081]](https://chromium-review.googlesource.com/c/chromium/src/+/4811081), and [[4672658]](https://chromium-review.googlesource.com/c/chromium/src/+/4672658). I also cleaned up a couple sections in download_item_model that were creating warnings previously.

* The Bromite measuretext fingerprinting patch needed some small adjustments for [[4780556]](https://chromium-review.googlesource.com/c/chromium/src/+/4780556)

* New additions to the remove-uneeded-ui patch to remove the non-functional store entry in the extension section of the main menu, and a fix for #2547

* Added a new flag for the unpack command:  `--skip-unused`
    This skips extracting directories that would be removed by prune_binaries.py

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/117.0.5938.150..118.0.5993.71?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2023/10/stable-channel-update-for-desktop_10.html)

PR #2552, opened 2023-10-12
Fix renderer SIGSEGV observed on certain website and profile combinations

madscientist159:
Rarely, with specific website and profile combinations, canvas can be NULL inside blink::BaseRenderingContext2D::measureText(), causing a SIGSEGV.

Check for canvas validity before attempting to dereference the canvas pointer.

Tested to work on Debian Bullseye ppc64el


PF4Public:
Please change `revision.txt` to `2`.

Ahrotahn (approved):
Looks good to me!

uazo:
https://github.com/ungoogled-software/ungoogled-chromium/pull/2552/files/a10f85f230cb80cb7fa5dc17860478e33f0632f5
^^^ may not be the right way

Please try this:
https://github.com/uazo/cromite/commit/002dc9166f20c34dc2ea87004712725e9e4bc3b0#diff-36022052f5651cc967d7f98fc0dde92075cbc05b3b670f26e9ec0832ae30a210L498

otherwise the anti-fingerprint measure may not be applied if javascript uses an offlinecanvas.
try with:
```
c = new OffscreenCanvas(500, 200).getContext('2d');
console.log(c.measureText("pippo"))
```

uazo:
> ** Currently debugging at assembler level due to oddities in the debug symbol 

@madscientist159 I am detecting, I think since v115, some problems myself with the blink side crash logs, which I can no longer get as easily as before.
Unfortunately, I haven't identified the cause yet, but I was wondering how you had [recovered the stack trace](https://github.com/ungoogled-software/ungoogled-chromium/pull/2548#issuecomment-1760222626)

PF4Public:
My concern was that the fact that this piece of code worked without errors for so long and suddenly failed in 118 could signify that we might just hit the tip of an iceberg, which means that we may have a deeper issue. 

I also compared our patches to Cromite/Bromite, found that they have this hunk in a different file, but also do not check canvas for null, so, I thought it would be a good idea to notify uazo about this issue. Oh boy how on-point that thought was :) @uazo thank you for your input.

uazo:
> My concern was that the fact that this piece of code worked without errors for so long and suddenly failed in 118 could signify that we might just hit the tip of an iceberg

the story was that previously neither ungoogled nor (b/c)romite foresaw that possibility, leaving open a door to fingerprinting, although minimised by the fact that fonts installed on the machine are not recoverable in cromite.
because of (thanks to :) https://chromium-review.googlesource.com/c/chromium/src/+/4791824 pointed out to me that that api is also available for offscreen canvas, which is why I have to decide to set up https://github.com/uazo/bromite-buildtools/issues/229 at some point (little time and too many things to do)
there are and may also be other patches to reanalyse, unfortunately I too have grown with time and only now understand (a little more :) the relationships between ExecutionContext, Frame and Document, which I ignored before. partly my fault, if I remember correctly I wrote that piece of code.

> I thought it would be a good idea to notify uazo about this issue

absolutely, don't ever make such problems, please.

PF4Public:
@uazo You do not check `canvas` validity in Cromite, could you try reproducing this issue on https://secure26ea.chase.com/web/auth/dashboard ? Do you observe SEGFAULT? I cannot build Cromite on Linux since 118 as my patches do not apply anymore and need updating (I didn't have time for that yet).

uazo:
>  could you try reproducing this issue 

no crashes, but you can also try https://abrahamjuliot.github.io/creepjs/tests/fonts.html

PF4Public:
> > could you try reproducing this issue
> 
> no crashes, but you can also try https://abrahamjuliot.github.io/creepjs/tests/fonts.html

This also crashes ungoogled-chromium for me :)

uazo:
> Would you be willing to allow us to use your solution?

Sure!

> It could also be simplified a bit:

i'm always undecided as to which method is best, so i'm doing different tests here and there on how to write code that fails in case of changes on the chromium team side.
not only that, I have realised that when a crash occurs, it is easier to work out which line is responsible, but these are all tests, I haven't yet decided how to always write it.

networkException:
Sorry I didn't quite follow the discussion, do we want to merge this?

PF4Public:
Better solution was in @Ahrotahn's message above, @Ahrotahn Would you be willing to submit a PR?

uazo:
> Better solution was in @Ahrotahn's message above,

for the record, technically it is the exact same thing writing with fewer bytes of text code :)

> credit @uazo for solution

personally, I don't care but of course do as you wish

PF4Public:
> for the record, technically it is the exact same thing writing with fewer bytes of text code :)

Sorry, @uazo, I didn't mean that _that code_ was better than _yours_! I meant that _that code_ was better than checking for `canvas` validity as suggested in present PR.

Ahrotahn:
If @madscientist159 wants to update this PR with those changes we can keep everything here.  I can make a separate PR if that's what you'd prefer though!

madscientist159:
> > ** Currently debugging at assembler level due to oddities in the debug symbol
> 
> @madscientist159 I am detecting, I think since v115, some problems myself with the blink side crash logs, which I can no longer get as easily as before. Unfortunately, I haven't identified the cause yet, but I was wondering how you had [recovered the stack trace](https://github.com/ungoogled-software/ungoogled-chromium/pull/2548#issuecomment-1760222626)

In my case, I take advantage of the fact that basically every function call on ppc64[el] stores the previous link register in the prior stack frame as an architectural ABI requirement (we have 32 GPRs available, not counting any vector or float registers).  It's pretty hard to get into a situation on ppc64 where you can't recover at least some kind of call chain, unless you're dealing with JITted or otherwise generated code where the addresses are mostly meaningless.

For x64 I'm pretty sure there's a compiler option to force something similar, but on x64 that will come at a performance penalty due to the lower register count, so it's only ever enabled on debug builds.  Ideally however we'd figure out why Chrome isn't respecting the `-g` flag as it would make debugging a lot easier...

bsdice:
+1 on enabling the craftier users to debug such issues easier. I hit this bug on nginx.com but all I saw at first was two lines of kernel messages with the SIGSEGV message. Nothing more on stdout or anywhere. Granted I have coredumps off via "kernel.core_pattern=|/bin/false".

I looked for basic tips and then tried gdb, but there was no output or backtrace whatsoever. I know this thing is a monster to compile and maintain, but we're likely stuck with this monoculture until something libre based on Rust/QuickJS comes out.

PF4Public:
I still get SEGFAULTs after rebuilding, but I noticed that
>chrome[499]: segfault at 20 ip 

Which suggests me that it is now not a NULL dereference, but a SPACE dereference :D

Maybe my local problem or something, does it work for anyone?

PF4Public (approved):
No more crashes!

Ahrotahn (approved):
Thanks y'all!

PR #2555, opened 2023-10-15
Exit on error for version checker workflow

### LINKED ISSUE #2544: GitHub Actions: New version script does not exit on error
Currently, it was decided that `jq -re` and `set -e` will be set to in order to exit on error. However, this does not work in subshells.

For example, you can try the following:

`exits_with_success.sh`:

```shell
#!/bin/bash
noop() { :;}
set -euo pipefail
noop $(false)
exit 0
```

In a shell,

```
$ bash exits_with_success.sh; echo Return value: $?
Return value: 0
```

rany2:
Closes #2544

### DIFF
--- a/.github/workflows/new_version_check.yml
+++ b/.github/workflows/new_version_check.yml
@@ -23,10 +23,14 @@ jobs:
       - name: Get the latest Chromium version
         id: latest-version
         run: |
-          set -e
-          echo "linux_version=$( curl -s 'https://versionhistory.googleapis.com/v1/chrome/platforms/linux/channels/stable/versions/all/releases?filter=endtime=none,fraction>=0.5&order_by=version%20desc'  | jq -re '.releases | first | .version' )" >> $GITHUB_OUTPUT
-          echo "win_version=$( curl -s 'https://versionhistory.googleapis.com/v1/chrome/platforms/win/channels/stable/versions/all/releases?filter=endtime=none,fraction>=0.5&order_by=version%20desc'  | jq -re '.releases | first | .version' )" >> $GITHUB_OUTPUT
-          echo "mac_version=$( curl -s 'https://versionhistory.googleapis.com/v1/chrome/platforms/mac/channels/stable/versions/all/releases?filter=endtime=none,fraction>=0.5&order_by=version%20desc'  | jq -re '.releases | first | .version' )" >> $GITHUB_OUTPUT
+          set -eo pipefail
+          BASE_URL="https://versionhistory.googleapis.com/v1/chrome/platforms"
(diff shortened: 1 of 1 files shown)


PF4Public:
So what happens if it's `null`?

rany2:
When it's `null` , jq won't return anything and will exit with failure causing the shell itself to exit with failure. Excluding null from the result is not actually needed as it will error out when the key is not found either way but I only left in case the version itself was set to null (no idea if it can happen but better be safe than sorry)

```
➜  ~ set -o pipefail
➜  ~ echo '{"releases": [{"version": null}]}' | jq -re '.releases | first | .version | select (.!=null)'
➜  ~ echo $?
4
➜  ~ echo '{"releases": [{"version": 1}]}' | jq -re '.releases | first | .version | select (.!=null)'
1
➜  ~ echo $?
0
➜  ~ echo '{"releases": [{"version": 1}]}' | jq -re '.releases | first | .versionNotExist | select (.!=null)'
➜  ~ echo $?
4
➜  ~ echo '{"releases": [{"version": 1}]}' | jq -re '.releases | first | .versionNotExist'
null
➜  ~ echo $?
1
```

Ahrotahn (approved):
Thanks @rany2!

PF4Public:
> When it's `null` , jq won't return anything and will exit with failure causing the shell itself to exit with failure.

@rany2 great job, but there is something that bothers me. Does it mean that it fails the whole workflow? If a recent situation, where it was 49% for one release, 49% for the other and the rest for the third one, would repeat, we get failed workflow notification every hour for the whole duration of this state. I'm concerned because at present it is me, who (still) receives them. ;) Maybe it would make sense to allow it to be null now that we have all "null" issues closed?

rany2:
@PF4Public I could change it such that it will only fail the workflow if it received an unexpected JSON. That way it shouldn't annoy you when the 49/49/1 condition occurs.

Right now this will indeed fail the whole workflow.

PF4Public:
> it will only fail the workflow if it received an unexpected JSON

That'd be great.

PF4Public:
This is awesome! You improved maintainability immensely!

Have you tested this last version in your fork to be sure it works as expected?

rany2:
Yes, it works.

PF4Public (dismissed):

networkException (dismissed):

rany2:
Sorry everyone for yet another commit, but jq works in some unexpected ways... I should have tested that case in particular, but felt like it goes without saying that `null | length` is an error, it's not :/

PF4Public:
@rany2 Good that you've caught that before we merged :)

PF4Public (approved):

Ahrotahn (approved):

PR #2560, opened 2023-10-17
Verbose patch validation

PF4Public:
This should make patch validation verbose, which allows someone not familiar with our workflows to infer what the issue with patches might be.


Ahrotahn (approved):
Seems like a good idea to me!

networkException (approved):

PR #2562, opened 2023-10-18
Update to Chromium 118.0.5993.88

Ahrotahn:
This PR updates to Chromium 118.0.5993.88.

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/118.0.5993.70..118.0.5993.88?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2023/10/stable-channel-update-for-desktop_17.html)


PF4Public:
Maybe we should also release .71 for Windows as containing the fix before .89 goes to stable on Windows?

Ahrotahn:
It seems like since .71 the trailing release is only for the Extended Stable channel now.  If you check out [the dash for windows](https://chromiumdash.appspot.com/releases?platform=Windows) .89 is also in Extended but not Stable.  Currently there is no tarball available for it, so maybe we shouldn't bother with those releases now?  They usually were only version bumps anyways.

PF4Public:
> there is no tarball available for it

Still no tarball? Shame.

We can hope that 89 comes to Windows fast enough.

PF4Public (approved):

teeminus:
Maybe we should switch to the git approach for windows to be able to release updates when no tarball is available.

PF4Public:
> Maybe we should switch to the git approach for windows to be able to release updates when no tarball is available.

Would it be doable w.r.t. changes in respective repo?

teeminus:
I tested the git method back when it was developed. It was not a big deal to change the windows build script. We could even keep the tarball method as a commandline option.

Unfortunately, I am currently out of town and won't be able to work on both this update and the git integration before next monday.

networkException (approved):

PR #2705, opened 2024-02-07
Update to Chromium 121.0.6167.160

### LINKED ISSUE #2693: Prevent displaying of feature advertisement labels in context menus
### Description

Hide promotional "New" labels in the context menu

### Who's implementing?

### The problem

Not only ridiculously dumb, annoying, and distracting since the only dismissible one is in the main menu (Password manager), but also misleading because there is nothing actually new; most of these features already existed for a long time in the browser or OS, and/or are already implemented by extensions, or even user scripts in a more convenient ways.

image

- "New" Password manager existed since 2008;

- "Open in reading mode" opens the stupid side panel instead of changing style of the page; DOM distiller existed since 2014;

- "Save video frame" is something for which Print Screen and OS screenshot key combinations exist.

Instead of educating ignorant users and computer noobs who don't want to take care of their needs themself, Chromium developers give up to this ignorance (or use it as inspiration when they're bored) and add bloat to the browser, then forcefully make everyone notice their masterpieces.

### Possible solutions

Remove "New" context menu labels and prevent possible other labels from appearing in context menus

### Alternatives

Turn off by default features that insert these "release notes" and advertisements in the functional UI, preserve flags that disable these features

### Additional context

Disabling #read-anything flag removes "Open in Reading mode" entry for now.

### LINKED ISSUE #2697: `chrome://safe-browsing` crashes ungoogled-chromium on Windows
### OS/Platform

Windows

### Installed

Other (please use 'Additional context' field below to mention how exactly)

### Version

121.0.6167.85（Official Build, ungoogled-chromium） 

- [X] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there

### Description

`chrome://safe-browsing` crashes ungoogled-chromium on Windows

### How to Reproduce?

1. Go to `chrome://safe-browsing`

### Actual behaviour

crashes ungoogled-chromium

### Expected behaviour

should not crash

### Relevant log output

_No response_

### Additional context

binary downloading from: 
https://github.com/ungoogled-software/ungoogled-chromium-windows/releases

Ahrotahn:
This PR updates to Chromium 121.0.6167.160, there were no notable changes needed on our end.  
I've also included a commit that removes the safebrowsing internal page (fixes #2697) and another that removes the new feature badges (fixes #2693).  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/121.0.6167.139..121.0.6167.160?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/02/stable-channel-update-for-desktop.html)


networkException (approved):

networkException:
> Is it automatically excluded from chrome://chrome-urls/? Because, you know, it would be tempting to click it :))))

Appears to be removed, yes

PF4Public (approved):
Awesome!

PR #2753, opened 2024-03-11
Upstream extension-search-path.patch

Pandapip1:
Upstreams the extension search path patch from https://github.com/NixOS/nixpkgs/pull/188086


PF4Public (changes requested):
Same considerations as voiced in nix issue tracker.

Besides, this patch does not apply.

networkException:
I avoided the nixpkgs issue largely because of its slightly convoluted history.

Given that I both maintain ungoogled-chromium and it's nixpkgs packaging (CC @emilylange), I'll try to take a closer look (can't promise when) at declarative extensions and whether ungoogled-chromium should put patches for it into its patchset

Pandapip1:
> Besides, this patch does not apply.

One of the stated goals of the `ungoogled-chromium` project is:

> ungoogled-chromium features tweaks to enhance privacy, control, and transparency. However, almost all of these features must be manually activated or enabled. For more details, see [Feature Overview](https://github.com/ungoogled-software/ungoogled-chromium/tree/master#feature-overview).

As this is a patch that enhances control, I'd argue this patch *does* apply.

networkException:
> Besides, this patch does not apply.

that is a technical observation:

```
WARNING: Patch failed validation: extra/ungoogled-chromium/extension-search-path.patch
DEBUG: Specifically, file "chrome/common/chrome_paths.cc" failed validation: Line '#include "base/files/file_path.h"' does not match context line '#include "base/files/file_util.h"' from patch
DEBUG: PATCH_BIN env variable is not set
DEBUG: Output of "patch --dry-run" for this patch on this file:
checking file chrome/common/chrome_paths.cc
Hunk #1 FAILED at 4.
Hunk #2 FAILED at 511.
2 out of 2 hunks FAILED
```

Pandapip1:
Sorry! Misinterpreted your statement.

Pandapip1:
The patch should apply now.

PF4Public:
> I'll try to take a closer look (can't promise when) at declarative extensions and whether ungoogled-chromium should put patches for it into its patchset

@networkException Have you had a chance to glance through? What's your opinion?

networkException:
*with my nixpkgs chromium maintainer hat on* I don't see the patch as useful here or in packaging as there are ways not needing modifications which I would prefer anyways

bendlas:
I would love to see this in UC. It's not just for including extensions from chromestore, but also for developing your own.

1. Carrying this patch in nixpkgs doesn't make sense, because it enables a substantial use case, which would then differ from other distros.
2. While there are workarounds, like managing the user's extention dir, this is substantially different and valuable in it's own right, because a system level extension also can't be messed up by a user messing up their browser profile.

Pandapip1:
Just fixed a merge conflict. Would anyone from UC be willing to review this?

networkException (changes requested):
No, this is not happening until it gets decided in upstream.

bendlas:
> No, this is not happening until it gets decided in nixpkgs upstream.

In what way is nixpkgs to be considered upstream to this?

What decision should be taken by nixpkgs and how would it influence UC?

Pandapip1:
I have also sent this patch to chromium upstream. It can be tracked here: https://chromium-review.googlesource.com/c/chromium/+/5650013

networkException:
I think this has been discussed enough, solutions have been pointed out. This is definitely not a fit for us and very likely not for nixpkgs either.

Closeing as per https://github.com/ungoogled-software/ungoogled-chromium/issues/2932#issuecomment-2212516130

PR #2755, opened 2024-03-12
Update to Chromium 122.0.6261.128

networkException:
This PR updates to Chromium 122.0.6261.128.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/122.0.6261.111..122.0.6261.128?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/03/stable-channel-update-for-desktop_12.html)


Ahrotahn:
Looks like there's some differences between the clone and tarball again, I'll spend some time looking into that.  I thought I had already added some of those...
The checks should pass without the list changes for now.

networkException:
Should I regenerate the list based on the tarball and we release that for now?

Ahrotahn:
I've sorted out what we'd need to change to get them synchronized, so you can regenerate the lists on the tarball and I'll add these changes to the 123 PR or if you're comfortable adding them to this PR:

```diff
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -102,7 +102,7 @@
     'net/http/transport_security_state_static.json',
     'net/http/transport_security_state_static_pins.json',
     # Exclusions for Visual Studio Project generation with GN (PR #445)
-    'tools/gn/src/gn/visual_studio_writer.cc',
+    'tools/gn/',
     # Exclusions for files covered with other patches/unnecessary
     'components/search_engines/prepopulated_engines.json',
     'third_party/blink/renderer/core/dom/document.cc',
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -17,6 +17,7 @@
 from subprocess import run
 
 from _common import add_common_params, get_chromium_version, get_logger
+from prune_binaries import CONTINGENT_PATHS
 
 # Config file for gclient
 # Instances of 'src' replaced with UC_OUT, which will be replaced with the output directory
@@ -226,6 +227,9 @@
             (args.output / 'v8' / 'test' / 'torque' / 'test-torque.tq'),
         )
         keep_suffix = ('.gn', '.gni', '.grd', '.gyp', '.isolate', '.pydeps')
+        # Include Contingent Paths
+        for cpath in CONTINGENT_PATHS:
+            remove_dirs += (args.output / Path(cpath),)
         for remove_dir in remove_dirs:
             for path in sorted(remove_dir.rglob('*'), key=lambda l: len(str(l)), reverse=True):
                 if path.is_file() and path not in keep_files and path.suffix not in keep_suffix:
--- a/utils/prune_binaries.py
+++ b/utils/prune_binaries.py
@@ -31,10 +31,11 @@
     'third_party/dawn/tools/golang/',
     'third_party/depot_tools/external_bin/',
     'third_party/devtools-frontend/src/third_party/esbuild/',
-    'third_party/google-java-format/'
+    'third_party/google-java-format/',
     'third_party/libei/',
     'third_party/llvm-build-tools/',
     'third_party/ninja/',
+    'third_party/screen-ai/',
     'third_party/siso/',
     'third_party/updater/chrome_linux64/',
     'third_party/updater/chromium_linux64/',
```

I forgot to add a comma when adding the google-java-format line which made an invalid path.  We already removed screen-ai in fix-building-with-prunned-binaries.patch due to building without tflite.  The contingent paths get pruned via the pruning script, but having the clone remove them as well means it should more closely match the tarball with --skip-unused.  And we already have an exemption for a path in tools/gn for windows, adding the whole path seems to be the more maintainable option for keeping them in sync.  

</details>

networkException:
Let's just add them here, can you push a commit for them onto some branch so that I can fast forward merge?

Ahrotahn:
Ok, I've just pushed the changes to [listsync](https://github.com/ungoogled-software/ungoogled-chromium/tree/listsync)

Ahrotahn (approved):
Looks good to me!

PF4Public (approved):

PF4Public:
Sorry for not participating in the discussion, I'm beyond of "occupied" offline right now.

utils/prune_binaries.py
```diff
@@ -31,10 +31,11 @@
     'third_party/dawn/tools/golang/',
     'third_party/depot_tools/external_bin/',
     'third_party/devtools-frontend/src/third_party/esbuild/',
-    'third_party/google-java-format/'
```
rany2, line 34: Shouldn't it error out if it failed to remove anything? To make it easier to detect errors like this? (It seems like this fixed an issue where it was trying to prune `third_party/google-java-format/third_party/libei/`)

Ahrotahn: The pruning list will error out when a path doesn't exist, so the contingent paths were set up specifically to be pruned only *if* they exist without throwing errors.  But you're right that there should at least be some sort of output that allows us to catch problems.  I'll experiment with adding info-level output for the contingent paths.

PR #2772, opened 2024-03-26
Update to Chromium 123.0.6312.86

networkException:
This PR updates to Chromium 123.0.6312.86.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/123.0.6312.58..123.0.6312.86?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/03/stable-channel-update-for-desktop_26.html)


Ahrotahn:
Builds and runs fine on my end!  I'll rerun the task once the tarball is available.  

They added an assignment operator that isn't supported until python 3.8+
`if val := GIT.rev_parse_cache.get(cache_key):` could be updated to
```python
val = GIT.rev_parse_cache.get(cache_key)
if val:
```
but I think this is a sign I need to spend some time updating the tasks to use a newer version of python.

Ahrotahn (approved):

clickot:
portable linux build works fine and result is running at least on Manjaro and ubuntu 20.04  upwards

teeminus:
Builds and runs fine for windows as well.

PF4Public (approved):

PR #2780, opened 2024-03-30
Add desktop, metainfo and screenshot files for Flatpak

rany2:
While I was in the process of submitting the Flatpak under the new app ID, a reviewer noted that these files should be stored in upstream repository. So we do that here.

The metainfo file should be updated as we release a new version; the first `<release>` tag in `<releases>` should be updated as we make changes to the app.

CC @networkException


rany2:
P.S. seeing that this is going be done, I'll try to get the screenshots adds as well.

rany2:
I guess this supersedes https://github.com/ungoogled-software/ungoogled-chromium/pull/1971

networkException:
hm I'm not sure I like having platform specific metadata here, is it possible to have all the flatpak specific things in a repository under this organization instead?

rany2:
@networkException I'll ask, I'm not sure about Flathub's current policies but what I've seen so far as that it is stored in the main repository.

rany2:
I really hate this too because it makes updating more complicated for me and we always risk forgetting to update the metainfo file... this sucks

rany2:
It appears that I'm allowed to do that.

PR #2819, opened 2024-04-23
[README] Remove openSUSE mention

Michal-Szczepaniak:
Unfortunately I've stopped maintaining ungoogled-chromium for opensuse as there are talks about opensuse not building chromium at all and frankly I've switched to firefox, though I still love what you guys are doing and support you all the way. 

For more info if you don't want to remove the entry, the devel package (package that's still built on obs but not in the standard repos) will be left alone 
https://build.opensuse.org/package/show/network:chromium/ungoogled-chromium
when someone wants to pick up maintenance they can do so.


Ahrotahn (approved):
Your time and effort is appreciated!  We can always add the link back if someone else steps up.

networkException (approved):

PR #2835, opened 2024-05-01
Update to Chromium 124.0.6367.118

Ahrotahn:
This PR updates to Chromium 124.0.6367.118.

Notes:
* If you've updated to ninja 1.12+ then you may need to apply the changes in [[5487538]](https://chromium-review.googlesource.com/c/chromium/src/+/5487538)
* The depot_tools patch has a new hunk to update the version of gsutil to one that supports python 3.12 (also tested with 3.9 & 3.11).  I've also included a new line to the clone script to silence syntax warnings since their scripts contain regex strings which cause the newer python to log new warnings.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/124.0.6367.91..124.0.6367.118?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/04/stable-channel-update-for-desktop_30.html)


clickot:
portable linux builds fine (when including dependency patch mentioned above) and result runs at least on amnjaro and ubuntu 20.04 upwards including 24.04

networkException (approved):

PR #2847, opened 2024-05-07
Update to Chromium 124.0.6367.155

Ahrotahn:
This PR updates to Chromium 124.0.6367.155.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/124.0.6367.118..124.0.6367.155?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/05/stable-channel-update-for-desktop_7.html)


networkException:
I'm getting

```
In file included from ../../chrome/browser/extensions/api/file_system/chrome_file_system_delegate.cc:16:
../../chrome/browser/download/chrome_download_manager_delegate.h:29:10: fatal error: 'components/safe_browsing/content/common/proto/download_file_types.pb.h' file not found
   29 | #include "components/safe_browsing/content/common/proto/download_file_types.pb.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

but only on arch so far

rany2:
@networkException might be an issue with your end, it works fine on Flatpak

Ahrotahn:
It seems like ninja 1.12 is causing some files to be generated out of order.  My initial build ran fine, but I ran into a similar issue with a generated buildflags.h file on another build I just did.  I ran it yet again and encountered the same thing for a different file elsewhere.  I guess the issue that [[5487538]](https://chromium-review.googlesource.com/c/chromium/src/+/5487538) fixed was only the most common occurrence.  I'll see what I can find out, maybe there's a way to work around that temporarily.

rany2:
Ah I see, that checks out. Freedesktop SDK is still at Ninja 1.11

clickot:
builds and runs fine for portable linux, result tested on manjaro and ubuntu 20.04/24.04

networkException (approved):
Thanks, that fixed the build on Arch!

teeminus:
Builds and runs fine on windows as well.

PR #2885, opened 2024-05-26
Fix for client certificates

Ahrotahn:
This PR reverts some changes for the client certificate store.  It seems that the changes that required these removals no longer exist or have had their safebrowsing sections gated since they were added.  This should fix https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/351

@Cubik65536 @teeminus let me know if this gives you any trouble building or linking since it also reverts the previous fix.


Cubik65536:
I will start testing this as soon as I got some spare time.

Cubik65536:
Builds fine with these patches removed on macOS.

teeminus:
Builds and runs fine on windows. Client certificates are working now. Thanks a lot for your efforts!

PF4Public (approved):

networkException (approved):
Sadly don't have a test setup for this but lgtm

Issue #2779, opened 2024-03-29
Add flag to hide URL scheme and subdomains


yokoffing:
### Description

Reenable URL hiding URL scheme

### Who's implementing?

- [ ] I'm willing to implement this feature myself

### The problem

There is no way to reverse “Disable automatic formatting of URLs in Omnibox (e.g. stripping http://, hiding certain parameters)”.

### Possible solutions

There were two flags for this before Chrome standardized them called `Omnibox UI Hide Steady-State URL Scheme` and `Omnibox UI Hide Steady-State URL Trivial Subdomains`. They’re still in Microsoft Edge under `edge://flags` and I use both of them.

Issue #2721, opened 2024-02-18
Site information dialog always show full URL instead of only domain (like Chrome)


kkocdko:
### OS/Platform and Version

Portable Linux 121.0.6167.184 https://ungoogled-software.github.io/ungoogled-chromium-binaries/

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [X] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [X] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [X] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### How to Reproduce?

Picture 1: Official Google Chrome.

Picture 2: Ungoogled Chromium. Both address bar "Always show full URLs" on/off.

Picture 3: Ungoogled Chromium with a very long URL. The under entries will be **out of screen and unable to access anymore**.

![Screenshot_2024-02-19_00-55-59](https://github.com/ungoogled-software/ungoogled-chromium/assets/31189892/7c924a70-978a-40ed-ae25-056e7760575d)

### Expected behaviour

Like Official Google Chrome.

PF4Public:
@kkocdko Have you attempted to reproduce this issue in the latest available version? Is this still an issue?

kkocdko:
@PF4Public Yes, this is still an issue.

Chromium 128 (google build) - Ungoogled Chromium 125

```
https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Win_x64%2F1315554%2Fchrome-win.zip?generation=1718425628787761&alt=media

https://github.com/ungoogled-software/ungoogled-chromium-windows/releases/download/125.0.6422.141-1.1/ungoogled-chromium_125.0.6422.141-1.1_windows_x64.zip

```

![screenshot-new](https://github.com/ungoogled-software/ungoogled-chromium/assets/31189892/86f92c0a-ecda-4fa9-933d-08c8986bfd08)

PF4Public:
I suspect `patches/extra/ungoogled-chromium/disable-formatting-in-omnibox.patch` might influence this. 

@Ahrotahn @networkException @teeminus @clickot @rany2 Is it important for us to fix this?

networkException:
I don't really mind this to be honest

Ahrotahn:
[...]

PR #2926, opened 2024-06-23
Various UI and history fixes

### LINKED ISSUE #1051: Remove patch for Omnibox URL formatting
In wchen342/ungoogled-chromium-android/issues/21 someone mentioned that in vanilla `chromium`, if `#enable-mark-http-as` is enabled then the protocal part (`https://`) will be hidden, while only `http://` will be shown as a warning. Supposedly it will be better this way and the reasoning is in the mentioned issue.

However in `ungoogled-chromium` there is a patch (I believe it is `disable-formatting-in-omnibox.patch`) explicitly disables that behavior. @Eloston I wonder why the patch is included in the first place? Is there a conclusion on which way is actually better?

### LINKED ISSUE #2899: History fails to import from Firefox when "Keep old history" flag is on
### OS/Platform

Windows

### Installed

OS/Platform's package manager

### Version

125.0.6422.141

- [X] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

History fails to import from Firefox when "Keep old history" flag is on

### How to Reproduce?

1. Enable "Keep old history" flag
2. Restart browser
3. Import history from Firefox

### Actual behaviour

Nothing is imported and no error is displayed

### Expected behaviour

All my history should be imported

### Relevant log output

_No response_

### Additional context

I have one year of history in Firefox, spanning 150000+ entries. 

If "Keep old history" is disabled, history is imported, but is truncated to three months.  Obviously, not reproducible on Google Chrome since it has no such flag.

Ahrotahn:
This PR contains a series of commits that address some existing issues:  
* The status info on the about page is removed.  This seems to have only manifested as an issue on mac but the underlying problem was mostly platform agnostic.  
  Fixes https://github.com/ungoogled-software/ungoogled-chromium-macos/issues/145
* The history submenu in the main menu had an entry that requires sign-in and has been removed.  
* Importing history with the keep-old-history flag enabled should now import all history.  
  Fixes #2899 
* The URL formatting patch has been overhauled to allow the omnibox toggle to function correctly and show only the domain info in the site info popup.  Showing the full URL info is still the default and can be changed by right-clicking the addressbar and toggling the "Always show full URLs" option.  I've also added a change to always show the full URL info in the status bubble.  
  Fixes #1051 #2779 #2721

This is also the first time the new changes in the close confirmation patch are being refreshed by quilt, no functional changes have been made.

### DIFF
--- a/patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
+++ b/patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
@@ -20,6 +20,7 @@
 # the 'Learn more' link from the search engine entry on the settings page
 # Safety Check entry on the side menu on the settings page
 # the (?) learn more button on many settings pages
+# 'Sign in to see tabs from other devices' in the history submenu
 # the 'Vist Chrome Web Store' entry in the extensions section of the main menu
 # the side panel entry in All Bookmarks
 # the feedback entry in the third party cookie popup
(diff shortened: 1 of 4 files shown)


patches/extra/ungoogled-chromium/remove-uneeded-ui.patch
```diff
@@ -152,6 +153,15 @@
        <template is="dom-if" if="[[!captionSettingsOpensExternally_]]">
 --- a/chrome/browser/resources/settings/about_page/about_page.html
 +++ b/chrome/browser/resources/settings/about_page/about_page.html
+@@ -64,7 +64,7 @@
+         </div>
+ </if>
+         <div class="flex cr-padded-text">
+-<if expr="not chromeos_ash">
++<if expr="False">
```
Cubik65536, line 162: Can confirm that this fixes the extra error message! 🎉

image

However, there's still an error icon, which I believe is at line 55 of `chrome/browser/resources/settings/about_page/about_page.html` you patched (the other `<if expr="not chromeos_ash">`). Maybe we should also make that `<if expr="False">`?

Cubik65536 (approved):

PF4Public:
@Ahrotahn Have you checked that `mailto:` links didn't regress with your changes?

Ahrotahn:
Yup!  Copying mailto links doesn't include the mailto prefix when I tested.

PF4Public (approved):

networkException (approved):

PR #2931, opened 2024-06-25
Update to Chromium 126.0.6478.126

networkException:
This PR updates to Chromium 126.0.6478.126.

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/126.0.6478.114..126.0.6478.126?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/06/stable-channel-update-for-desktop_24.html)


networkException:
I'll test #2926 before this, if all goes well and we can merge that I'll rebase the bump ofc

PF4Public (approved):

rany2:
Why doesn't this get merged seeing it's ready and then we make a new revision when that other patch is for sure ready? I feel like we should just make -2 revisions instead of trying to keep UGC and Chromium versions one-to-one?

Ahrotahn (approved):
Either way should be fine!

networkException:
Well sure, I'm not trying to keep chromium version one to one. Now the other pr is only blocked by my review though and I don't see an issue just doing both :)

PR #2959, opened 2024-07-26
Document OpenMandriva packaging

berolinux:
Document the availability of ungoogled-chromium in OpenMandriva main repositories


Ahrotahn (approved):
Neat!

networkException (approved):

networkException:
wohey, I like how the wording "the browser was built with ungoogled patches" matches really nicely here :)

PF4Public (approved):

PF4Public:
Is it worth noting, that they exclude some of the patches?
https://github.com/OpenMandrivaAssociation/chromium/blob/16b67fb137084e925f7942f27a40e1f84d0373d0/chromium.spec#L544-L552

Ahrotahn:
We had removed the cookie preference patch a while back so they only have two patches they are excluding:  The one that disables autofill by default and the one that removes the google search option.  

I feel like this is fine, especially since there are others that pick and choose patches as well.  Though it does bring up a good question about what changes could be made before it shouldn't be considered ungoogled anymore.

PR #2954, opened 2024-07-18
Update to Chromium 127.0.6533.72

Ahrotahn:
This PR updates to Chromium 127.0.6533.72.

### Changes:

* The DOH patch has been updated for the changes in [[5526034]](https://chromium-review.googlesource.com/c/chromium/src/+/5526034)

* The pruned binaries patch had a removal for [[5596349]](https://chromium-review.googlesource.com/c/chromium/src/+/5596349) and a new addition for [[5541202]](https://chromium-review.googlesource.com/c/chromium/src/+/5541202)

* The safebrowsing patch had changes for [[5562985]](https://chromium-review.googlesource.com/c/chromium/src/+/5562985), and new additions for [[5510834]](https://chromium-review.googlesource.com/c/chromium/src/+/5510834) & [[5447140]](https://chromium-review.googlesource.com/c/chromium/src/+/5447140)

* remove-unused-preferences-fields.patch had changes for [[5604230]](https://chromium-review.googlesource.com/c/chromium/src/+/5604230), a new addition for [[5594029]](https://chromium-review.googlesource.com/c/chromium/src/+/5594029), and a removal for [[5625866]](https://chromium-review.googlesource.com/c/chromium/src/+/5625866)

* fingerprinting-flags-client-rects-and-measuretext.patch needed the header to be readded due to [[5553003]](https://chromium-review.googlesource.com/c/chromium/src/+/5553003)

* The tabsearch flag patch had a removal for [[5532843]](https://chromium-review.googlesource.com/c/chromium/src/+/5532843).  This is for the windows-specific section, so it would nice to have verification that still works as expected.  

* The sidepanel flag had a removal for [[5550932]](https://chromium-review.googlesource.com/c/chromium/src/+/5550932)

* The prepopulated-search-engines patch has a new addition to remove the starter pack entries added a while ago in [[5283103]](https://chromium-review.googlesource.com/c/chromium/src/+/5283103).  This removes the non-functional `@Gemini` entry, fixes #2947

* `third_party/instrumented_libs` has been added to the contingent paths.  This path isn't needed for building and wasn't ever downloaded when using the clone script.  There are files in the path that probably should have been pruned and this should also prevent that focal build file from bouncing in and out of the substitution list.  

<br>

### Build notes:

* The `rust_bindgen_root` gn flag will now need to be set to a path containing a bin directory containing the program

[...]

PR #2961, opened 2024-07-28
Pruning and container changes for new DEPS inclusions

Ahrotahn:
This PR makes a few changes in response to the migration of more Chromium binaries to DEPS:

* The Cirrus download cache is now removed.  The container requires enough space to create a tar file of the cache directory in addition to what we already have in there.  We rarely have any cache hits in practice and the time required to create and upload the cache archive spends more time than it saves.  This should keep our container from running out of memory until the tarball + source grows another ~5GB.  
* The contingent paths have been updated to remove some obsolete paths and include all of the current GCS paths.  
* A new flag, `--sysroot`, has been added that allows specifying a sysroot to be kept/downloaded.  
  This is available for `utils/prune_binaries.py` which will skip pruning the path for the sysroot.  
  The flag is also available for `utils/downloads.py` which will extract the sysroot when used in conjunction with `--skip-unused`.  The Chromium tarballs only contain sysroots for amd64 and i386, so those are the only options available.  
  The cloning script has more sysroot architectures available which may be helpful for cross-compilation.  
* A new flag, `--keep-contingent-paths`, has been added to the pruning script which will skip pruning the contingent paths.  This should be useful for builds that require Google toolchain binaries while still pruning the rest.  
* I've also included a commit to add some more entries to the OS options for the github bug report template.


Ahrotahn:
Sorry for the pings, but I know this would affect some of your workflows.  

* Sysroots:
  @rany2 and any others using a sysroot.  Rather than leaving the sysroot paths intact but with missing binaries they will now be fully removed unless the `--sysroot` flag is supplied.  For example, passing `--sysroot amd64` to the clone/download script will download/extract the amd64 sysroot and passing the same flag to the pruning script will prevent it from being pruned.  
* `--keep-contingent-paths`:
  @alienbob This flag should hopefully maintain everything you need for your build.  This will also keep the sysroots in place.  
* Path changes:
  @teeminus @Cubik65536, The `third_party/llvm-build/Release+Asserts` path will now be pruned, so if that isn't automatically created by the build scripts then you may have to create the directory before using it.  
  @rany2 This should now remove the `third_party/llvm-build` as well.

teeminus:
Do I need the sysroot parameter?

```
INFO: Running gsync
  File "c:\ucw\build\src\uc_staging\depot_tools\gclient.py", line 825
    if not f'{self._get_option('sysroot', 'None')}-sysroot' in name:
                                ^^^^^^^
SyntaxError: f-string: unmatched '('
```

alienbob:
Thanks, I have pulled this PR and am building again with the new parameter. So far, so good (ninja step 6164 of 61217).

alienbob:
... and thanks for adding Slackware to the Bug report OS dropdown ;-)

rany2:
`--keep-contingent-paths` is very helpful, thanks!

teeminus:
Compilation finished without problems on windows.

networkException (approved):

Cubik65536:
Sorry for the late response, but the issue mentioned [here](https://github.com/ungoogled-software/ungoogled-chromium/pull/2954#issuecomment-2241453278) (`Unable to find libclang` issue) came back, the bindgen compiling script used to solve the problem was not touched, and I don't think that `third_party/llvm-build/Release+Asserts` is the issue because everything else works file as far as I can tell, what else could influence this?

For now, the `build_bindgen.py` is executed right after `./tools/gn/bootstrap/bootstrap.py -o out/Default/gn --skip-generate-buildfiles`, before `./out/Default/gn gen out/Default --fail-on-unused-args`.

Ahrotahn:
Were you using a clone or the tarball for the build when that happens?  Do you recall what you had tried setting `LIBCLANG_PATH` to when the problem happened previously?  Maybe that just needs to be set to a different path.  Do you also have llvm installed with brew?  I wonder if it's picking up the wrong one.

Cubik65536:
> Were you using a clone or the tarball for the build when that happens? Do you recall what you had tried setting `LIBCLANG_PATH` to when the problem happened previously? Maybe that just needs to be set to a different path. Do you also have llvm installed with brew? I wonder if it's picking up the wrong one.

I was using a clone when building the bindgen...

I didn't set `LIBCLANG_PATH` to anything at the end as setting it in terminal changes nothing and it was handled by the `build_bindgen` script. Still remains the same when I try to build now.

I'll check if it picked up the wrong LLVM later.

Cubik65536:
As far as I can tell, it used the right LLVM, and 127.0.6533.72 compiles just fine without any changes...

Ahrotahn:
Were you using `--keep-contingent-paths`?  Maybe something being pruned is causing the problem.

Cubik65536:
> Were you using `--keep-contingent-paths`? Maybe something being pruned is causing the problem.

No I was not, and 96d8e277e52503ca28bdbb34729be64723ece6ad is compiling fine right now with nothing changed, I'll just see if it fails at some point later, or it just magically fixed itself (maybe).

Cubik65536:
> Were you using `--keep-contingent-paths`? Maybe something being pruned is causing the problem.

I tried `--keep-contingent-paths` once but it was not working, however it seems that it is working now for some reason (compiling [38851/61580]), I'll keep watching it.

teeminus:
Maybe it's the same problem I experienced where compiling previously working releases again which the failed. My problem has been solved now without me changing anything. Maybe it's the same for you...

teeminus:
Nevermind, I just found out that I need an older version of GN I hardcoded to the clone script for test purposes.

Ahrotahn:
Oh interesting!  I hadn't considered that gn could be the issue.  It turns out that the latest commit was the culprit (in your case at least):
https://gn.googlesource.com/gn/+/0ee833e823f2e11be136728169906d0710bee910%5E%21/

They haven't updated the code to account for that change yet, but the revision in the current DEPS should work at least:  https://source.chromium.org/chromium/chromium/src/+/refs/tags/127.0.6533.88:DEPS;l=512

@Cubik65536 I hope we can eventually work out the paths that are most commonly needed so we can whitelist them.  That way most people won't have to set that flag.  

I'm about to go on a vacation in a couple days so I probably won't be able to dig into this much more until after next week, but I'll keep poking around to see what other info I can find out.

teeminus:
The .88 tarball uses gn version `b3a0bff47dd8`. Manually checking out that version fixed the windows build.

But there is an other issue with this PR. The checkout and the tarball now differs significantly which leaves some paths not being pruned (like the rust bin dir). This might cause issues on other platforms (like it did on windows as the python tried to execute the Linux rustc).

PR #2967, opened 2024-07-30
Update to Chromium 127.0.6533.88

networkException:
This PR updates to Chromium 127.0.6533.88.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/127.0.6533.72..127.0.6533.88?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/07/stable-channel-update-for-desktop_30.html)


PF4Public:
The question is, which PR goes first :)

PF4Public (approved):

Ahrotahn (approved):
All good on my build!

teeminus:
Did someone else has trouble building? I get the following `gn` error:

```
C:\ucw\build\src>"out\Default\gn.exe" "gen" "out\Default" "--fail-on-unused-args"
ERROR This tool specifies runtime_outputs.
This is only valid for linker tools (rust_staticlib doesn't count).
See //build/toolchain/win/toolchain.gni:672:5: whence it was called.
    msvc_toolchain(cl_toolchain_prefix + target_name) {
    ^--------------------------------------------------
See //build/toolchain/win/BUILD.gn:34:3: whence it was called.
  win_toolchains("x86") {
  ^----------------------
See //BUILD.gn:85:1: which caused the file to be included.
group("gn_all") {
^----------------
```

At first I thought it was due to the DEPS PR (which I tested and which did not fail), but I get the error now even when I try to recompile `127.0.6533.72` without the DEPS PR changes - which just compiled successfully in the CI. I have not yet tried compiling using the tarball but I would guess that this will work.

Ahrotahn:
Were you using the clone script for the builds that failed?  If so were you using `--keep-contingent-paths`?  Maybe there's something it's not downloading that should be.

teeminus:
I do use the clone script but I also tried to compile the previous release without the DEPS changes and it failed as well.

Ahrotahn:
One idea you could test is to skip the windows-specific pruning.  Maybe that list is removing something needed for the build.

teeminus:
We need the pruning, otherwise the additional downloads can't be extracted as some target files are (already) present.

I am currently comparing the tarball with the clone script, hoping that this will show what's wrong.

Ahrotahn:
`third_party/instrumented_libs` and `third_party/rust-toolchain` were two of the biggest differences I can recall.  The first one we had in .72 so maybe there's something needed in the rust-toolchain path.  You could try removing the line for that path in the pruning script to see if there's any difference.

teeminus:
It just fixed itself... While tinkering around I restarted compiling from scratch multiple times and now it's working without me changing anything...

teeminus:
Turns out I need an older version of GN to build successfully when using the clone script. I need to further investigate that problem later...

Issue #2963, opened 2024-07-28
Secure DNS fails on .works domain


Nifury:
When Secure DNS is turned on and set to use anything other than OS default, `.works` domain cannot be resolved and returns `ERR_NAME_NOT_RESOLVED`. 

Example: https://howhttps.works/

I'm not sure if this issue is specific to UG, but I cannot reproduce it on Chrome.

PF4Public:
@Ahrotahn Could you reproduce this issue on your system?

rany2:
It might have nothing to do with `.works`. I could resolve `19216811.works` for example, maybe the fact that the domain contains `https` is the problem?

Ahrotahn:
I can reproduce this.  It only seems to be an issue for howhttps.works, other .works addresses function normally like the one @rany2 mentioned.  

Even weirder, if you visit howhttps.works once without secure DNS enabled then the site will work fine afterwards once it's re-enabled.  This works on other pages on the domain and with a hard refresh so it isn't just a cached version of the page.

PF4Public:
> Even weirder, if you visit howhttps.works once without secure DNS enabled then the site will work fine afterwards once it's re-enabled. This works on other pages on the domain and with a hard refresh so it isn't just a cached version of the page.

Same for me.

networkException:
maybe something to do with HSTS?

alienbob:
I just upgraded to my fresh Un-Googled Chromium 127 on Slackware, switched on Secure DNS and accessed https://howhttps.works/ .
It works, No errors or issues and I get to see the page - and I honestly never visited this URL before, did not knw of its existence.
Examining the network interactions (which I did activate via Developer Tools before accessing this URL for the first time) I only see loading error for https://tag.getdrip.com/9686301.js due to `Referrer Policy: "strict-origin-when-cross-origin"`.

rany2:
@alienbob Did you access it without secure DNS beforehand? See https://github.com/ungoogled-software/ungoogled-chromium/issues/2963#issuecomment-2256554151

alienbob:
> @alienbob Did you access it without secure DNS beforehand? See [#2963 (comment)](https://github.com/ungoogled-software/ungoogled-chromium/issues/2963#issuecomment-2256554151)

Nope. As I said (but pashaps not explicitly enough) , I had never even heard of the site, and turned on Secure DNS **before** accessing that URL for the first time.

Ahrotahn:
[...]

PR #2970, opened 2024-08-01
DNS fixes, clone patch update, and config cleanup

### LINKED ISSUE #2093: Not able to put in a custom DOH address
### OS/Platform

Arch Linux

### Installed

OS/Platform's package manager

### Version

105.0.5195.125

- [ ] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile and it could not be reproduced there

### Description

Issue Adding A Custom DOH Service Provider

### How to Reproduce?

Naviagte to: Privacy and Security => Security Advanced => Use secure DNS => With Custom

Copy or manually add a custom DOH service provider

### Actual behaviour

Tried to add Quad9 to Chromium and I am getting an error.

When I use https://dns.quad9.net/dns-query I get the error message, "**Please verify that this is a valid provider or try again later**".

This is also happens with other DOH service providers like BlahDNS and UncensoredDNS.

### Expected behaviour

After adding a DOH servcie provider I should be able to hit enter for Chromium to accept it.

### Relevant log output

_No response_

### Additional context

_No response_

Ahrotahn:
This PR contains a handful of smaller fixes:
*  The `0009-disable-google-ipv6-probes.patch` patch has been removed since `add-ipv6-probing-option.patch` removes that change.  
* When inputting a custom secure DNS address a probe is run to check if it is valid.  This would fail due to the checked address being a google domain that gets substituted.  This changes that domain to `example.com` so the check can succeed.  Fixes #2093 
* There is a sysroot in third_party/dawn that is allowed to be downloaded with the clone script when the sysroot flag is passed since it follows the same pattern as the regular build sysroots.  This change makes a more specific check to be sure only the intended one is downloaded.  
* The Deepsource config has been removed.  This is just a cleanup since we haven't used this in years.


PF4Public:
> * Fixes [Secure DNS fails on .works domain #2963](https://github.com/ungoogled-software/ungoogled-chromium/issues/2963)

But that wasn't the problem! the domain itself is!

rany2:
I noticed this about howhttps.works' AAAA responses, it returns a SERVFAIL. It seems like a DNSSEC issue on their end:

```
~ $ dig howhttps.works aaaa

; <<>> DiG 9.16.41 <<>> howhttps.works aaaa
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: SERVFAIL, id: 22456
;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
; EDE: 12 (NSEC Missing): (Invalid denial of existence of howhttps.works/aaaa)
;; QUESTION SECTION:
;howhttps.works.                        IN      AAAA

;; Query time: 92 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Thu Aug 01 18:12:18 EEST 2024
;; MSG SIZE  rcvd: 99

~ $ dig nsa.gov aaaa

; <<>> DiG 9.16.41 <<>> nsa.gov aaaa
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 58717
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;nsa.gov.                       IN      AAAA

;; AUTHORITY SECTION:
nsa.gov.                1800    IN      SOA     w000-dsid-u05.nsa.gov. hostmaster.nsa.gov. 2024030527 3600 1080 2419200 900

;; Query time: 108 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Thu Aug 01 18:12:33 EEST 2024
;; MSG SIZE  rcvd: 97
```

PF4Public:
> it returns a SERVFAIL. It seems like a DNSSEC issue on their end

It is exactly what I wrote: https://github.com/ungoogled-software/ungoogled-chromium/issues/2963#issuecomment-2259120412

Ahrotahn:
Yup!  The core issue is some kind of misconfiguration on the site's end and that was exposed by our changes.  

But it's also true that vanilla Chromium has no issue visiting the site, and it's also true that the patch causes the probe to always return true when the flag is not set.  It should only be changing the value to false when the flag is set unless the goal is to remove the probe entirely.  

If that is the case then we should look at removing `patches/core/inox-patchset/0009-disable-google-ipv6-probes.patch` since `add-ipv6-probing-option.patch` currently removes that change entirely.

PF4Public:
> unless the goal is to remove the probe entirely.

I thought that was the idea behind this change. May we have following options here: "IPv4", "IPv6", "probes enabled / default behaviour", and the last one if it would make any sense: "default behaviour, but without any probes"?

Ahrotahn:
Sorry for the delay, I just returned from a trip a day ago.  

I can remove the inox patch and we can keep things how they are since it would only affect misconfigured sites.  If that site is the only one that's been found since the patch has been in place then I don't think it's too big of an issue.  

Another option that keeps the probe removed would be to update the patch to have the result return false by default and change the flag to `SetIpv6ProbeTrue`.  This should allow misconfigured sites to function correctly like in vanilla chromium while not changing behavior for anyone that still has `SetIpv6ProbeFalse` set.  The only difference is that people on IPv6-only networks might have to set the the flag to function correctly.

PF4Public:
I'm just trying to wrap my head around it and cannot make sense of it :(

> […] to have the result return _false_ by _default_ and change the flag to SetIpv6Probe _True_. […] for anyone that still has SetIpv6Probe _False_ _set_.

It easily gets sooooo confusing. I suggest we change the name of the option to `IPv6Reachability`. With the following description: `Forces the result of the browser's IPv6 probing (i.e. IPv6 connectivity test) to be successful without causing additional network activity. This causes IPv6 addresses to be prioritized over IPv4 addresses. With this flag disabled, the probing result is set to be unsuccessful, which causes IPv4 to be used over IPv6 when possible. ungoogled-chromium flag.`. Enabled by default. (I think, it is what @Ahrotahn suggested, I'm just trying to rephrase it and make the statement more clear).

Why do we keep this chunk? Maybe remove it too?
```c
  // Don't bother checking if the request will use WiFi and IPv6 is assumed to
  // not work on WiFi.
  if (!check_ipv6_on_wifi_ && RequestWillUseWiFi(target_network_)) {
    probing_ipv6_ = false;
    last_ipv6_probe_result_ = false;
    last_ipv6_probe_time_ = base::TimeTicks();
    return OK;
  }
```

Ahrotahn:
Yeah, the whole function would be changed to the inside of that chunk and `last_ipv6_probe_result_` would be set to the state of the flag.  

For now I've updated the PR to keep the patch as it was but remove the redundant patch.  I guess if we encounter more sites that cause problems we could change the flag, but I think it might cause too much confusion if we don't have stronger reasons to make that change.

PF4Public (approved):
lgtm

networkException (approved):

rany2:
Can you make a -2 release?

Ahrotahn:
Sure, I'll push the tag in a bit

PR #2997, opened 2024-08-23
Disable DownloadsPageReferrerUrl feature by default

MikeWang000000:
Disable DownloadsPageReferrerUrl feature by default

This patch will disable `DownloadsPageReferrerUrl` feature by default.

The `DownloadsPageReferrerUrl` feature is introduced by https://chromium-review.googlesource.com/c/chromium/src/+/5589888 .

> [Downloads] Show referrer url on chrome://downloads items
> 
> This swaps the existing download url on a download item with the referrer url. Clicking the
> link will open the download's referrer/origin in a new tab. It's wrapped behind a feature flag.

Since chromium 127, users are unable to get real download url from `chrome://downloads`.

For example, download this file using chromium: https://commondatastorage.googleapis.com/chromium-browser-official/chromium-128.0.6613.84.tar.xz

The `chrome://downloads` page only shows its referrer. Click "Copy Link Address" in the menu will get its referrer `https://github.com/...` instead of real download url `https://commondatastorage.googleapis.com/...`.

If we copy this link and paste to omnibox to download (without referrer), the url part will be completely empty.

I don't think it is a meaningful feature. It makes user more difficult to get the real url (especially the download link has redirects). Also, the original url will be lost if we download mannually by url. Disable it so chromium's `chrome://downloads` will behave the same as before version 127.

### DIFF
--- a/patches/extra/ungoogled-chromium/disable-downloads-page-referrer-url.patch
+++ b/patches/extra/ungoogled-chromium/disable-downloads-page-referrer-url.patch
@@ -1,0 +1,11 @@
+--- a/components/safe_browsing/core/common/features.cc
++++ b/components/safe_browsing/core/common/features.cc
+@@ -205,7 +205,7 @@ BASE_FEATURE(kHashPrefixRealTimeLookupsFasterOhttpKeyRotation,
+ 
+ BASE_FEATURE(kDownloadsPageReferrerUrl,
+              "DownloadsPageReferrerUrl",
+-             base::FEATURE_ENABLED_BY_DEFAULT);
++             base::FEATURE_DISABLED_BY_DEFAULT);
+ 
+ BASE_FEATURE(kLogAccountEnhancedProtectionStateInProtegoPings,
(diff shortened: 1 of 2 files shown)


PF4Public:
Is it platform independent?

MikeWang000000:
> Is it platform independent?

Yes, it is platform independent.

PF4Public:
Disabling referrer makes URL part be gone right from the first attempt. :)

MikeWang000000:
btw, `chrome --disable-features=DownloadsPageReferrerUrl` can be a quick fix. This patch makes this option the default.

I can verify that after applying this patch, we can see the url part with `chrome://flags/#minimal-referrers` enabled.

Ahrotahn (approved):
Seems fine to me and works as expected in my build.

PF4Public (approved):

networkException:
Nice patch, thanks!

I'd say we just wait for the next upstream bump to include this

Issue #2974, opened 2024-08-02
Browser doesn't correctly enforce the DownloadRestrictions policy


thapabishwa:
### OS/Platform

Debian, Ubuntu, and derivatives

### Installed

OS/Platform's package manager

### Version

Version 126.0.6478.182 (Official Build, ungoogled-chromium) built on Ubuntu , running on Ubuntu 24.04 (64-bit)

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [X] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [X] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [ ] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

Browser doesn't correctly enforce the DownloadRestrictions policy

### How to Reproduce?

1. Set following policy to Block All Downloads 
`/etc/chromium/policies/managed/test_policy.json`
```
{
    "DownloadRestrictions": 3
}
```

2. Navigate to [file-example-download](https://file-examples.com/wp-content/storage/2017/02/file-sample_500kB.doc)

### Actual behaviour

The downloaded is not blocked.
- policy recommended
![image](https://github.com/user-attachments/assets/9501ade4-d684-45c4-9217-a0e8416dcb1e)
- policy managed
![image](https://github.com/user-attachments/assets/00b550bc-f650-4e5e-9dc3-88955803b218)

### Expected behaviour

The downloaded is blocked.

PR #3000, opened 2024-08-27
Enable blocking all downloads via managed policy

berkley4:
In a nutshell, this commit allows users to set the DownloadRestrictions managed policy key to either 0 or 3 to respectively permit or block all downloads.

The [key values](https://chromeenterprise.google/policies/#DownloadRestrictions) are as follows.....

0 = No special restrictions. Default.
1 = Block malicious downloads and dangerous file types.
2 = Block malicious downloads, uncommon or unwanted downloads and dangerous file types.
3 = Block all downloads.
4 = Block malicious downloads. Recommended.

In this commit, DownloadRestrictions values of 1, 2 and 4 are effectively neutered by disentangling them from safebrowsing and treating them as invalid. None of the reinstated code appears to be connected to safebrowsing.

The relevant section of chrome/browser/download/chrome_download_manager_delegate.cc will end up looking like this.....

```
bool ChromeDownloadManagerDelegate::ShouldBlockFile(
    download::DownloadItem* item,
    download::DownloadDangerType danger_type) const {
  DownloadPrefs::DownloadRestriction download_restriction =
      download_prefs_->download_restriction();

  switch (download_restriction) {
    case (DownloadPrefs::DownloadRestriction::NONE):
      return false;

    case (DownloadPrefs::DownloadRestriction::ALL_FILES):
      return true;

    // DownloadRestrictions policy key values 1, 2 and 4 treated as invalid
    case (DownloadPrefs::DownloadRestriction::POTENTIALLY_DANGEROUS_FILES):
    case (DownloadPrefs::DownloadRestriction::DANGEROUS_FILES):
    case (DownloadPrefs::DownloadRestriction::MALICIOUS_FILES):
    default:
      LOG(ERROR) << "Invalid download restriction value: "
                 << static_cast<int>(download_restriction);
  }

  return false;
}
```

I have tested DownloadRestrictions values of 0-5, with 0 and 3 behaving as expected. Running without a policy file also behaves as expected, same as 0 (ie there is no intrinsic change in behaviour from this commit).

Values 1, 2 and 4 behave the same as 5, emitting a non-fatal "Invalid download restriction value" error and allowing the download to proceed. 

This PR will fix issue #2974.

### DIFF
--- a/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
+++ b/patches/core/ungoogled-chromium/fix-building-without-safebrowsing.patch
@@ -250,7 +250,7 @@
    }
    if (ShouldBlockFile(item, target_info.danger_type)) {
      MaybeReportDangerousDownloadBlocked(
-@@ -1767,53 +1756,6 @@ bool ChromeDownloadManagerDelegate::IsOp
+@@ -1767,49 +1756,20 @@ bool ChromeDownloadManagerDelegate::IsOp
  bool ChromeDownloadManagerDelegate::ShouldBlockFile(
      download::DownloadItem* item,
      download::DownloadDangerType danger_type) const {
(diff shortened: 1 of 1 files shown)


Ahrotahn (approved):
Thanks @berkley4, works as expected:
Screenshot_20240827

PF4Public:
So it is either allow downloads (by default), or block all of them?

berkley4:
> So it is either allow downloads (by default), or block all of them?

That's correct.

PF4Public (approved):

PR #3009, opened 2024-08-30
Fix builds with DCHECK enabled

### LINKED ISSUE #3008: Build fails when DCHECK is enabled
### OS/Platform

Gentoo

### Installed

Compiled from source

### Version

128.0.6613.84

### Description

Preference members that are otherwise patched out are still referenced in places compiled only when DCHECK is enabled.

### How to Reproduce?

Attempt to build ungoogled-chromium with GN flags `is_debug=true` and `dcheck_always_on=true`.

### Actual behaviour

The build fails in a portion guarded by `#if DCHECK_IS_ON()` due to references to members that are otherwise pacthed out.

### Expected behaviour

The build completes without errors even when DCHECK is enabled.

### Relevant log output

```sh
../../components/signin/internal/identity_manager/primary_account_manager.cc:400:47: error: no member named 'kGoogleServicesAccountId' in namespace 'prefs'
  400 |         client_->GetPrefs()->GetString(prefs::kGoogleServicesAccountId);
      |                                        ~~~~~~~^
../../components/signin/internal/identity_manager/primary_account_manager.cc:402:48: error: no member named 'kGoogleServicesConsentedToSync' in namespace 'prefs'
  402 |         client_->GetPrefs()->GetBoolean(prefs::kGoogleServicesConsentedToSync);
      |                                         ~~~~~~~^
```

### Additional context

The errors can be fixed simply by adding this chunk to the `components/signin/internal/identity_manager/primary_account_manager.cc` section of patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch:
```diff
@@ -527,10 +399,9 @@ void PrimaryAccountManager::SetSyncPrimaryAccountInternal(
 
 #if DCHECK_IS_ON()
   {
-    std::string pref_account_id =
-        client_->GetPrefs()->GetString(prefs::kGoogleServicesAccountId);
+    std::string pref_account_id;
     bool consented_to_sync =
-        client_->GetPrefs()->GetBoolean(prefs::kGoogleServicesConsentedToSync);
+        false;
 
     DCHECK(pref_account_id.empty() || !consented_to_sync ||
            pref_account_id == account_info.account_id.ToString())
```

In #2034 it was concluded that this problem should be left unfixed and higher symbol levels used for debugging instead of DCHECK. Arguably, however:

 1. A DCHECK failure with its resulting backtrace can make for a much swifter diagnosis of a problem than loads of trace output.
 2. As seen in pf4public/gentoo-overlay#365, a trivially small fix like the above looks to be all it takes to build with DCHECKs enabled. This feels like a very small price to pay for avoiding guaranteed recurring showstoppers in some build configurations.

In light of the above, would you be willing to reconsider adding the fix to your patches?

skrinakron:
Removes stranded uses of preferences members patched out elsewhere. As the only occurrences appear to be in portions compiled only when DCHECK is enabled, this allows full debug builds to now succeed. Fixes #3008.

### DIFF
--- a/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
+++ b/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
@@ -6558,7 +6558,20 @@
      } else if (ShouldSigninAllowedPrefAffectPrimaryAccount(
                     pref_consented_to_sync)) {
        SetPrimaryAccountInternal(CoreAccountInfo(), /*consented_to_sync=*/false,
-@@ -545,10 +417,6 @@ void PrimaryAccountManager::SetSyncPrima
+@@ -527,10 +399,9 @@ void PrimaryAccountManager::SetSyncPrima
+ 
+ #if DCHECK_IS_ON()
+   {
+-    std::string pref_account_id =
+-        client_->GetPrefs()->GetString(prefs::kGoogleServicesAccountId);
++    std::string pref_account_id;
+     bool consented_to_sync =
+-        client_->GetPrefs()->GetBoolean(prefs::kGoogleServicesConsentedToSync);
++        false;
+ 
+     DCHECK(pref_account_id.empty() || !consented_to_sync ||
+            pref_account_id == account_info.account_id.ToString())
+@@ -545,10 +416,6 @@ void PrimaryAccountManager::SetSyncPrima
    // Go ahead and update the last signed in account info here as well. Once a
    // user is signed in the corresponding preferences should match. Doing it here
    // as opposed to on signin allows us to catch the upgrade scenario.
@@ -6569,7 +6582,7 @@
  }
  
  void PrimaryAccountManager::SetPrimaryAccountInternal(
-@@ -560,22 +428,6 @@ void PrimaryAccountManager::SetPrimaryAc
+@@ -560,22 +427,6 @@ void PrimaryAccountManager::SetPrimaryAc
    // 'account_info' might be a reference to the contents of `primary_account_`.
    // Create a PrimaryAccount object before calling emplace to avoid crashes.
    primary_account_.emplace(PrimaryAccount(account_info, consented_to_sync));
@@ -6592,7 +6605,7 @@
  }
  
  void PrimaryAccountManager::RecordHadPreviousSyncAccount() const {
-@@ -585,7 +437,7 @@ void PrimaryAccountManager::RecordHadPre
+@@ -585,7 +436,7 @@ void PrimaryAccountManager::RecordHadPre
    }
  
    const std::string& last_gaia_id_with_sync_enabled =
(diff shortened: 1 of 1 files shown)


PF4Public:
Interesting, the whole patch fails to apply.

Ahrotahn:
The offsets just need to be updated since there's one fewer line now:

```diff
--- a/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
+++ b/patches/core/ungoogled-chromium/remove-unused-preferences-fields.patch
@@ -6571,7 +6571,7 @@
  
      DCHECK(pref_account_id.empty() || !consented_to_sync ||
             pref_account_id == account_info.account_id.ToString())
-@@ -545,10 +417,6 @@ void PrimaryAccountManager::SetSyncPrima
+@@ -545,10 +416,6 @@ void PrimaryAccountManager::SetSyncPrima
    // Go ahead and update the last signed in account info here as well. Once a
    // user is signed in the corresponding preferences should match. Doing it here
    // as opposed to on signin allows us to catch the upgrade scenario.
@@ -6582,7 +6582,7 @@
  }
  
  void PrimaryAccountManager::SetPrimaryAccountInternal(
-@@ -560,22 +428,6 @@ void PrimaryAccountManager::SetPrimaryAc
+@@ -560,22 +427,6 @@ void PrimaryAccountManager::SetPrimaryAc
    // 'account_info' might be a reference to the contents of `primary_account_`.
    // Create a PrimaryAccount object before calling emplace to avoid crashes.
    primary_account_.emplace(PrimaryAccount(account_info, consented_to_sync));
@@ -6605,7 +6605,7 @@
  }
  
  void PrimaryAccountManager::RecordHadPreviousSyncAccount() const {
-@@ -585,7 +437,7 @@ void PrimaryAccountManager::RecordHadPre
+@@ -585,7 +436,7 @@ void PrimaryAccountManager::RecordHadPre
    }
  
    const std::string& last_gaia_id_with_sync_enabled =
@@ -6614,7 +6614,7 @@
    const bool existed_primary_account_with_sync =
        !last_gaia_id_with_sync_enabled.empty();
  
-@@ -735,38 +587,6 @@ PrimaryAccountChangeEvent::State Primary
+@@ -735,38 +586,6 @@ PrimaryAccountChangeEvent::State Primary
  void PrimaryAccountManager::ComputeExplicitBrowserSignin(
      const PrimaryAccountChangeEvent& event_details,
      ScopedPrefCommit& scoped_pref_commit) {
@@ -6653,7 +6653,7 @@
  }
  
  void PrimaryAccountManager::FirePrimaryAccountChanged(
-@@ -850,7 +670,6 @@ void PrimaryAccountManager::OnSigninAllo
+@@ -850,7 +669,6 @@ void PrimaryAccountManager::OnSigninAllo
  bool PrimaryAccountManager::ShouldSigninAllowedPrefAffectPrimaryAccount(
      bool is_sync_consent) {
    return switches::IsExplicitBrowserSigninUIOnDesktopEnabled() &&
```
</details>

skrinakron:
Sorry and thanks for the correction. I did wonder about the line offset but got a false sense of security when `patch` gave neither errors nor rejections. Using verify_patches.py I now see the errors as well. I've amended the commit with the above changes.

Do you happen to know if there's a way to make `patch` as fastidious about matches as verify_patches.py is? I've tried the options `-N -F0 --posix --verbose -r /known/path/rejects_file`, and yet the faulty patch keeps applying without so much as a warning.

Ahrotahn:
I'm not aware of any way to make `patch` fail when encountering offset differences, but you could instead use `git apply` as a replacement which will fail if the patch can't be applied cleanly (it also defaults to `-p1`).  

Granted most of our patches wouldn't apply cleanly without many of the of previous patches in the series already applied.  You would want to use `quilt` to make refreshing the patches for this repo easier.

Ahrotahn (approved):

PF4Public (approved):

PR #3027, opened 2024-09-11
Update to Chromium 128.0.6613.137

Ahrotahn:
This PR updates to Chromium 128.0.6613.137
This also includes a small fix to the RemoveClientHints patch which returns an empty object instead of null for userAgentData since some extensions don't expect that to be null.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/128.0.6613.119..128.0.6613.137?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/09/stable-channel-update-for-desktop_10.html)


clickot:
portable linux builds fine and runs at least on Manjaro, Ubuntu 20.04/24.04 and Fedora WS 40

networkException:
can do

patches/extra/ungoogled-chromium/add-flag-to-remove-client-hints.patch
```diff
@@ -59,12 +59,12 @@
...
- namespace blink {
- 
+@@ -14,6 +15,8 @@ namespace blink {
  NavigatorUAData* NavigatorUA::userAgentData() {
-+  if (base::FeatureList::IsEnabled(blink::features::kRemoveClientHints))
-+    return nullptr;
    NavigatorUAData* ua_data =
        MakeGarbageCollected<NavigatorUAData>(GetUAExecutionContext());
++  if (base::FeatureList::IsEnabled(blink::features::kRemoveClientHints))
++    return ua_data;
```
networkException, line 67: This condition looks inverted in reference to the previous iteration of the patch

Ahrotahn: ua_data is a newly initialized object which gets filled in below that section, so everything still returns empty:
Screenshot

PF4Public (approved):

networkException (approved):

PR #3034, opened 2024-09-18
Enhancements to downloads.py and filescfg.py

iskunk:
This is a byproduct of my work on the [Windows cross build](https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/293).

### utils/downloads.py

* Add a `--components` option to allow downloading a subset of the items listed in the `downloads.ini` file(s). This is not a big deal in the main `ungoogled-chromium` repo, but in `ungoogled-chromium-windows`, the file lists 13 items and I need to download just one.

  The terminology is, admittedly, a bit mixed-up here. The existing code refers to these items as "sections," as they are indeed sections of the `.ini` file. This is followed in the new method `check_sections_exist()`, that guards against mistypes. But the sections represent things that you can download, which I have termed "components." They could also be called "packages," but that may be confused with system-installable packages. Perhaps not "dependencies," as `[chromium]` can't be called one. Let me know if an update to the wording is desired.

* For the `--7z-path` option, use a default of `_use_registry` only on Windows. On other platforms, use `7z`. This not only gives expected behavior on Linux, it avoids an ugly exception from trying to read a non-existent Windows Registry.

  (I did not apply this to `--winrar-path`, since WinRAR does not exist outside of Windows, though this could conceivably be extended to support the command-line version of RAR.)

### utils/filescfg.py

* When creating an archive, allow specifying a timestamp that will be applied to all member files. This makes the resulting archive reproducible, since the variability of timestamps on files from the preceding build no longer affect the output.

  Note that this does not fully address `.tar.gz` archives, as the gzip compression layer has an embedded timestamp that [cannot be set via the `tarfile` API](https://github.com/python/cpython/issues/75707). That said, `.tar.xz` archives are fine.

* Allow specifying to `filescfg_generator()` a set of files to exclude from the iteration it produces. This is intended as an improvement to the current practice (as seen in `u-c-windows/package.py`) of deleting files to exclude them from the iteration, which can lead to non-idempotent behavior.

The changes to this script will be used in a PR update to `package.py` that I will submit shortly.


iskunk:
I fixed most of the issues flagged by **code_check**, save for the "Too many arguments" on `unpack_downloads()`. Not sure what can be done there, without more invasive refactoring.

Ahrotahn:
We'll need to break up `unpack_downloads()` into smaller parts eventually, but for now you could bump the max-args to seven [here](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/devutils/run_utils_pylint.py#L34).

iskunk:
> We'll need to break up `unpack_downloads()` into smaller parts eventually, but for now you could bump the max-args to seven [here](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/devutils/run_utils_pylint.py#L34).

Thanks, I've included that edit in my commit. (I tried to see if I could get away with a comment directive, but couldn't get pylint working locally to try it out.)

Ahrotahn (approved):
Thanks @iskunk, looks good here

PF4Public (approved):
If it's good it is good :)

iskunk:
Thanks everyone! This clears the way for ungoogled-software/ungoogled-chromium-windows#383.

PR #3045, opened 2024-10-01
Improvements in hash checking in downloads.py

lifesminder:
Previous iteration checked whole file and caused memory issues, making programs non-responding. This one improves it, by reading chunks by bytes (default is 8 megabytes).

Tested in Fedora 40.


Ahrotahn:
This is a good idea considering how large the archives are getting, but there's a couple changes that would need to be made:  
* There is trailing whitespace on the three empty lines you've added.  That is why one of the tests has failed and that would need to be removed.  
* The buffer size should be increased.  The common knowledge of using 4k/8k chunks is somewhat misapplied in situations where the file is always going to be large.  Python added the helper function `file_digest` which isn't available to us with the version we use, but they have chosen a size of 2<sup>18</sup> (262144) which I think we should use instead and will make that take less time to process:  https://github.com/python/cpython/blob/1f4a49ea53516e7ff177beedc09a1e4439b3be1f/Lib/hashlib.py#L195

lifesminder:
> This is a good idea considering how large the archives are getting, but there's a couple changes that would need to be made:
> 
> * There is trailing whitespace on the three empty lines you've added.  That is why one of the tests has failed and that would need to be removed.
> * The buffer size should be increased.  The common knowledge of using 4k/8k chunks is somewhat misapplied in situations where the file is always going to be large.  Python added the helper function `file_digest` which isn't available to us with the version we use, but they have chosen a size of 218 (262144) which I think we should use instead and will make that take less time to process:  https://github.com/python/cpython/blob/1f4a49ea53516e7ff177beedc09a1e4439b3be1f/Lib/hashlib.py#L195

cool, for these purposes I have added an extra `chunk_bytes` argument, in order to make it scalable and set required size in specific scenarios.

about whitespace I can't say much, as I don't have an experience on python 2.x versions programming :)

lifesminder (commented):
minor fixes

utils/downloads.py
```diff
@@ -286,27 +286,32 @@ def retrieve_downloads(download_info, cache_dir, show_progress, disable_ssl_veri
...
+                # Read file in chunks. Default is 8 megabytes
+                chunk = file_obj.read(chunk_bytes)
+                while chunk:
+                    hasher.update(chunk)
+                    chunk = file_obj.read(chunk_bytes)
             if not hasher.hexdigest().lower() == hash_hex.lower():
                 raise HashMismatchError(download_path)
 
-
 def unpack_downloads(download_info, cache_dir, output_dir, skip_unused, sysroot, extractors=None):
```
rany2, line 315: Should fix code_check CI issue:

```suggestion

def unpack_downloads(download_info, cache_dir, output_dir, skip_unused, sysroot, extractors=None):
```

rany2 (changes requested):

Ahrotahn:
There's a different problem this time:  There needs to be two line breaks before `unpack_downloads`.  @rany2's suggestion should fix that.  

The problem you had previously wasn't with the empty lines, it was that those lines *weren't* empty.  They contained space characters.  

Since your commits are signed by Github it seems you're using their editor, and unfortunately that doesn't have the option to show whitespace characters.  If you happen to use a userscript extension there is a script that can be used to add that functionality:  https://github.com/Mottie/GitHub-userscripts/wiki/GitHub-code-show-whitespace

lifesminder:
> Minor thing: you keep saying the default is 8 megabytes in the code and PR, but it's actually 8192 bytes (8 KiB). Anyway as stated in [#3045 (comment)](https://github.com/ungoogled-software/ungoogled-chromium/pull/3045#issuecomment-2385931844), the default is too low and should be increased.

oh okay, which size should it be?

utils/downloads.py
```diff
@@ -286,23 +286,29 @@ def retrieve_downloads(download_info, cache_dir, show_progress, disable_ssl_veri
                                 disable_ssl_verification)
 
 
-def check_downloads(download_info, cache_dir):
+def check_downloads(download_info, cache_dir, chunk_bytes=8192):
```
rany2, line 289:
```suggestion
def check_downloads(download_info, cache_dir, chunk_bytes=262144):
```

rany2 (changes requested):

utils/downloads.py
```diff
@@ -286,23 +286,29 @@ def retrieve_downloads(download_info, cache_dir, show_progress, disable_ssl_veri
...
         download_path = cache_dir / download_properties.download_filename
-        with download_path.open('rb') as file_obj:
-            archive_data = file_obj.read()
         for hash_name, hash_hex in _get_hash_pairs(download_properties, cache_dir):
-            get_logger().debug('Verifying %s hash...', hash_name)
-            hasher = hashlib.new(hash_name, data=archive_data)
+            logger.info('Verifying %s hash...', hash_name)
+            hasher = hashlib.new(hash_name)
+            with download_path.open('rb') as file_obj:
+                # Read file in chunks. Default is 8 megabytes
```
rany2, line 307:
```suggestion
                # Read file in chunks. Default is 262144 bytes.
```

rany2 (changes requested):

lifesminder:
hi, is code reviewed yet?

rany2:
You marked my suggestions as resolved without making any changes

lifesminder:
> You marked my suggestions as resolved without making any changes

so how to put them into commit? It's my first time to make a pull request, sorry for too much questions :3

Ahrotahn:
There should be a button to add the suggestion as a commit.  There has also been a change we recently merged that causes a conflict with these changes, so you would also need to use the resolve conflicts button to handle that.  

Alternatively I could take some time later to rebase the changes and add those suggestions for you so long as your branch allows us to push to it.

PF4Public:
I think we'll have to squash commits before merging to tidy it up a bit.

utils/downloads.py
```diff
@@ -301,26 +301,33 @@ def retrieve_downloads(download_info,
...
 
     Raises source_retrieval.HashMismatchError when the computed and expected hashes do not match.
     """
+    logger = get_logger()
     for download_name, download_properties in download_info.properties_iter():
         if components and not download_name in components:
             continue
         get_logger().info('Verifying hashes for "%s" ...', download_name)
+        
         download_path = cache_dir / download_properties.download_filename
```
Ahrotahn, line 321: One last set of spaces:  
```suggestion
        get_logger().info('Verifying hashes for "%s" ...', download_name)

        download_path = cache_dir / download_properties.download_filename
```

Ahrotahn:
@PF4Public The 'squash and merge' option should handle that for us

PF4Public:
> @PF4Public The 'squash and merge' option should handle that for us

yep, I meant just that

Ahrotahn (approved):
Thanks @lifesminder!

networkException (approved):

PF4Public (approved):
I trust you guys with python :)

lifesminder:
> Thanks @lifesminder!

no problem :Ճ

PR #3050, opened 2024-10-04
Remove the flag that eventually turned to be a no-op

PF4Public:
This flag became a no-op since the last chunk was removed in 37452608f8406c6d8b5f7fa3f27fa69a905109d2. It actually does nothing.


Ahrotahn (approved):
Thanks @PF4Public!

PF4Public:
@networkException ?

networkException (approved):

PR #3057, opened 2024-10-11
Update to Chromium 130.0.6723.58

### LINKED ISSUE #1130: Remove Sync (ie: "Not Signed In", "Turn on sync...", "Sync and Google services", etc) UI

Ahrotahn:
This PR updates to Chromium 130.0.6723.58

Notes:

* A new binary exclusion has been added for bindings_wasm_bg.wasm which was added in [#5073](https://www.github.com/rollup/rollup/pull/5073) for version 4.0 of rollup.  This is included in the node_modules download so I'm not sure if there is a better means of handling that.  

* The doh-changes patch had a section moved to a new area in [[5546423]](https://chromium-review.googlesource.com/c/chromium/src/+/5546423)

* The safebrowsing patch had changes for [[5791595]](https://chromium-review.googlesource.com/c/chromium/src/+/5791595)

* remove-unused-preferences-fields.patch had changes for [[5823396]](https://chromium-review.googlesource.com/c/chromium/src/+/5823396)

* The MDNS patch had a new addition for [[5828629]](https://chromium-review.googlesource.com/c/chromium/src/+/5828629).  For some reason they added the the dep twice and only gated the second instance.  

* remove-uneeded-ui.patch had a removal for [[5855283]](https://chromium-review.googlesource.com/c/chromium/src/+/5855283) and a small change to an existing section to fix #1130.  

* A new patch has been added to the upstream fixes directory.  They introduced new changes in [[5767325]](https://chromium-review.googlesource.com/c/chromium/src/+/5767325) using `std::hardware_destructive_interference_size` which requires a newer version of LLVM and then partially reverted the change in [[5841144]](https://chromium-review.googlesource.com/c/chromium/src/+/5841144).  The new patch reverts the remaining change.  

* A new addition to the missing dependencies patch for a usage of `base::Contains` that no longer has the include supplied transitively.  Their own code browser isn't able to find a reference so I'm not sure how that is building for them.  

* The change in [[5850578]](https://chromium-review.googlesource.com/c/chromium/src/+/5850578) causes implicit conversion warnings.  If the spam causes problems for your build then `-Wno-shorten-64-to-32` might be necessary.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/129.0.6668.100..130.0.6723.58?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2024/10/stable-channel-update-for-desktop_15.html)


Ahrotahn:
I've pushed an update to `add-flag-to-close-window-with-last-tab.patch` moving the new tab creation ahead of the reentrancy checks.  They started enforcing those with 130.  

<hr>

Also there's a fix for the GBM log spam when playing videos on wayland in [[5873796]](https://chromium-review.googlesource.com/c/chromium/src/+/5873796).  It probably won't land until 131, but applies cleanly on 130.

clickot:
portable linux builds and runs fine

teeminus:
.44 builds and runs fine on windows as well.

clickot:
404 on chromium-130.0.6723.58.tar.xz download. Not yet published?

Ahrotahn:
The tarball for .58 is not yet available since their builder seems to be having issues at the moment:  https://ci.chromium.org/ui/p/infra/builders/cron/publish_tarball

I'm not sure how long until that gets resolved, so I guess this is only really ready for clones at the moment.

PF4Public:
Should we be waiting for a tarball?

Ahrotahn:
The tarball is available now.  
Hopefully they can get that cleared up before next week...

networkException (approved):

clickot:
.58 builds and runs fine on portable linux

Cubik65536:
Builds fine on macOS (Apple Silicon).

PF4Public:
```diff
-                                         net::SecureDnsMode::kAutomatic)));
+                                         net::SecureDnsMode::kOff)));
```
Are we setting it to "off" because it'll try to connect google servers in "automatic"?

Ahrotahn:
The Google servers are also removed in that patch so it shouldn't use them by default.  I think the setting was disabled by default due to the patch allowing DoH to work on linux back when that wasn't supported upstream and we didn't want to override the system configuration.  

We could remove that now that most people are aware of the setting.  Although I've still seen some people surprised to find that bypasses their Pi-hole setup or can route their traffic to Cloudflare for example.

PF4Public:
Well, it should be safer with "off" then

PF4Public (approved):

PR #3071, opened 2024-10-26
when pruning files log missing file paths together with the error message

clickot:
in case of missing files to prune the prune_binary.py should log their paths together with the error message to ease fixing this. Until now these paths were only logged when debug level is enabled.

### DIFF
--- a/utils/prune_binaries.py
+++ b/utils/prune_binaries.py
@@ -128,9 +128,9 @@ def _callback(args):
     prune_list = tuple(filter(len, args.pruning_list.read_text(encoding=ENCODING).splitlines()))
     unremovable_files = prune_files(args.directory, prune_list)
     if unremovable_files:
-        file_list='n'.join(f for f in itertools.islice(unremovable_files, 5))
+        file_list = 'n'.join(f for f in itertools.islice(unremovable_files, 5))
         if len(unremovable_files) > 5:
-            file_list+='\n... and ' + (len(unremovable_files) - 5) + ' more'
+            file_list += '\n... and ' + str(len(unremovable_files) - 5) + ' more'
         get_logger().error('%d files could not be pruned:\n%s', len(unremovable_files), file_list)
         sys.exit(1)
 
--- a/utils/prune_binaries.py
+++ b/utils/prune_binaries.py
@@ -129,7 +129,7 @@ def _callback(args):
     unremovable_files = prune_files(args.directory, prune_list)
     if unremovable_files:
         get_logger().error('%d files could not be pruned:\n%s', len(unremovable_files), 
-        '\n'.join(f for f in unremovable_files))
+                           '\n'.join(f for f in unremovable_files))
         sys.exit(1)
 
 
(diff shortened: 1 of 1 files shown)


PF4Public:
;)

https://github.com/ungoogled-software/ungoogled-chromium/pull/2954#issuecomment-2246230907
https://github.com/ungoogled-software/ungoogled-chromium/pull/2954#issuecomment-2246666744

clickot:
i don't understand what cirrus is complaining about the code formatting, original and reformatted code looks equal to me... 
@PF4Public  i'll double check if this PR is really needed

Ahrotahn:
@PF4Public My main concern at the time was there were 76 files that would be listed.  

When there are a lot files that will be listed you can end up in a situation where that list is longer than your terminal buffer, which pushed off the original error and leaves you with less helpful information.  Maybe we could instead list the first half-dozen files with a following message saying something like "plus X more files" if there's more than that.  

@clickot There's some whitespace at the end of the first line.  I couldn't see the difference at first either.

PF4Public:
> Maybe we could instead list the first half-dozen files with a following message saying something like "plus X more files" if there's more than that.

That is a good idea! @clickot Would you be able to add this condition?

clickot:
done as requested :)
btw: after @Ahrotahn s question in #3069 i noticed that the node_modules.tar.gz file exists after runing the clone script on it's own. so i restarted my build from scratch without removing the line from the pruning list and things went fine.
So the reason for this PR has disappeared in a way... 
but nonetheless it might be useful in future cases :-)

PF4Public:
Would it then make sense to keep listing all of the files with debug flag?

> but nonetheless it might be useful in future cases :-)

absolutely! the original idea was even improved :)

clickot:
> Would it then make sense to keep listing all of the files with debug flag?
good idea, i'll add that line

utils/prune_binaries.py
```diff
@@ -128,9 +129,12 @@ def _callback(args):
     prune_list = tuple(filter(len, args.pruning_list.read_text(encoding=ENCODING).splitlines()))
     unremovable_files = prune_files(args.directory, prune_list)
     if unremovable_files:
-        get_logger().error('%d files could not be pruned.', len(unremovable_files))
-        get_logger().debug('Files could not be pruned:\n%s',
-                           '\n'.join(f for f in unremovable_files))
+        file_list = 'n'.join(f for f in itertools.islice(unremovable_files, 5))
```
Ahrotahn, line 132: The first newline is missing the backslash:  
```suggestion
        file_list = '\n'.join(f for f in itertools.islice(unremovable_files, 5))
```
<br>
Everything else looks good to me and works as expected.

clickot: thanks, fixing this immediately

Ahrotahn (approved):
Thanks @clickot!

PF4Public (approved):

clickot:
@PF4Public @Ahrotahn since i have no write access, would one of you be so kind and merge this?

networkException (approved):

PR #3072, opened 2024-10-26
Enable extra locales not normally shipped with Chromium on desktop

PF4Public:
Google does not ship some of the locales on desktop by default, we can fix that :)
Tested on Linux:
image

It'd be great if @Cubik65536 could test the patch on Mac and @Nifury or @teeminus on Windows.

### DIFF
--- a/patches/extra/ungoogled-chromium/enable-extra-locales.patch
+++ b/patches/extra/ungoogled-chromium/enable-extra-locales.patch
@@ -1,0 +1,112 @@
+# Enable extra locales not normally shipped with Chromium on desktop
+
+--- a/build/config/locales.gni
++++ b/build/config/locales.gni
+@@ -166,12 +166,6 @@
+ 
+ # Setup |platform_pak_locales| for each platform.
+ platform_pak_locales = all_chrome_locales
+-if (is_android) {
+-  platform_pak_locales -= _non_android_locales
+-  extended_locales -= _non_android_locales
+-} else {
+-  platform_pak_locales -= extended_locales
+-}
+ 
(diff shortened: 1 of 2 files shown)


Cubik65536:
So just to make sure, we're testing if `Sinhalese (si)` is enabled after building with this patch?

PF4Public:
> So just to make sure, we're testing if `Sinhalese (si)` (and similar disabled by default locales) is enabled after building with this patch?

Exactly.

Cubik65536:
CleanShot 2024-10-27 at 01 30 14

CleanShot 2024-10-27 at 01 31 13

Works on macOS!

Nifury:
image

Works on Windows as well!

Ahrotahn (approved):
All good here, thanks @PF4Public!

networkException (approved):

clickot:
@PF4Public: layman question: Doesn't it depend on what fonts are installed on a linux system if such languages are properly displayed? I ask because I struggled with Farsi (not Arabic!) the get properly displayed in browsers on Linux.

PF4Public:
> Or does this PR only concern the display language of the browser itself, not websites?

Only the UI language.



> Doesn't it depend on what fonts are installed on a linux system. […] I struggled with Farsi to get properly displayed in browsers on Linux

In simple cases yes. I got used to have Noto on my machines to cover maximum glyphs possible. So if one of the fonts does not have needed glyphs browsers substitute from Noto. But your case is something else. I couldn't even think of browsers confusing writing systems. :-/ Have you tried to install the whole Noto family? Have you tried searching skia/harfbuzz bugtracker? It is very improbable that no one else noticed.

PR #3079, opened 2024-11-01
Manually set up gsutil path for newer versions of Python

Ahrotahn:
This PR adds a new section to the clone script that manually sets up the gsutil directory with a newer version of apitools so Python 3.13+ can be used.  

Normally gsutil creates this path during runtime and makes an `install.flag` file to signify completion so it only needs to do the work once.  This change writes to the same file in order to prevent gsutil from overwriting our changes.

### DIFF
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -47,13 +47,15 @@ target_cpu_only = True;
 """
 
 
-def clone(args): # pylint: disable=too-many-branches, too-many-statements
+def clone(args): # pylint: disable=too-many-branches, too-many-locals, too-many-statements
     """Clones, downloads, and generates the required sources"""
     get_logger().info('Setting up cloning environment')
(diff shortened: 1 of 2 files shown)


Ahrotahn:
@tsuliwaensis @teeminus @clickot Let me know if that gives you any trouble.  This method is more involved but should hopefully avoid the platform-specific pitfalls.

tsuliwaensis:
I was able to run the clone script with Python 3.13

teeminus:
No issues on windows 👍

clickot:
Additionally to the sysroot feature I use a debian-bullseye based docker image for the build to ensure that the build result runs on "older" linux verions. in this docker image is a python 3.9.2. 
In this setting this PR works :+1:

PF4Public (approved):

networkException (approved):

PR #3108, opened 2024-11-23
Remove client hints from frame fetch when flag is enabled

### LINKED ISSUE #3105: Remove Client Hints flag only works on the first request
### OS/Platform

Windows

### Installed

https://ungoogled-software.github.io/ungoogled-chromium-binaries/

### Version

131

### Description

Browser sends sec-ch-ua headers even if the #remove-client-hints flag enabled

### How to Reproduce?

1. Enable `#remove-client-hints` flag
1. Open any site that returns request headers
1. Make sure that the sec-ch-ua headers have not been sent
1. Send the same request to the site in devtools console using fetch
1. Get the sec-ch-ua headers in the response

Or just open any other site and check the network tab in devtools

### Actual behaviour

Browser sends sec-ch-ua headers even if the `#remove-client-hints` flag enabled

### Expected behaviour

Browser doesn't send sec-ch-ua headers even if the `#remove-client-hints` flag enabled

### Relevant log output

_No response_

### Additional context

image
image

Ahrotahn:
This PR updates the client hints patch to account for a separate section that was missed for fetches.  
Fixes #3105

Screenshot1

### DIFF
--- a/patches/extra/ungoogled-chromium/add-flag-to-remove-client-hints.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-to-remove-client-hints.patch
@@ -68,3 +68,22 @@
  
    UserAgentMetadata metadata = GetUserAgentMetadata();
    ua_data->SetBrandVersionList(metadata.brand_version_list);
+--- a/third_party/blink/renderer/core/loader/frame_fetch_context.cc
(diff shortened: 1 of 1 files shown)


networkException (approved):

PF4Public (approved):
Such a simple fix 👍🏻

PR #3143, opened 2025-01-09
Update to Chromium 132.0.6834.83

Ahrotahn:
This PR updates to Chromium 132.0.6834.83

Notes:

* The rollup files are now duplicated for devtools-frontend, so another exclusion for the same file has been added to update_lists.py

* The Inox safebrowsing patch had a small update for the changes in [[5923123]](https://chromium-review.googlesource.com/c/chromium/src/+/5923123)

* There was a new addition to the trk patch for [[5964077]](https://chromium-review.googlesource.com/c/chromium/src/+/5964077)

* disable-google-host-detection.patch had changes do to [[5970653]](https://chromium-review.googlesource.com/c/chromium/src/+/5970653)

* The pruned binaries patch had new additions for [[5934367]](https://chromium-review.googlesource.com/c/chromium/src/+/5934367), [[5998631]](https://chromium-review.googlesource.com/c/chromium/src/+/5998631), and [[5917173]](https://chromium-review.googlesource.com/c/chromium/src/+/5917173)

* The safebrowsing patch had changes for [[5989132]](https://chromium-review.googlesource.com/c/chromium/src/+/5989132), [[6001419]](https://chromium-review.googlesource.com/c/chromium/src/+/6001419), [[5939832]](https://chromium-review.googlesource.com/c/chromium/src/+/5939832), [[6006096]](https://chromium-review.googlesource.com/c/chromium/src/+/6006096), [[6019823]](https://chromium-review.googlesource.com/c/chromium/src/+/6019823), [[5958668]](https://chromium-review.googlesource.com/c/chromium/src/+/5958668), and [[5990352]](https://chromium-review.googlesource.com/c/chromium/src/+/5990352)

* remove-unused-preferences-fields.patch had removals for [[5985667]](https://chromium-review.googlesource.com/c/chromium/src/+/5985667) and [[5963058]](https://chromium-review.googlesource.com/c/chromium/src/+/5963058), changes for [[5983026]](https://chromium-review.googlesource.com/c/chromium/src/+/5983026), and new additions for [[5998292]](https://chromium-review.googlesource.com/c/chromium/src/+/5998292) and [[6000926]](https://chromium-review.googlesource.com/c/chromium/src/+/6000926)

* disable-capture-all-screens.patch had changes for [[5998751]](https://chromium-review.googlesource.com/c/chromium/src/+/5998751)

[...]

PR #3197, opened 2025-02-18
Switch to the upstream's lite source tarball

iskunk:
In recent weeks, I've submitted a number of updates to Google's tarball-generation scripts to make their lite source tarball usable for typical Chromium builds, as well as reduce its size further, and clean up other aspects of it (e.g. file permissions).

The lite tarball for 133.0.6943.98 is now one-fifth the size of the standard one (1.36 GB vs. 6.90 GB). @Ahrotahn [confirmed](https://github.com/ungoogled-software/ungoogled-chromium/pull/3143#issuecomment-2585317200) being able to build u-c for Windows using it---I've done so as well, in my cross build---and over in Debian-land, we [switched to it](https://salsa.debian.org/chromium-team/chromium/-/blob/master/debian/rules?ref_type=heads#L213) already.

If folks using tarballs for their builds can confirm that the lite variant works for them, this will save quite a bit of bandwidth. (And if there are any issues with missing files, I'll be happy to look into them, and potentially prepare another upstream fix.)


clickot:
@iskunk do have to take some action to use the light source tarball? When i start my build, it still downloads a tar of 6-7G size

iskunk:
Hi @clickot,

The main change is to the `downloads.ini` file. Do you see `lite.tar.xz` in there?

Also, I see the CI failed at the pruning step. You may need to remove from `pruning.list` the lines starting with `ios/` and `third_party/closure_compiler/`. This is a problem that has come up before; better solution to come...

clickot:
@iskunk when applying your changes to `download.ini` and patching the `pruning.list` the portable linux build works :-)
Thanks for the change, it shortens the build time and resources a lot !

Ahrotahn:
Yeah there were some contingent paths that needed to be removed after I tested with the lite tarball in order to keep things working with the current system.  I hope to have a PR out soon that should allow this one to clear.

PF4Public:
I'm building Gentoo with lite tarballs without issues for quite some time. I suspect Android would be impossible to build with it, but Android repo is not active anyway.

Ahrotahn:
You've got the general idea, there's just a lot of changes that will need to be made to some of the other scripts to make everything work correctly.  I have those changes ready but I'll wait until after .126 to make a PR.

iskunk:
> You've got the general idea, there's just a lot of changes that will need to be made to some of the other scripts to make everything work correctly. I have those changes ready but I'll wait until after .126 to make a PR.

Ah, I wasn't aware you had worked on this! Would you say your changes are a superset of mine? I can reduce the scope of this PR, or close it altogether if that'll help ease the way.

I see that the "prune all files except for the GN/GRIT bits" approach is a bit hard to reconcile with `--skip-unused`, since GNU Tar doesn't seem to support what would be needed...

Ahrotahn:
They were changes I had been working on in order to move the contingent paths to a list, but they are also a prerequisite for using the lite tarball.  Your original commit was fine, it just won't pass the checks until those changes are made.  

The skip-unused flag was added to help reduce CI space requirements when the tarballs were getting larger.  It basically pre-pruned those paths by simply not extracting them.  The work you've done on the lite tarballs has made that flag unnecessary since those changes have already been done for us.

PF4Public:
> I'm building Gentoo with lite tarballs without issues for quite some time. I suspect Android would be impossible to build with it, but Android repo is not active anyway.

FYI I didn't mention that I had to haphazardly fix the pruning script to make it work with lite tarball because this problem was already being discussed here.

> If you can try an Android build (even if not ungoogled), and determine/confirm for me which files need to be included

In case it is a "time-limited offer", I'm afraid I'm not in a position to do it quickly.

> I've been working on [a small side project](https://github.com/chromium-linux-tarballs/chromium-tarballs) with Matt Jolly. This monitors Google for Chromium version bumps, and builds new stable source tarballs ASAP.

Is it the script that generates tarballs that Gentoo actually uses? I remember I tried those tarballs, but they were lacking something, breaking the build, so I dropped it.

iskunk:
> FYI I didn't mention that I had to haphazardly fix the pruning script to make it work with lite tarball because this problem was already being discussed here.

And I'd made the same fixes so long ago that I forgot they were necessary (in my initial PR commit) :-]

> > If you can try an Android build (even if not ungoogled), and determine/confirm for me which files need to be included
> 
> In case it is a "time-limited offer", I'm afraid I'm not in a position to do it quickly.

I see. I ask because I'm preparing what will hopefully be a final update to the tarball scripts---I've already pushed a handful, and don't want to wear out my welcome with the reviewers at Google.

There's no immediate urgency on that update, however, just some final clean-up. I can hold off for a time. Would be good if we can do a similar exercise for macOS, too, as that might also be supportable without much extra weight...

> Is it the script that generates tarballs that Gentoo actually uses? I remember I tried those tarballs, but they were lacking something, breaking the build, so I dropped it.

I don't know if the Gentoo build directly uses this tarball (cf. Debian, which makes their own ~800 MB tarball from the upstream one with a bunch of stuff they don't need cut out), but the goal is for this tarball to exactly match the lite tarball rather than being its own thing. And I've been specifically going after build breakage due to the lite tarball missing necessary files.

iskunk:
All right, the changes I have here are now subsumed into #3202, which is passing CI checks. Thanks @Ahrotahn for all the heavy lifting getting the scripts in order.

And thanks for your support everyone! I'm glad to see the lite tarball cleanup work finally bear fruit in this project. Less time spent watching `tar xJf` commands is always nice ^_^

PF4Public:
> I see. I ask because I'm preparing what will hopefully be a final update to the tarball scripts---I've already pushed a handful, and don't want to wear out my welcome with the reviewers at Google.
> 
> There's no immediate urgency on that update, however, just some final clean-up. I can hold off for a time.

In that case do not wait for me, just submit what you have when you're ready. I'm not expecting anything change favourably for me soon enough.

iskunk:
> In that case do not wait for me, just submit what you have when you're ready. I'm not expecting anything change favourably for me soon enough.

Ah, understood :-(  If you know anyone who might be up for it, or if anyone reading this happens to be so, feel free to let me know...

PR #3202, opened 2025-02-19
Consolidate contingent paths in prune_binaries.py

Ahrotahn:
In the past we have only needed to deal with the differences between the full tarball and cloned sources.  The lite tarballs are significantly smaller now with the changes @iskunk has made, so now we need to be able to reconcile the differences between all three sources.  

This PR moves the contingent paths logic from the cloning script into the pruning script so all three sources are treated equally.  

This also removes the `--skip-unused` flag from `downloads.py` and `_extraction.py`.  That flag skipped extraction of the contingent paths in order to save space, but there are still some files in those paths that are necessary for building and the lite tarball can now be used instead.  

The `--sysroot` flag was removed from `downloads.py` and `_extraction.py` as well since it was only used to exclude the sysroot when `--skip-unused` is present.  The flag is still present and functions normally in the other scripts.  

If either of those flags are passed to `downloads.py` a warning will be logged noting that they are non-functional and will be fully removed in the future.  This should hopefully allow leeway for anyone that might still have the flags set in their builds.  

These changes will allow us to proceed with using the lite tarball (#3197) and should help with the possibility of moving the contingent paths into their own list file in the future.


Ahrotahn:
Right... without `--skip-unused` there isn't enough space on cirrus to complete the check.  @iskunk do you mind me adding  this change to your PR, or would you be ok with me adding yours to this one?

iskunk:
Better, I think, to just elaborate this PR. All it seems you're missing here (quickly eyeballing) is the change to `downloads.ini` and matching edit to `.cirrus.yml`. My main concern was to bring in / update the lite and testdata dirs to match the upstream scripts.

utils/prune_binaries.py
```diff
@@ -16,47 +16,101 @@
...
-    'third_party/llvm/',
-    'third_party/llvm-build/',
     'third_party/llvm-build-tools/',
+    'third_party/llvm-build/',
+    'third_party/llvm/',
     'third_party/node/linux/',
     'third_party/rust-src/',
     'third_party/rust-toolchain/',
     'third_party/webgl/',
 )
```
iskunk, line 102: I'd suggest keeping the lite/tarball sections of this list as near-copies of the upstream lists, just to make cross-checking and future merges easier to deal with. That's why I moved the `VK-GL-CTL` entry, etc.

iskunk (commented):
Just a couple remarks. I'm definitely liking the simplification of getting rid of `--skip-unused` and `--sysroot`.

Ahrotahn:
Sure thing, I've added the specific test data paths and re-ordered everything.  The only exception are the sysroots since the clone script is able to obtain ones that aren't available in the tarballs.

iskunk:
Thanks, this LGTM. I see the new entries in `pruning.list`, thankfully not too many. (They never fixed the the bug of a non-test target [`content_shell`] requiring test-data files in order to build)

PF4Public:
My only concern is whether lite tarball is usable for macOS build. 

@Cubik65536 have you tried building from lite tarball by any chance?

Cubik65536:
> My only concern is whether lite tarball is usable for macOS build. 
> 
> 
> 
> @Cubik65536 have you tried building from lite tarball by any chance?

I am super busy this week so the earliest time I can confirm this is next week, maybe weekend. I will try this as soon as I got time to do so.

Ahrotahn:
I'd give it another day or so to see if that causes problems with anyone's setup.  I've been thinking about any possible impact these changes might have.  

I've just made a small change that adds the sysroots.json as an exclusion to domain substitution.  This is because the lite tarball doesn't contain any sysroots.  So now anyone that relied on that would need to run install-sysroot.py to obtain them and the domain substitution would previously cause that to fail.  

I've also added a few more contingent paths to handle possible differences between clones and tarballs.

clickot:
portabl linux build works for me

networkException:
On NixOS as well, with minor modifications

iskunk:
@Cubik65536:
> I am super busy this week so the earliest time I can confirm this is next week, maybe weekend. I will try this as soon as I got time to do so.

Much obliged, please let us know what you find. Note that the lite tarball has [these directories](https://chromium.googlesource.com/chromium/tools/build/+/78948dcad2f195feed2fef0457660f92866e5c1b/recipes/recipes/publish_tarball.py#294) removed, so what I need to know is which (if any) of those needs to be restored for a successful macOS build.

If you find that files are missing, you can just copy in the appropriate directory from the regular/big tarball, and try again. (In the lite tarball, those directories aren't fully gone, they just have most of the file content removed. So you'll need to replace the appropriate directory in the lite tarball with one from the full tarball.)

teeminus:
Windows build works as well.

Cubik65536:
Works well directly on macOS (arm64)!

iskunk:
> Sorry for the delay, it's now tested! Works well directly on macOS (arm64)!

Thanks for confirming! So the lite tarball [in its current state] supports building for Linux, Windows, *and* macOS.

All that remains to look into, as far as I'm aware, is Android.

utils/prune_binaries.py
PF4Public: @iskunk was your concern addressed?

clickot:
May I ask what is preventing this PR from being merged? I would like to use the lite source tar for building 133.0.6943.141 :-)

PF4Public:
I see requests from @iskunk not marked as resolved

utils/prune_binaries.py
iskunk: > @iskunk was your concern addressed?

Yes, @Ahrotahn took the maintainability aspect into account. (I'm planning on at least one more change to the upstream lists, after all.)

I would have resolved this conversation, but I don't have the button for it...

Ahrotahn: Sorry, it seems that was available for me for some reason...

rany2:
> I see requests from @iskunk not marked as resolved

Everything seems to be marked as resolved? I can't find anything unresolved

PF4Public:
> Everything seems to be marked as resolved? I can't find anything unresolved

It is now, but it does not matter anymore, since these changes are incorporated into https://github.com/ungoogled-software/ungoogled-chromium/pull/3221

Ahrotahn:
That works as well, I'll close this one then.

Issue #3012, opened 2024-09-01
YouTube "Something went wrong. Refresh or try again later."


Fox8284:
### OS/Platform

Windows

### Installed

OS/Platform's package manager

### Version

128.0.6613.84

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [X] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [ ] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [ ] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

Can't play youtube videos

### How to Reproduce?

Clicking on a youtube video

### Actual behaviour

Whenever I click on a youtube video it will just be a black loading screen and it will say "Something went wrong. Refresh or try again later" after a while. However this only happens on a single one of my accounts, it will work fine signed out or on another account. But I have tried on other browsers and it works with the same account. I have tried disabling extensions and clearing cookies and updating chromium.

### Expected behaviour

Play the video

PF4Public:
Please provide additional information to support your issue. Developer tools console might have something of interest. Also Media tab.

likeyeeah:
Its the new Nvidia driver. Either rollback to the older version of it or sign out from the google and watch it without singing it. I dont know what did they changed but im using the older version of the nvidia driver.

Haus1:
Is it loading the first 30 seconds and then refusing to buffer any further? If you hold down reload it'll let you clear the cache, that and clearing cookies may help.

Fox8284:
> Please provide additional information to support your issue. Developer tools console might have something of interest. Also Media tab.

This is the log from the developer tools console
[www.youtube.com-1725398704722.log](https://github.com/user-attachments/files/16856605/www.youtube.com-1725398704722.log)

These are the logs from the media tab
[36206F9CF53C36A7637332F9C495554A.json](https://github.com/user-attachments/files/16856617/36206F9CF53C36A7637332F9C495554A.json)
[1DB67132BFE976838958335963265C5F.json](https://github.com/user-attachments/files/16856618/1DB67132BFE976838958335963265C5F.json)
[07A32B9A094E793A783FDD417636C807.json](https://github.com/user-attachments/files/16856619/07A32B9A094E793A783FDD417636C807.json)
[...]

PR #3209, opened 2025-02-23
Allow FedCm disabled response

Ahrotahn:
This PR changes how we handle part of disabling FedCm, rather than rejecting the request this returns that it is disabled.  

I believe this should solve some of the youtube issues mentioned in #3012.  It seems like youtube had been experimenting with integrating FedCm and finally made the changes generally available a few days ago.  

There are reports for other browsers having the same issue, so it's possible they also don't handle that the way the site expects.  All I can suggest for users of other browsers is to try launching with `--disabled-features=FedCm` on chromium-based browsers, and on firefox use an older user-agent since it doesn't support the API.


PF4Public (approved):

PF4Public:
I would suggest keeping the related issue open for longer time after merging this PR and leaving a comment that it should be fixed with the next release.

Ahrotahn:
Sorry for the bump, there was a counter that got forgotten.  

The issue isn't linked so it *should* remain open afterwards.

rany2:
This patch doesn't actually fix the issue for me and YouTube remains broken. https://github.com/ungoogled-software/ungoogled-chromium/commit/2dee7ebe4d8014f015ae1d9b40cea1a72ea5015c was tested specifically. I'm not sure if this is expected and I need to do something else?

rany2:
I started a new build, will let you know if it works out: https://github.com/flathub/io.github.ungoogled_software.ungoogled_chromium/pull/84

rany2:
Same deal @Ahrotahn :(

Ahrotahn:
Is there any difference when running that build with `--disabled-features=FedCm`?

rany2:
@Ahrotahn unfortunately it doesn't help either,

rany2:
Well, it actually works but I needed to wipe my profile.

Ahrotahn:
If you happen to have a backup copy of your prior profile I'd be curious to know if just clearing the site data for youtube works.  Maybe they are acting on some cookie/storage setting from before.  If a brand new profile is needed then maybe there's a profile pref involved.

rany2:
Unfortunately I don't have a backup of my old profile.

InfernalWraith:
Just like the others, I was completely blocked from YT aside from the first 20-ish seconds. After applying the flag you mentioned here along with disabling graphics acceleration (found at `chrome://settings/system`) worked for me without having to clear any site data. Kudos to https://github.com/ungoogled-software/ungoogled-chromium/issues/3012#issuecomment-2680582997 for figuring it out.

Ahrotahn:
Ok, I'll close this for now.  I'll have to do some experimentation later in the week to dig into those changes and see why they'd have an effect.

PR #3211, opened 2025-02-23
tracked_persistent_pref: don't use empty validation delegate

dumbmoron:
This is a specific snippet I have taken from chromium/chromium@d9a89d6e5cfa15c2a58091f972ccdd6fde57ebb3, which fixes a crash that happens on launch in a component/non`official` build when safe browsing is disabled. I included only this snippet from the patch since it's pretty large, but if you think I should include anything else, do let me know.

(The motivation behind this is that, while the upstream patch has already landed in >= 135, the stable release is still a month+ away. Once it's out, _this_ patch can just get removed.)

### DIFF
--- a/patches/series
+++ b/patches/series
@@ -1,6 +1,7 @@
 upstream-fixes/glue_core_pools.patch
 upstream-fixes/hardware_destructive_interference_size.patch
 upstream-fixes/missing-dependencies.patch
+upstream-fixes/fix-building-without-safe-browsing.patch
 
 core/inox-patchset/0001-fix-building-without-safebrowsing.patch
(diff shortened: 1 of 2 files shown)


networkException:
Interesting, is that an actual upstream effort that might render our `patches/core/{inox-patchset/0001-fix-building-without-safebrowsing.patch,ungoogled-chromium/fix-building-without-safebrowsing.patch}` obsolete?

Ahrotahn (approved):
This seems fine to me.  

The first of their changes landed in 134 and it'll take a bite out of the inox patch.  I haven't finished going through all of them yet but it seems like the most significant changes in the patch are being handled upstream now.

dumbmoron:
combined into #3212

PR #3217, opened 2025-02-26
patches/webgl: return generic info instead of empty strings

### LINKED ISSUE #3012: YouTube "Something went wrong. Refresh or try again later."

dumbmoron:
Instead of returning empty strings for WebGL info, return a generic renderer & vendor taken from Windows 10 [[1]](https://chromium.googlesource.com/angle/angle/+/main/doc/ExtensionSupport.md) [[2]](https://android.googlesource.com/platform/external/angle/+/b526112daf/scripts/extension_data/swiftshader_win10.json).

Seemingly fixes #3012

### DIFF
--- a/patches/extra/ungoogled-chromium/spoof-webgl-renderer-info.patch
+++ b/patches/extra/ungoogled-chromium/spoof-webgl-renderer-info.patch
@@ -1,25 +1,24 @@
-# Return blank values for WebGLDebugRendererInfo to remove a potential data
+# Return generic values for WebGLDebugRendererInfo to remove a potential data
 # leak while preventing potential website breakage
 
 --- a/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
 +++ b/third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc
-@@ -3833,8 +3833,7 @@ ScriptValue WebGLRenderingContextBase::g
-               pname, IdentifiabilityBenignStringToken(
+@@ -3834,7 +3834,7 @@ ScriptValue WebGLRenderingContextBase::g
                           String(ContextGL()->GetString(GL_RENDERER))));
          }
--        return WebGLAny(script_state,
+         return WebGLAny(script_state,
 -                        String(ContextGL()->GetString(GL_RENDERER)));
-+        return WebGLAny(script_state, String(""));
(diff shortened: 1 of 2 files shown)


Ahrotahn (approved):
Thanks @dumbmoron

PF4Public:
Are those strings generic enough? I mean, could it become a good fingerprinting point in the future?

dumbmoron:
> Are those strings generic enough? I mean, could it become a good fingerprinting point in the future?

I was curious, so I tried to perform some analysis on vendor/renderer strings seen in the wild. I found this dataset: https://github.com/apify/fingerprint-suite/blob/master/packages/fingerprint-generator/src/data_files/fingerprint-network-definition.zip
I have no idea how reliable or unreliable the dataset is, but it looks like `ANGLE (Google, Vulkan 1.3.0 (SwiftShader Device (Subzero) (0x0000C0DE)), SwiftShader driver)` (without the -5.0.0) might be more commonly seen in the wild. If you think it would be better, I could replace the renderer string with this one.

For reference, here are the results and code I used to analyze this dataset: https://gist.github.com/dumbmoron/bfa33483d9ea1b1c5ac08b5a281af46e

Ahrotahn:
While being less identifiable individually we were likely more identifiable as a group being the only browser that returned a blank string.  This would at least help to blend in with legitimate values.  It's also a good choice since it's platform-agnostic, that document mentioned windows 10 as an example but there's entries for other platforms such as [linux for example](https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/scripts/extension_data/swiftshader_linux.json).

PF4Public (approved):

networkException (approved):
Should we just merge this and do platform specific strings later?

Ahrotahn:
I don't see any reason we'd want to make them more specific.  It's better to keep them generic in order to reduce fingerprintability.

networkException:
ah oops, looks like I misread your previous comment

PR #3231, opened 2025-03-05
Updated WebGL info string

### LINKED ISSUE #3226: Infinite Cloudflare verification loop
### OS/Platform

NixOS

### Installed

OS/Platform's package manager

### Version

133.0.6943.141 (Official Build, ungoogled-chromium) (64-bit)

### Description

"Verify you are human" page keeps looping

### How to Reproduce?

1) Clean install browser
2) Allow third-party cookies everywhere
3) Attempt to open ryuugames.com (nsfw)

### Actual behaviour

Verification circle continues to rotate, then new "Verify you are human" element appears

### Expected behaviour

Pass Cloudflare checks, like in vanilla Chromium 133.0.6943.141 (Official Build) (64-bit)
The console log seems to be the same in all browsers, yet ungoogled variant stuck on the verification page for whatever reason.

### Relevant log output

```sh

```

### Additional context

_No response_

MVDW-Java:
This issue fixes #3226 by changing the WebGL string, I've tested this and it seems to work
Screenshot From 2025-03-05 07-39-23

### DIFF
--- a/patches/extra/ungoogled-chromium/spoof-webgl-renderer-info.patch
+++ b/patches/extra/ungoogled-chromium/spoof-webgl-renderer-info.patch
@@ -8,7 +8,7 @@
          }
          return WebGLAny(script_state,
 -                        String(ContextGL()->GetString(GL_RENDERER)));
-+                        String("ANGLE (Google, Vulkan 1.3.0 (SwiftShader Device (Subzero) (0x0000C0DE)), SwiftShader driver-5.0.0)"));
++                        String("ANGLE (Intel, Intel(R) UHD Graphics 620 (0x00005917) Direct3D11 vs_5_0 ps_5_0, D3D11)"));
        }
        SynthesizeGLError(
            GL_INVALID_ENUM, "getParameter",


PF4Public:
I believe @dumbmoron is already researching that.

Ahrotahn:
This should be fine so long as this string works for those that were having issues.  All the ones I tried worked for me, so I can't really make a determination on the best choice.  

@MVDW-Java When you get the chance please rebase this over the latest changes

dumbmoron:
You probably also want to update the vendor string, to `Google Inc. (Intel)`

MVDW-Java:
Just finished compiling and testing, everything seems to work fine

PF4Public:
> Just finished compiling and testing, everything seems to work fine

How do we test those cases where "Google" was problematic and make sure "Intel" is better?

PF4Public:
@MVDW-Java Would you be able to make the string configurable?

MVDW-Java:
> > Just finished compiling and testing, everything seems to work fine
> 
> How do we test those cases where "Google" was problematic and make sure "Intel" is better?

What I did is to go to spigotmc.org or try to register on the GNOME gitlab.
It failed when I tried it on the current flathub build using the `--user-data-dir` argument.

> @MVDW-Java Would you be able to make the string configurable?

I have the patch done, but it takes a while to compile for me to test.
I'll let it run overnight and push the changes when it's fully ready tomorrow morning.

PF4Public:
> What I did is to go to spigotmc.org or try to register on the GNOME gitlab.

Which implies the string was tested on your hardware and fixes your case. No guaranties it wouldn't fail elsewhere.

PF4Public:
I would imagine making the default behaviour to return real strings and spoofed one to return something that user had chosen. One of the options would be to put "Intel" into flag description similar to [add-flag-to-change-http-accept-header.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/patches/extra/ungoogled-chromium/add-flag-to-change-http-accept-header.patch).

MVDW-Java:
> @MVDW-Java Sorry, I forgot to add, maybe prefix the switches with `spoof-` for correctness and to future-proof it in case it ever makes sense to actually choose a renderer.
> 
> As for the strings returned by `SpoofWebGLInfo`, if the default Google ones prove to be too problematic for general use, we could try using an empty space `" "` (according to #3226, it works). Spoofing specific devices should best be done with the new granular controls.

Yeah, I was thinking about making it `" "` as well, but I wonder about the privacy implications. Having only a space can be somewhat identifiable, as I think no other browser does this.

> I would imagine making the default behaviour to return real strings and spoofed one to return something that user had chosen. One of the options would be to put "Intel" into flag description similar to [add-flag-to-change-http-accept-header.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/patches/extra/ungoogled-chromium/add-flag-to-change-http-accept-header.patch).

Sounds good, but I sadly don't have much time this week (and me being new with the chromium codebase took me a while to figure things out 😅)
If you want, you can contribute to this pr

Ahrotahn:
Unfortunately a switch won't work for this part of the code so this will have to be implemented with a feature flag.  It's possible to have one that can be dynamically set but I recall that being pretty awkward to use.  We should at least be able to have a few different options to choose from.  I'll try to set up some changes to work with that soon.

Ahrotahn:
I've pushed an update to your PR to make the flag have multiple options in chrome://flags as well as allowing the launch flag to be configurable.  

I dug through the data @dumbmoron had posted in the issue and made some of the most popular generic strings the choices.  Being generic is important since a user would be more identifiable than using the real string if they chose one with Direct3D on a non-windows platform for example.  That can still happen if, for another example, a user chooses the Apple string on a non-Apple device, but at least it should lessen the chance of inadvertently choosing a bad option that would make sense otherwise.  

Interestingly the blank string shows up in that data and it shows up for seven other user agents, so there seems to be other browsers using that.  Some info about how the strings are handled can bee seen in [GrGLUtil.cpp](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/gpu/ganesh/gl/GrGLUtil.cpp) which helps explain why that would work.  A site can compare the GL extensions available to your vendor & renderer and if they don't match could assume you're not a real agent.  The blank string means that comparison can't be made.  So I've made that the default if the flag is enabled without any options.

To set a custom string launch with:
`--enable-features="SpoofWebGLInfo:renderer/CUSTOMRENDERER/vendor/CUSTOMVENDOR"`

You can paste this in the console to see what is being used:
```javascript
let e=document.createElement("canvas").getContext("webgl"),n=e.getExtension("WEBGL_debug_renderer_info");console.log("renderer: "+e.getParameter(n.UNMASKED_RENDERER_WEBGL)+"\nvendor: "+e.getParameter(n.UNMASKED_VENDOR_WEBGL));
```

MVDW-Java:
@Ahrotahn Thanks for helping out :smile: 
I just tested it, and everything works fine on my machine.
Would it be possible to merge this for the next release?

Ahrotahn:
This'll need some more confirmations and approvals before it can be merged.  
Since I didn't open the PR it seems I can add my own 😆

Ahrotahn (approved):

PF4Public (approved):

oep42:
Thanks for implementing the spoof-webgl-info or SpoofWebGLInfo flag. For those who do not understand the details of this flag, it suffers from a lack of documentation.

Could someone (perhaps @Ahrotahn) be so kind to answer the following questions:

1. What does the "Default" value of this flag do? Is it the same as any of the other values?
2. What does the "Enabled" value of this flag do? 
3. Which circumstances (i.e. which combination of operating system, CPU brand, GPU brand, and/or other system characteristics) warrant which values of this flag? In other words, which of the values (other than Disabled) can sensibly be used under which circumstances?

Ahrotahn:
Features always have an Enabled and Disabled option even when it doesn't make sense.  I recall reading a comment from one of the upstream devs lamenting that.  When I get the time I'll update the docs to include this info as well as add the certificate transparency flag.  

1.  Since the feature is disabled by default both the `Default` option and `Disabled` option do  the same thing:  Return the real strings like every other chromium browser.  
2.  The `Enabled` option is the same as the `Blank` option which returns a blank string.
3.  This depends on the hardware you have.  These strings were intended for debugging purposes but now it is used as a fingerprinting vector.  So you would want to choose the option that aligns with your hardware or blank.

PR #3235, opened 2025-03-08
pruning: update openscreen contingent path

dumbmoron:
Currently, when trying to build from a fresh clone (not a tarball), I am seeing the following error when pruning binaries:
```
ERROR: 1 files could not be pruned:
third_party/openscreen/src/buildtools/linux64-format/clang-format
```

This is because the `openscreen` third party module changed the path for this binary from `linux64/format/clang-format` to `linux64-format/clang-format`, which means it ended up on the pruning list even though it's not present in the Chromium git repo. This PR updates the CONTINGENT_PATHS to consider this new updated path.


Ahrotahn (approved):
Thanks @dumbmoron!

PF4Public (approved):

PR #3243, opened 2025-03-14
Enable Certificate Transparency Enforcement

### LINKED ISSUE #3242: Enable certificate transparency enforcement
### Description

Enable certificate transparency enforcement

### Who's implementing?

- [x] I'm willing to implement this feature myself

### The problem

Certificate transparency isn't enforced. You can check for this by using a test site like https://no-sct.badssl.com/. On chrome this results in an error, but on ungoogled-chromium it loads fine. 

Enabling this enhances security because it makes it more difficult for misissued certificates to go undetected. [Most browsers enforce certificate transparency](https://developer.mozilla.org/en-US/docs/Web/Security/Certificate_Transparency#browser_requirements), so ungoogled-chromium benefits from some "herd immunity", but it's still better to enforce it ourselves than to rely on that.

### Possible solutions

The comment in chromium source code suggests this requires a bunch of work:
```
// Enables Certificate Transparency on Desktop and Android Browser (CT is
// disabled in Android Webview, see aw_browser_context.cc).
// Enabling CT enforcement requires maintaining a log policy, and the ability to
// update the list of accepted logs. Embedders who are planning to enable this
// should first reach out to chrome-certificate-transparency@google.com.
```
But I toggled that FEATURE_ENABLED_BY_DEFAULT, and it seems to work as expected with no further changes.

### Alternatives

_No response_

### Additional context

_No response_

dot-gov:
Enabled by default because it improves security and is the default on Chrome. There's a flag added anyways in case it goes haywire and needs to be disabled.

closes https://github.com/ungoogled-software/ungoogled-chromium/issues/3242

### DIFF
--- a/patches/extra/ungoogled-chromium/enable-certificate-transparency-and-add-flag.patch
+++ b/patches/extra/ungoogled-chromium/enable-certificate-transparency-and-add-flag.patch
@@ -1,0 +1,22 @@
+--- a/chrome/browser/browser_features.cc
++++ b/chrome/browser/browser_features.cc
+@@ -50,7 +50,7 @@ BASE_FEATURE(kBookmarkTriggerForPrerender2,
+ // switch.
+ BASE_FEATURE(kCertificateTransparencyAskBeforeEnabling,
+              "CertificateTransparencyAskBeforeEnabling",
+-#if BUILDFLAG(GOOGLE_CHROME_BRANDING)
++#if true
+              base::FEATURE_ENABLED_BY_DEFAULT);
+ #else
+              base::FEATURE_DISABLED_BY_DEFAULT);
+--- a/chrome/browser/ungoogled_flag_entries.h
++++ b/chrome/browser/ungoogled_flag_entries.h
+@@ -140,4 +140,8 @@
+      "Spoof WebGL Info",
+      "Return generic values for WebGLDebugRendererInfo to remove a potential data leak while preventing potential website breakage. ungoogled-chromium flag.",
+      kOsAll, FEATURE_WITH_PARAMS_VALUE_TYPE(blink::features::kSpoofWebGLInfo, kSpoofWebGLChoices, "SpoofWebGLInfo")},
++    {"enforce-certificate-transparency",
++     "Enforce Certificate Transparency",
++     "Enforce Certificate Transparency for certificates that sites present. This is enabled by default. ungoogled-chromium flag.",
++     kOsAll, FEATURE_VALUE_TYPE(features::kCertificateTransparencyAskBeforeEnabling)},
+ #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
--- a/patches/extra/ungoogled-chromium/enable-certificate-transparency-and-add-flag.patch
+++ b/patches/extra/ungoogled-chromium/enable-certificate-transparency-and-add-flag.patch
@@ -14,7 +14,7 @@
 @@ -140,4 +140,8 @@
       "Spoof WebGL Info",
       "Return generic values for WebGLDebugRendererInfo to remove a potential data leak while preventing potential website breakage. ungoogled-chromium flag.",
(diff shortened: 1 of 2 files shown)


PF4Public:
> closes #3242

Haven't you forgot to address the concerns voiced there?

dot-gov:
If you're talking about

>// Enabling CT enforcement requires maintaining a log policy, and the ability to
>// update the list of accepted logs. Embedders who are planning to enable this
>// should first reach out to chrome-certificate-transparency@google.com.

I figured that's not needed because:

1. It builds fine and works with only the patches in this PR
2. Firefox doesn't maintain an accepted log list, and instead [piggybacks off Chrome's](https://wiki.mozilla.org/SecurityEngineering/Certificate_Transparency#Enterprise_Policies). That same list is [already included in chromium](https://source.chromium.org/chromium/chromium/src/+/main:components/certificate_transparency/data/log_list.json). If it's good enough for Firefox it must be good enough for us as well.

My guess is that the warning for embedders is there because they might be less frequently updated, which might cause issues if the logs that are compiled into a specific build expire (all the logs in the current list expire by 2027), and breaks https connections. For ungoogled-chromium, which is updated just as regularly as chrome this shouldn't be an issue.

Ahrotahn:
This seems fine to me, but I worry that would cause us to receive new issues for people that add custom CAs and aren't aware of the change.  I'll run with `--enable-features=CertificateTransparencyAskBeforeEnabling` for a while to see if there are any unforeseen issues with having it enabled.  

If this doesn't work out then I think it could still be a good addition to `add-flags-for-existing-switches.patch` so that can be enabled from chrome://flags.

One small nit:  from a maintainability standpoint it would be better to remove the guard instead of shorting it.

dot-gov:
> This seems fine to me, but I worry that would cause us to receive new issues for people that add custom CAs and aren't aware of the change. I'll run with `--enable-features=CertificateTransparencyAskBeforeEnabling` for a while to see if there are any unforeseen issues with having it enabled.

It should be fine because locally installed CAs are excluded by default. 
https://chromium.googlesource.com/chromium/src/+/lkgr/net/docs/certificate-transparency.md#locally_trusted-cas

> One small nit: from a maintainability standpoint it would be better to remove the guard instead of shorting it.

Removing #if entirely would definitely be cleaner, but is arguably less maintainable, because it would require more lines to be modified, and thus present a higher chance that the patch would produce conflicts that need to be manually resolved.

Ahrotahn:
Good, I don't see any issues with that then.  

> Removing #if entirely would definitely be cleaner, but is arguably less maintainable, because it would require more lines to be modified, and thus present a higher chance that the patch would produce conflicts that need to be manually resolved.

It might seem backwards at first, but that's exactly what makes it more maintainable!  Updating the patches is the easy part, debugging a build that doesn't work correctly is much more difficult and time-consuming.  

It is less likely to be an issue here where only a single line is being gated, but in the half-decade I've been dealing with these patches the sections that short defines have caused some of the largest time-sinks when updating.  I've considered making a write-up about this topic including alternative methods I've experimented with.  I might need to prioritize doing that when I get some free time.

COD3HUNT3R:
CT is the same type of snake oil as EV was.

dumbmoron:
> CT is the same type of snake oil as EV was.

can you elaborate on that?

COD3HUNT3R:
> > CT is the same type of snake oil as EV was.
> 
> can you elaborate on that?

I am sure nobody is going to read a wall of text. In short it looks great on paper but in reality this is same chicken egg issue like all other security protocols to make **lives easier**. A handful of Goliaths running CAs, CT Logs, log policies through embedders aka browsers. This doesn't help normies of the internet or the said domain owners. It is mostly used by ethical hackers or bad actors to monitor **assets** of any organizations.

networkException:
I don't feel like this belongs here, can we stay on topic please? Not implementing this in here will not make CAs log less domains

networkException (approved):

Ahrotahn (approved):

PR #3255, opened 2025-03-23
Windows-specific improvements (ARM64 and Python)

tangalbert919:
This pull request does the following:
* Enables downloading PGO profiles for Win-ARM64 builds.
* Forces `gclient.bat` to call system Python directly instead of Google's vpython scripts.

The second one is primarily due to [missing httplib2](https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/349), which I have been hitting repeatedly, despite having it installed. This could be fixed simply by explicitly calling the `vpython.bat` file, but it would be better to skip that and call `python3` directly.

### DIFF
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -268,7 +268,7 @@ def main():
     parser.add_argument('-p',
                         '--pgo',
                         default='linux',
-                        choices=('linux', 'mac', 'mac-arm', 'win32', 'win64'),
+                        choices=('linux', 'mac', 'mac-arm', 'win32', 'win64', 'win-arm64'),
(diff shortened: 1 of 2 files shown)


PF4Public:
Would it have any implications on other win-builds?

Ahrotahn (approved):
Seems fine to me. It would be good to get input from another on the windows team that the change has no unintended side-effects.

teeminus (approved):
Compiling still works fine with this changes. (I tried the default build target: x64 using the clone script)

PF4Public (approved):

Issue #3270, opened 2025-04-10
Cookies are deleted after closing the browser


jroovy:
### OS/Platform

Arch Linux

### Installed

https://ungoogled-software.github.io/ungoogled-chromium-binaries/

### Version

135.0.7049.52

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [ ] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [ ] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [ ] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

Closing the browser deletes saved cookies

### How to Reproduce?

1. Login to any website
2. Close & relaunch the browser
3. Open the website again

### Actual behaviour

User is logged out of website, and `F12 > Application > Cookies` shows that the saved cookies are gone.

### Expected behaviour

Cookies should still be in the browser after relaunching.

### Relevant log output

```sh

```

### Additional context

Tested on AppImage from https://github.com/ungoogled-software/ungoogled-chromium-portablelinux/releases/download/135.0.7049.52-1/ungoogled-chromium_135.0.7049.52-1.AppImage

`--user-data-dir` also has the same problem

Links2004:
running in to the same problem.
Version: 135.0.7049.52

stalker0912:
Version 135.0.7049.41 on Windows has the same issue.

LightTreasure:
I have the same problem with v135.0.7049.52 on Nobara Linux. 

Some additional info: The `Cookies` file in my user profile dir remained intact after I closed the browser, but got overwritten when I opened it.

erreing:
![Image](https://github.com/user-attachments/assets/3a6fcb14-ace8-4d74-a1b3-2147c685430a)

jroovy:
> ![Image](https://github.com/user-attachments/assets/3a6fcb14-ace8-4d74-a1b3-2147c685430a)

Where is this setting located in?

cmcaine:
I think the setting is no longer present. You may be able to change the setting here instead: chrome://flags/#clear-data-on-exit

![Image](https://github.com/user-attachments/assets/1b93fec1-3868-482f-a6d0-7f215f4e5814)

jroovy:
@erreing @cmcaine disabling this setting didn't fix the issue.

erreing:
[...]

PR #3271, opened 2025-04-12
docs: update location of clear-data-on-exit flag

cmcaine:
See #3270

### DIFF
--- a/docs/default_settings.md
+++ b/docs/default_settings.md
@@ -9,7 +9,7 @@ Allow sites to check if you have payment methods saved | Disabled | chrome://set
 Ask where to save each file before downloading | Enabled | chrome://settings/downloads
 Auto sign-in | Disabled | chrome://settings/passwords
 Block third-party cookies | Enabled | chrome://settings/cookies
-Clear cookies and site data when you close all windows | Enabled | chrome://settings/cookies
+Clear cookies and site data when you close all windows | Enabled | chrome://flags/#clear-data-on-exit
 Continue running background apps when Chromium is closed | Disabled | chrome://settings/system
 Hyperlink auditing (\<a ping>) | Disabled | NA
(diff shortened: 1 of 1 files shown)


Ahrotahn:
The clear-data-on-exit flag is [disabled by default](https://github.com/ungoogled-software/ungoogled-chromium/blob/135.0.7049.84-1/patches/extra/ungoogled-chromium/add-flag-to-clear-data-on-exit.patch#L7).  

That line in docs/default_settings.md exists because, prior to version 121, we had a patch that enabled the toggle [erreing posted](https://github.com/ungoogled-software/ungoogled-chromium/issues/3270#issuecomment-2798116534) in the comments of that issue:  
https://github.com/ungoogled-software/ungoogled-chromium/blob/120.0.6099.234-1/patches/extra/iridium-browser/prefs-only-keep-cookies-until-exit.patch

It should have been removed back when that patch was removed, so it would be great if you could update the PR to do that.  

I'm able to reproduce the cookie issue in vanilla chromium, so something has changed when using new profiles.  I'll see if I can find further information.

docs/default_settings.md
```diff
@@ -9,7 +9,7 @@ Allow sites to check if you have payment methods saved | Disabled | chrome://set
 Ask where to save each file before downloading | Enabled | chrome://settings/downloads
 Auto sign-in | Disabled | chrome://settings/passwords
 Block third-party cookies | Enabled | chrome://settings/cookies
-Clear cookies and site data when you close all windows | Enabled | chrome://settings/cookies
+Clear cookies and site data when you close all windows | Enabled | chrome://flags/#clear-data-on-exit
```
cmcaine, line 12:
```suggestion
```

Ahrotahn (approved):
Thanks!

PF4Public (approved):

PR #3297, opened 2025-05-01
patches/referrer-customization: update component name for export

dumbmoron:
Fixes a failing component build due to the component export name being changed to `NETWORK_CPP_FLAGS_AND_SWITCHES` upstream: 
https://source.chromium.org/chromium/chromium/src/+/main:services/network/public/cpp/features.h


PF4Public (approved):

Ahrotahn (approved):
Thanks @dumbmoron

PR #3353, opened 2025-06-16
Remove the F1 shortcut + Add an optional tweak for Incognito/Guest profiles

implicitfield:
The first change here removes the F1 shortcut, which just tries to open a Google help page (unless you're on macOS, where it does nothing since this block of code is `#ifdef`ed out there).

The second change adds a new flag that allows increasing the storage quota for Incognito/Guest (aka OTR / off-the-record) profiles. The point of this is that makes it harder for sites to detect that you're browsing in Incognito mode (because it feels pretty invasive when they do that). This works because AFAIK these days the only remaining clear indicator that you're using Incognito mode is the extremely small quota size there, which can be obtained through various browser APIs as well as probably many side-channels.
Concretely, this patch allows you to get a higher quota by switching to the same algorithm that's used for "regular" profiles. Naturally, you'll need a fair amount of ram to overcome whatever thresholds scripts use as the cutoff point, but at least 8GiB should be enough to trick the most popular library.

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -69,6 +69,7 @@ These are also available on the `chrome://flags` page.
   `RemoveClientHints` | Removes client hints (information sent to servers about your system, similar to a user agent).
   `SetIpv6ProbeFalse` | Forces the result of the browser's IPv6 probing (i.e. IPv6 connectivity test) to be unsuccessful. This causes IPv4 addresses to be prioritized over IPv6 addresses. Without this flag, the probing result is set to be successful, which causes IPv6 to be used over IPv4 when possible.
   `SpoofWebGLInfo` | Return generic values for WebGLDebugRendererInfo to remove a potential data leak while preventing potential website breakage.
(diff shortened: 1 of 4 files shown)


Ahrotahn:
This looks good to me.  The flag worked well on https://detectincognito.com on my setup.  

Would you mind adding that to the feature flags section of docs/flags.md as well?

implicitfield:
> Would you mind adding that to the feature flags section of docs/flags.md as well?

No problem, I've added this flag to there as well.

Ahrotahn (approved):
Thanks @implicitfield!

PF4Public (approved):

PF4Public:
As a side note, I wonder if we have a definition of what goes to core and what is extra…

Ahrotahn:
Yeah in retrospect the F1 patch would make more sense in extra, but that's something that can be moved later.  

@implicitfield Due to the changes in 138 the first patch will need to have the offsets updated from 129 to 132 before we can merge this

PR #3369, opened 2025-06-30
Update to Chromium 138.0.7204.96

networkException:
This PR updates to Chromium 138.0.7204.96

This includes a fix for CVE-2025-6554 for which exploits exist in the wild.

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/138.0.7204.49..138.0.7204.96?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2025/06/stable-channel-update-for-desktop_30.html)


networkException:
There's a mismatch between API and blog regarding the specific version :shrug:

Ahrotahn:
deopt_tools updated to use syntax only available in python 3.10+, so CI isn't happy atm.  I'll rerun it once the tarball is available in about an hour and it should pass. I'll start working on updating our config to use a newer version over the next day or so as I get time.

Ahrotahn (approved):
Builds and runs fine here, thanks @networkException!

teeminus:
> deopt_tools updated to use syntax only available in python 3.10+, so CI isn't happy atm. I'll rerun it once the tarball is available in about an hour and it should pass. I'll start working on updating our config to use a newer version over the next day or so as I get time.

We might have a solution for the python version requiredment change: https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/468

clickot:
just for sure: what chromium version should i use to build for portable linux, .92 or .96? Does it matter anyways?

Ahrotahn:
I'm not sure why their API and blog are at odds, but after looking at the changes it would probably be best to have everyone on .96.

iskunk:
Note that the lack of a Linux .92 release means that the Debian `chromium` package release of .92 can't be converted into a matching u-c package. The latest u-c packages that I have for Ubuntu remain at 138.0.7204.49.

I'd request tagging each (Linux) stable release even if the project itself goes with a slightly later version, so that other builds that track the stable channel can get their updates.

PR #3372, opened 2025-07-02
Update CI config, scripts, and docs for python 3.10

Ahrotahn:
This PR updates the Cirrus configs to use an image with python 3.10 so it is able to run depot_tools again.  This also updates our scripts for the image's new version of pylint.

Pylint removed `bad-continuation` so occurrences of that option have been removed.  It also required changes to spacing, using f-strings instead of format, using brackets and braces instead of empty dicts and lists, explicitly setting encoding, and using more specific exceptions.  There were a few lines that I've disabled `consider-using-with` since it would require more work to adjust and might be better for a future overhaul.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ container:
 
 code_check_task:
     pip_cache:
-        folder: /usr/local/lib/python3.9/site-packages
+        folder: /usr/local/lib/python3.10/site-packages
         fingerprint_script: cat .cirrus_requirements.txt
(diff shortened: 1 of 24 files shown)


PF4Public (approved):
Shouldn't we update to the current stable python-3.13?

Ahrotahn:
We generally used the earliest version possible for broader support, but there's nothing stopping us from requiring a later version.

PF4Public:
Well, if python-3.10 is fine, it is fine by me too :)

PF4Public:
@networkException

PR #3387, opened 2025-07-15
Add -snld flag to 7z

derron:
Fix "ERROR: Dangerous symbolic link path was ignored" when extracting chromium tarball sources.


Ahrotahn:
This seems fine to me, but it would be good if someone on the windows team can verify that doesn't cause any unintended side-effects.

teeminus:
Compiling from tarball still works on windows with the proposed change 👍

Ahrotahn (approved):
Ok, thanks y'all!

Issue #3385, opened 2025-07-10
Updating to 138.0.7204.101 Causes Chromium To Crash on Start-up


muHashh:
### OS/Platform

macOS

### Installed

OS/Platform's package manager

### Version

138.0.7204.101

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [x] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [x] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [x] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

This version keeps crashing startup after upgrading from v137. I can provide the detailed crash logs privately.

### How to Reproduce?

Self-explanatory

### Actual behaviour

.

### Expected behaviour

.

### Relevant log output

```sh

```

dumbmoron:
> ### Installed
> OS/Platform's package manager
> 
> ### Version
> 138.0.7204.101

i fail to see how this is possible when there is 
1. no 138.0.7204.101 ungoogled-chromium release, and
2. no 138.0.7204.100 macos build

fordfrog:
i'm also having crashes with the 138 releases, on gentoo. but i am able to run a clean profile (new user data dir) with those versions, currently 138.0.7204.100_p1. in my case, all the tabs open, proxy credentials window appears, and then all the windows crash, even before i have time to provide credentials for the proxy.

YesTrustMe:
I suppose one shouldn't be surprised. Someone mentioned this browser on reddit now that Chrome has killed uBlock. Nowhere here does it say which version we should download. 

Not sure why we just don't have one version for W64, another for Linux 64, and another for Mac. Developers can and should do better.

networkException:
what

networkException:
what

PF4Public:
> i'm also having crashes with the 138 releases, on gentoo

Does the issue exist in https://github.com/PF4Public/ungoogled-chromium-bin/releases/tag/138.0.7204.100 ? And in Portable https://github.com/ungoogled-software/ungoogled-chromium-binaries/pull/663#event-18573763101 ?

muHashh:
> > ### Installed
> > OS/Platform's package manager
> > ### Version
> > 138.0.7204.101
> 
> i fail to see how this is possible when there is
> 
> 1. no 138.0.7204.101 ungoogled-chromium release, and
[...]

PR #3393, opened 2025-07-18
patches/disable-gcm: disable additional gcm client methods

dumbmoron:
Patches out call to `MCSClient::Login` (which was causing a crash as reported in #3385), as well as all other unconditional calls to MCSClient as a precaution.

I was unable to reproduce the crash locally, but my guess as to why the crash happened is:
- in `GCMClientImpl::StartGCM()`,  the condition `device_checkin_info_.IsValid()` succeeded for some users (not sure why, maybe a copypasted profile from chrome?)
- and so `GCMClientImpl::OnReady()` was called [here](https://source.chromium.org/chromium/chromium/src/+/main:components/gcm_driver/gcm_client_impl.cc;l=421;drc=d0260a368e65b2be56d179d21fab90976fd494b8;bpv=1;bpt=1), which called `GCMClientImpl::StartMCSLogin()`, which called `mcs_client_->Login()`, which caused a segmentation fault because `mcs_client_` was not constructed, as it was patched out.

The fix was confirmed as working by https://github.com/ungoogled-software/ungoogled-chromium/issues/3385#issuecomment-3089734292, but I would appreciate others on different platforms/environments checking that it also resolves the issue for them.

### DIFF
--- a/patches/core/ungoogled-chromium/disable-gcm.patch
+++ b/patches/core/ungoogled-chromium/disable-gcm.patch
@@ -10,7 +10,37 @@
    std::vector<GURL> endpoints;
    endpoints.push_back(gservices_settings_.GetMCSMainEndpoint());
    GURL fallback_endpoint = gservices_settings_.GetMCSFallbackEndpoint();
-@@ -610,23 +611,6 @@ void GCMClientImpl::RemoveHeartbeatInter
+@@ -482,8 +483,6 @@ void GCMClientImpl::OnReady(const std::v
+ void GCMClientImpl::StartMCSLogin() {
+   DCHECK_EQ(READY, state_);
+   DCHECK(device_checkin_info_.IsValid());
+-  mcs_client_->Login(device_checkin_info_.android_id,
+-                     device_checkin_info_.secret);
+ }
+ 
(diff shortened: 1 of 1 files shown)


Ahrotahn:
I was able to reproduce the problem by creating a new profile with a vanilla chromium build and then using it with an ungoogled-chromium build.  I guess previously running another chromium browser with their profile may be the common denominator for those experiencing that issue.  

I tried out the changes and the crash is solved but now the browser throws repeating errors attempting to contact a messaging registration URL:
```
[76576:76593:0718/105725.389966:ERROR:src/chromium/net/url_request/trk_protocol_handler.cc:17] Blocked URL in TrkProtocolHandler: trk:111:https://android.clients.9oo91e.qjz9zk/c2dm/register3
[76544:76559:0718/105725.390078:ERROR:src/chromium/google_apis/gcm/engine/registration_request.cc:274] Registration URL fetching failed.
```
Gutting `GCMClientImpl::Start` seems to prevent that.  I'll run with these changes for a while to make sure there aren't any other side-effects.

PF4Public:
> Gutting `GCMClientImpl::Start` seems to prevent that.

Does it mean the patch needs some more changes?

PR #3424, opened 2025-08-08
Disable all Google AI starter pack `@shortcuts`

wukko:
This PR disables all AI-related starter pack shortcuts: `@gemini`, `@aimode`, and `@page` (the last one is currently disabled behind a flag).

The [existing patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/a078b9be0f51f5b866edc0fd21996d4c7816f07f/patches/extra/ungoogled-chromium/prepopulated-search-engines.patch) that removes `@gemini` via `kStarterPackExpansion` doesn't disable the newly added `@aimode`, despite the two being (very) closely related. This seems to be due to an oversight on Google's part, because `@aimode` gating [was fixed two days ago in main](https://chromium-review.googlesource.com/c/chromium/src/+/6729947), but it is still not controlled by `kStarterPackExpansion`.

Since it's unknown when the `@aimode` gating fix will land in a release build, I think it's best to disable it and the other shortcuts by other means (at least for now). This patch will have to be refactored in the future, but I generally think this is a better approach to removing these shortcuts for good. It also ensures that profiles created with `@aimode` enabled will not have it after the merge.

Before:
image

After (same profile):
image

### DIFF
--- a/patches/extra/ungoogled-chromium/disable-ai-search-shortcuts.patch
+++ b/patches/extra/ungoogled-chromium/disable-ai-search-shortcuts.patch
@@ -1,0 +1,42 @@
+--- a/chrome/browser/ui/search_engines/template_url_table_model.cc
++++ b/chrome/browser/ui/search_engines/template_url_table_model.cc
+@@ -119,12 +119,11 @@ void TemplateURLTableModel::Reload() {
+   for (TemplateURL* template_url : urls) {
+     // Don't include the expanded set of starter pack keywords if the expansion
+     // feature flag is not enabled.
+-    if ((template_url->starter_pack_id() ==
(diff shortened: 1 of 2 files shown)


Rusenche:
The proposal is interesting, I support this idea.

I also think that: `gemini`, `aimode` and in general any form of `AI`, `Microsoft recall` **should be banned forever**.

Ahrotahn (approved):
Works as expected on my build.  Next time I think about it I might roll the change in the prepopulated search patch into this.  It used to do a lot more but over time had been whittled down to just the starter pack change.  

Thanks @wukko

PF4Public (approved):

PR #3442, opened 2025-08-20
patches/block-requests: prevent profile_resetter from making requests

dumbmoron:
Currently, whenever any settings page is opened in Chromium, it attempts to make a request to `https://tools.google.com/service/update2`.

This patch patch removes the entire construction of the class responsible for this, so this does not happen anymore.


Ahrotahn:
This seems fine to me.  I guess you were building with branding?  
When you get the chance please rebase over the latest commit so we can ff merge.

dumbmoron:
> I guess you were building with branding?

For some reason, on macOS, Chromium loads the brand code from a hardcoded path: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/google/google_brand_mac.mm;l=64-66;drc=61ba235815cd78abb9b943012cf74efa56975f42, which I presume was dropped there by a previous Chrome install. ¯\\\_(ツ)\_/¯

PF4Public (approved):

Ahrotahn (approved):
I see, thanks @dumbmoron!

PR #3483, opened 2025-09-25
Update to Chromium 141.0.7390.37

Ahrotahn:
This PR updates to Chromium 141 which should be released to stable on the 30th

Notes:

* The webstore urls patch had a removal for [[6827355]](https://chromium-review.googlesource.com/c/chromium/src/+/6827355)

* The pruned binaries patch was updated for [[6862004]](https://chromium-review.googlesource.com/c/chromium/src/+/6862004), and [[6806229]](https://chromium-review.googlesource.com/c/chromium/src/+/6806229)

* The safebrowsing patch had changes for [[6841438]](https://chromium-review.googlesource.com/c/chromium/src/+/6841438)

* The remove unused prefs patch has changes for [[6832241]](https://chromium-review.googlesource.com/c/chromium/src/+/6832241)

* Clang modules have been enabled upstream, so if your tooling doesn't yet support this then you'll need to build with `use_clang_modules=false`.  There are some dependencies missing building this way but luckily the gentoo crew have already figured out what is needed and can be resolved by applying their changes in [chromium-141-cssstylesheet-iwyu.patch](https://gitweb.gentoo.org/repo/gentoo.git/tree/www-client/chromium/files/chromium-141-cssstylesheet-iwyu.patch)

* Features are being updated to two-parameter definitions: [(436274260)](https://issues.chromium.org/issues/436274260).  I've kept the features we've added as they are currently since there is still some churn going on with those changes, but we'll most likely have to update them all for 142.

* v0.30 of `httplib2` removed the socks class, so I've added a new change to the depot_tools patch to account for that.  You'll need to install `pysocks` to use the cloning script when using the newer versions of `httplib2`.

* If colors look light/washed-out on linux then try launching with `--disable-features=WaylandWpColorManagerV1`

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/140.0.7339.207..141.0.7390.37?pretty=fuller&n=10000)


PF4Public:
> * If colors look light/washed-out on linux then try launching with `--disable-features=WaylandWpColorManagerV1`

Does it apply to Wayland or is X11 also affected?

Ahrotahn:
It should only affect wayland sessions, and from what I understand only ones that support HDR

PF4Public:
Are we merging latest version bump for 140 first?

Ahrotahn:
Yup!

implicitfield:
Builds and runs fine on macOS (just finished building this locally).

PR #3517, opened 2025-10-24
devutils/platform_patches: update shared patches instead of deleting

dumbmoron:
Instead of deleting copies of generic patches when unmerging them from platform patches, move them back into the shared repository.

This makes it much more pleasant and ergonomic to work on shared patches in a merged series, and allows to refresh, or otherwise modify existing patches by "unmerging", instead of having to move them back to the original folder manually.

My case for why this potentially breaking change should be made: I have been using this as a personal preference in helium for a few months: imputnet/helium@d56c8d5677107f01ac90049f537a43325a57ae8b, and it has made working with patch series much simpler - so much that I believe it should be a default everywhere. To my knowledge, it should not break any build processes, since unmerging is usually done as a manual action by a human when working on patches.

### DIFF
--- a/devutils/update_platform_patches.py
+++ b/devutils/update_platform_patches.py
@@ -63,17 +63,19 @@ def _dir_empty(path):
     return False
 
 
-def _remove_files_with_dirs(root_dir, sorted_file_iter):
+def _rename_files_with_dirs(root_dir, source_dir, sorted_file_iter):
     '''
-    Deletes a list of sorted files relative to root_dir, removing empty directories along the way
+    Moves a list of sorted files back to their original location,
+    removing empty directories along the way
     '''
     past_parent = None
     for partial_path in sorted_file_iter:
         complete_path = Path(root_dir, partial_path)
+        complete_source_path = Path(source_dir, partial_path)
         try:
-            complete_path.unlink()
+            complete_path.rename(complete_source_path)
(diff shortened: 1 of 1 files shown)


networkException:
Arch and NixOS won't be affected I believe, lets wait for everyone else to get a chance to look at this though :)

networkException:
@ungoogled-software/windows @ungoogled-software/macos

Qubik65536:
I believe macOS will also be unaffected. Worst case I will try to start a new build with this later to confirm.

teeminus:
I think windows should be unaffected as well. We first apply the common patches before (only) refreshing the platform patches to speedup refresh.

clickot:
fine for me

PF4Public:
I never used this script, so I cannot asses the changes, but if it doesn"t have any side-effects, should we merge the PR?

clickot:
@PF4Public @Ahrotahn @networkException shouldn't we merge this one?

Ahrotahn:
I've already given my approval!  @PF4Public @networkException Were there any objections to merging this?

networkException (approved):
Nope! I'd like to get the zero day bump fix out first though

Ahrotahn:
@dumbmoron When you get the time give this one more rebase over the latest changes and I'll merge when I get the chance!

PR #3519, opened 2025-10-26
Enable NoCrossOriginReferrers flag by default

Integral-Tech:
Disable cross-origin referrers by default while keeping same-origin referrers enabled to balance privacy and usability.


networkException:
This would actively break sites, we can't make such decisions on behalf of our users.

Integral-Tech:
Okay, I see :)

PR #3552, opened 2025-11-21
Update to Chromium 143.0.7499.40

Ahrotahn:
This PR updates to Chromium 143

Notes:

* The inox safebrowsing patch was updated for [[7014040]](https://chromium-review.googlesource.com/c/chromium/src/+/7014040)

* The trk patch had changes for [[7044668]](https://chromium-review.googlesource.com/c/chromium/src/+/7044668)

* The pruned binaries patch has new changes for [[7046577]](https://chromium-review.googlesource.com/c/chromium/src/+/7046577), and [[7002822]](https://chromium-review.googlesource.com/c/chromium/src/+/7002822)

* The safebrowsing patch had changes for [[6920366]](https://chromium-review.googlesource.com/c/chromium/src/+/6920366), [[7093137]](https://chromium-review.googlesource.com/c/chromium/src/+/7093137), and [[6990731]](https://chromium-review.googlesource.com/c/chromium/src/+/6990731)

* The remove unused prefs patch had changes for [[7048039]](https://chromium-review.googlesource.com/c/chromium/src/+/7048039), and [[7067894]](https://chromium-review.googlesource.com/c/chromium/src/+/7067894)

* The fingerprinting rects and mesasuretext patch was updated for the changes in [[6998835]](https://chromium-review.googlesource.com/c/chromium/src/+/6998835), and [[7083648]](https://chromium-review.googlesource.com/c/chromium/src/+/7083648)

* The max connections per host patch was updated for [[7050688]](https://chromium-review.googlesource.com/c/chromium/src/+/7050688)

* The add-credits patch was updated for the changes in [[7062893]](https://chromium-review.googlesource.com/c/chromium/src/+/7062893)

* The close confirmation patch was updated for [[7082325]](https://chromium-review.googlesource.com/c/chromium/src/+/7082325)

* The tabsearch button patch was updated for [[7006660]](https://chromium-review.googlesource.com/c/chromium/src/+/7006660)

* The paste and go patch was updated for [[7047457]](https://chromium-review.googlesource.com/c/chromium/src/+/7047457)

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/142.0.7444.175..143.0.7499.40?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2025/12/stable-channel-update-for-desktop.html)


Ahrotahn:
Just after making this PR my computer had hardware failure (I currently believe the motherboard is the problem) so it might be a bit before I can update this.  There is a difference in files between the tarball and clones which is why the check failed, so if you're testing these changes using the tarball you'll need to update the lists first.

PF4Public:
@Ahrotahn Take your time, there is plenty of time till 2nd :)

PF4Public:
~Had anyone tried building? I get~

```c
ERROR at //components/optimization_guide/core/inference/BUILD.gn:8:1: Assertion failed.
assert(build_with_tflite_lib)
^-----
See //media/webrtc/BUILD.gn:29:5: which caused the file to be included.
    "//components/optimization_guide/core/inference:op_resolver",
    ^-----------------------------------------------------------

```
~due to `build_with_tflite_lib` being false.~

Disregard the above. Patch wasn"t applied :-/

Ahrotahn:
The new motherboard came in yesterday.  Unfortunately I've learned that my power supply was the root cause since it took out my hard drive and caused problems with the new hardware.  

Luckily I had also ordered an adapter that allows me to access my backup drive from my steamdeck which I've been using in the interim.  That allowed me to get my keys set up on the deck so can work the repo again.  

I'll be spending time over the next few days attempting to get my computer up with some old spare parts, and I should hopefully be able to update the PR soon whether it's from my main rig or the deck.

Ahrotahn:
I've rebased over the new commits and updated the contingent paths to include the changes in [[7167308]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/7167308).  

Also, the missing `dawn_commit_hash.h` is now included in the tarballs due to [[7147287]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/7147287), thanks @iskunk!

madscientist159:
I'm seeing this near the end of the build:

```
../../chrome/browser/content_settings/generated_javascript_optimizer_pref.cc:31:15: error: no member named 'kJavascriptOptimizerBlockedForUnfamiliarSites' in namespace 'prefs'
   31 |       {prefs::kJavascriptOptimizerBlockedForUnfamiliarSites,
      |        ~~~~~~~^
../../chrome/browser/content_settings/generated_javascript_optimizer_pref.cc:74:14: error: no member named 'kJavascriptOptimizerBlockedForUnfamiliarSites' in namespace 'prefs'
   74 |       prefs::kJavascriptOptimizerBlockedForUnfamiliarSites,
      |       ~~~~~~~^
2 errors generated.
```

It looks like `kJavascriptOptimizerBlockedForUnfamiliarSites` is part of `safe_browsing_prefs.h` and might need to be stripped out as well?

Ahrotahn:
We added [move-js-optimizer-unfamiliar-sites.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/patches/core/ungoogled-chromium/move-js-optimizer-unfamiliar-sites.patch) last major version which should handle that.  Double-check that it's being applied.

madscientist159:
> We added [move-js-optimizer-unfamiliar-sites.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/patches/core/ungoogled-chromium/move-js-optimizer-unfamiliar-sites.patch) last major version which should handle that. Double-check that it's being applied.

That was it, thanks for the hint!  Had just missed that one patch -- with it applied, everything checks out on Debian Sid/ppc64le.

clickot:
builds and runs fine for portable linux

networkException (approved):

implicitfield:
Had to add a new patch to fix another case of a missing `re2/stringpiece.h` dependency:
```patch
--- a/components/contextual_tasks/internal/BUILD.gn
+++ b/components/contextual_tasks/internal/BUILD.gn
@@ -43,6 +43,7 @@ source_set("internal") {
     "//components/sync/model",
     "//components/url_formatter",
     "//net",
+    "//third_party/re2",
     "//ui/gfx",
     "//url",
   ]
```
Otherwise almost done building for macOS (linking now).

teeminus:
> Had to add a new patch to fix another case of a missing `re2/stringpiece.h` dependency:

Had to apply the same patch on windows as well. Besides that, it builds fine.

networkException:
I'd keep the patch in the platform repos then as it doesn't seem to be required on Linux, unless there are objections I'll merge this in a few hours

Thanks everyone!

PF4Public:
> it doesn't seem to be required on Linux

I actually had to patch that on Gentoo too 🤔

networkException:
hm wasn't required on NixOS... anyhow

iskunk:
The re2 dependency should be patched in, IMO. I'm seeing this in my Ubuntu builds:
```
FAILED: obj/components/contextual_tasks/internal/internal/account_utils.o 
clang++-19 -MD -MF obj/components/contextual_tasks/internal/internal/account_utils.o.d -D__STDC_CONSTANT_MACROS ...
../../third_party/re2/src/re2/re2.h:223:10: fatal error: 're2/stringpiece.h' file not found
  223 | #include "re2/stringpiece.h"
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

The `account_utils.cc` source file uses re2 unconditionally, so this is ultimately a matter of the listed dependencies being incomplete. It seems that re2 gets pulled in sometimes by the vagaries of transitive-dependency chains, but that's not something you want to rely on.

I've opened [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/7233853) to get the problem resolved upstream.

[This old bug report](https://issues.chromium.org/issues/381127086) describes a similar issue in an adjacent component, and makes mention of the transitive dependency chain involved.

Ahrotahn:
Sure, I can add the change with the next update.  They've had these kind of issues before and is the reason for the missing-dependencies patch.

wojnilowicz:
> > it doesn't seem to be required on Linux
> 
> I actually had to patch that on Gentoo too 🤔

It fails to build on Fedora and EPEL. @networkException could you add the patch as required?

arenekosreal:
I am building ungoogled-chromium on archlinux myself to avoid creating `.pki` directory.  And I am also hit by the `re2/stringpiece.h` problem. Applying patch mentioned above can fix it. 

However, noticed that `extra/chromium` is upgraded to `143.0.7499.40` normally, I am not sure if it is because I am downloading tarball from https://commondatastorage.googleapis.com/chromium-browser-official/chromium-$pkgver-lite.tar.xz directly, while archlinux uses the `fetch-chromium-release` script, which calls `gclient` to download source code.

Ahrotahn:
@wojnilowicz We'll most likely add the change the day after tomorrow.  

@arenekosreal The build order determines whether you'll run into the issue, so even something like the number of parallel jobs you are able to build with can have an effect.  I remember there was a change to how ninja ordered jobs that exposed a bunch of other missing dependencies some time ago.  There will be differences in that order compared to vanilla chromium since we change what components are built.

PR #3555, opened 2025-11-24
Fix zip compression level

Nifury:
When `ZipFile.writestr` takes a `ZipInfo` instance, it reads the compress_type member which by default is set to ZIP_STORED. The result zip file is uncompressed on Windows. 

Fixes https://github.com/ungoogled-software/ungoogled-chromium-windows/issues/504


teeminus:
Have you tested how big the archive will be in the future?

Nifury:
From ~370 MB down to ~170 MB.

networkException (approved):

clickot:
who merges this one ? :-)

PR #3558, opened 2025-11-30
patches/fixes: clean up obsolete upstream fixes

dumbmoron:
These are not needed anymore AFAICT, ungoogled-chromium builds fine for me on Linux and macOS without them.


Ahrotahn:
I'm able to build fine without `hardware_destructive_interference_size.patch` but removing `glue_core_pools.patch` gives me:
```
ERROR at //base/allocator/partition_allocator/partition_alloc.gni:242:47: Undefined identifier
assert(!enable_pointer_compression_support || glue_core_pools,
                                              ^--------------
See //gin/BUILD.gn:8:1: whence it was imported.
import("//base/allocator/partition_allocator/partition_alloc.gni")
^----------------------------------------------------------------
See //BUILD.gn:514:9: which caused the file to be included.
        "//gin:gin_unittests",
```
Do you have any of the `use_partition_alloc*` flags set?  
Maybe there's something specific with my setup that is causing those undefined variables to not be ignored.

dumbmoron:
> Do you have any of the `use_partition_alloc*` flags set? Maybe there's something specific with my setup that is causing those undefined variables to not be ignored.

Not as far as I can tell, I build with these flags:
```
build_with_tflite_lib=false
chrome_pgo_phase=0
clang_use_chrome_plugins=false
disable_fieldtrial_testing_config=true
enable_hangout_services_extension=false
enable_mdns=false
enable_remoting=false
enable_reporting=false
enable_service_discovery=false
enable_widevine=true
exclude_unwind_tables=true
google_api_key=""
google_default_client_id=""
google_default_client_secret=""
safe_browsing_mode=0
treat_warnings_as_errors=false
use_official_google_api_keys=false
use_unofficial_version_number=false
blink_symbol_level=0
enable_iterator_debugging=false
enable_mse_mpeg2ts_stream_parser=true
enable_rust=true
enable_swiftshader=true
enable_updater=false
fatal_linker_warnings=false
ffmpeg_branding="Chrome"
is_clang=true
is_debug=false
is_official_build=true
proprietary_codecs=true
symbol_level=1
use_sysroot=false
```

implicitfield:
> removing glue_core_pools.patch gives me:
> ```
> ERROR at //base/allocator/partition_allocator/partition_alloc.gni:242:47: Undefined identifier
> assert(!enable_pointer_compression_support || glue_core_pools,
>                                               ^--------------
> See //gin/BUILD.gn:8:1: whence it was imported.
> import("//base/allocator/partition_allocator/partition_alloc.gni")
> ^----------------------------------------------------------------
> See //BUILD.gn:514:9: which caused the file to be included.
>         "//gin:gin_unittests",
> ```
That seems odd, since `gn` seems to observe the usual boolean short-circuiting rules (it even goes as far as ignoring totally bogus identifiers if the first few conditions are enough to fulfill the entire statement). Nothing in Chromium appears set `enable_pointer_compression_support` to anything other than false, so I guess the only way that that could be true would be if your config specifically enables that.

Ahrotahn (approved):
That was it, now it builds and runs fine.  
Sorry for the noise!

PF4Public:
@Ahrotahn Was there any particular reason for having that option set?

PF4Public (approved):

Ahrotahn:
It reduces Chromium's memory usage by about 20%.  I remember reading an interesting writeup about it on HN ages ago.  Either I still had it set from then or I re-added it by using some old configs when I was setting everything back up from my recent hardware failure.

PR #3563, opened 2025-12-03
patches/disable-crash-reporter: re-enable crashpad on Linux

### LINKED ISSUE #3333: Tab crashes with SIGSEGV on eBay's challange page
### OS/Platform

Arch Linux

### Installed

Other (please use 'Additional context' field below to mention how exactly)

### Version

137.0.7151.55

### Description

When ebay.com sends you to a challenge page, the tab crashes with a SIGSEGV.

### How to Reproduce?

1. Go to https://www.ebay.com/splashui/challenge?ap=1&appName=orch&ru=https%3A%2F%2Fwww.ebay.com
2. Wait till tab crashes with SIGSEGV

### Actual behaviour

When visiting the splash page for the bot challenge by eBay, the animation is not present and the tab crashes. This behavior does not happen with Google Chrome, Chromium, or Brave. It does happen with an empty profile.

### Expected behaviour

eBay challenge passes successfully and I get to see the page.

### Relevant log output

```sh

```

### Additional context

The ungoogled-chromium versions I tested were ungoogled-chromium-bin from the AUR (or https://github.com/ungoogled-software/ungoogled-chromium-archlinux) and ungoogled-chromium from nixpkgs. In both versions the tab crashed, so the problem does not seem to be with the AUR binary.

dumbmoron:
Fixes #3333 by re-enabling Crashpad specifically on Linux, since Chromium relies on it (for whatever reason) as a [signal handler for WASM JIT stuff](https://source.chromium.org/chromium/chromium/src/+/main:chrome/app/chrome_main_delegate.cc;l=1576;drc=b2d540b15cad394746b9277911ae1192f25158aa;bpv=0;bpt=1). 

The crashpad handler should not generate any traffic even by default, since the URL is currently gated behind `BUILDFLAG(GOOGLE_CHROME_BRANDING)` (and domain substitution should otherwise catch it anyways), but I added some safeguards just in case anything changes.

### DIFF
--- a/patches/core/ungoogled-chromium/disable-crash-reporter.patch
+++ b/patches/core/ungoogled-chromium/disable-crash-reporter.patch
@@ -1,69 +1,49 @@
 # Disable some background communication with clients2.google.com
 
---- a/chrome/browser/chrome_content_browser_client.cc
-+++ b/chrome/browser/chrome_content_browser_client.cc
-@@ -2683,24 +2683,6 @@ void ChromeContentBrowserClient::AppendE
-     command_line->AppendSwitchASCII(switches::kMetricsClientID,
-                                     client_info->client_id);
-   }
--#elif BUILDFLAG(IS_POSIX)
--#if !BUILDFLAG(IS_ANDROID)
--  pid_t pid;
--  if (crash_reporter::GetHandlerSocket(nullptr, &pid)) {
--    command_line->AppendSwitchASCII(
--        crash_reporter::switches::kCrashpadHandlerPid,
--        base::NumberToString(pid));
--  }
--#endif
--  std::string switch_value;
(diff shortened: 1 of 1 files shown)


PF4Public:
Do we also need to build and ship `out/Release/chrome_crashpad_handler`?

dumbmoron:
> Do we also need to build and ship `out/Release/chrome_crashpad_handler`?

Yes, Chromium tries to launch it and `CHECK()`s if it was launched, so it would crash if it was absent. But from a cursorily glance, it looks like {portable,arch}linux [already ship it](https://github.com/search?q=org%3Aungoogled-software%20chrome_crashpad_handler&type=code), so hopefully it wouldn't be too big of an inconvenience (and I believe it should be already built when building the `chrome` target).

PF4Public (approved):

PR #3588, opened 2025-12-30
Ability to fetch version based on feature branch, rather than hardcode it

lifesminder:
I have added ability to check the latest version of Chromium from it's official Appspot, in order to gradually eliminate need of `chromium_version.txt` and add ability for users to select one of appropriate branches:

* Stable,
* Beta,
* Dev,
* Canary.

If you want I can make ability to turn it into a custom version if it differs from these 4 described above

### DIFF
--- a/utils/_common.py
+++ b/utils/_common.py
@@ -8,6 +8,8 @@ import argparse
 import enum
 import logging
 import platform
+import requests
+import json
 from pathlib import Path
 
 # Constants
@@ -96,13 +98,16 @@ def set_logging_level(logging_level):
     return logger
 
 
-def get_running_platform():
+def get_running_platform(mode="enum"):
     """
     Returns a PlatformEnum value indicating the platform that utils is running on.
 
     NOTE: Platform detection should only be used when no cross-platform alternative is available.
     """
     uname = platform.uname()
+    # check if mode is `native`. if yes, return the literal name
+    if mode == "native":
+        return uname.system
     # detect native python and WSL
     if uname.system == 'Windows' or 'Microsoft' in uname.release:
         return PlatformEnum.WINDOWS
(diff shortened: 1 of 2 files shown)


teeminus:
Wouldn't we loose the ability to build older versions of chromium that way?

dumbmoron:
This would likely not be very useful, because the patches are prepared for a specific version. To apply them to a different version - usually, but not always - requires some manual work, so you cannot just swap out the underlying tarball and build e.g. for a newer or older milestone.

networkException:
nixpkgs is currently relying on the txt file iirc, would need investigating

lifesminder:
> Wouldn't we loose the ability to build older versions of chromium that way?

this can be uplifted to mention specific version you want, so it won't make anything

lifesminder:
> This would likely not be very useful, because the patches are prepared for a specific version. To apply them to a different version - usually, but not always - requires some manual work, so you cannot just swap out the underlying tarball and build e.g. for a newer or older milestone.

oh

Ahrotahn:
Yeah, another way to think about chromium_version.txt is "this is the version these patches apply to" rather than "this is the version I wish to build".  

We tag the releases so checking out the tag for prior releases would allow you to build older versions.  
New stable versions require a lot of work on their own so I can't imagine we'd be able to support any of the other channels, but we would certainly welcome the increase in bus factor for anyone willing to put in that kind of effort.

lifesminder:
> Yeah, another way to think about chromium_version.txt is "this is the version these patches apply to" rather than "this is the version I wish to build".
> 
> We tag the releases so checking out the tag for prior releases would allow you to build older versions. New stable versions require a lot of work on their own so I can't imagine we'd be able to support any of the other channels, but we would certainly welcome the increase in bus factor for anyone willing to put in that kind of effort.

got it thanks

clickot:
shouldn't this PR be closed? @lifesminder Would you be willing to close this PR?

lifesminder:
ok

Issue #2869, opened 2024-05-17
Google sign-in prompt shows up on some websites


Swarkin:
### OS/Platform

Windows

### Installed

https://ungoogled-software.github.io/ungoogled-chromium-binaries/

### Version

Version 122.0.6261.94 (Official Build, ungoogled-chromium) (64-bit)

### Have you tested that this is not an upstream issue or an issue with your configuration?

- [ ] I have tried reproducing this issue in Chrome and it could not be reproduced there
- [ ] I have tried reproducing this issue in vanilla Chromium and it could not be reproduced there
- [ ] I have tried reproducing this issue in ungoogled-chromium with a new and empty profile using `--user-data-dir` command line argument and it could not be reproduced there

### Description

Google sign-in prompt shows up on some websites.

### How to Reproduce?

Go to medium.com
![image](https://github.com/ungoogled-software/ungoogled-chromium/assets/102416174/8226c7b7-2b16-423a-8299-8a0d9fbda604)

### Actual behaviour

Google sign-in prompt shows up on some websites.

### Expected behaviour

I expected ungoogled-chromium to be ungoogled

### Additional context

Is this left in for a reason? The browser managed to extract my account info from presumably YouTube and now shows it in a small popup when websites support Google sign-in.

Swarkin:
It is part of the interface, it sometimes even spawns outside of the browser's window area because of some incorrect position calculation and hovers above the browser contents itself

Swarkin:
It even shows up when I delete the entire html document:
![image](https://github.com/ungoogled-software/ungoogled-chromium/assets/102416174/43c18ac8-d399-40d7-bcb1-0976c63d0447)
It is a native browser popup

Swarkin:
Found the culprit and honestly surprised this isnt off by default or even removed entirely.
![WMLSszAkgS](https://github.com/ungoogled-software/ungoogled-chromium/assets/102416174/8feaa70d-1d32-49d8-bd41-2e019a60d649)

PF4Public:
@networkException @Ahrotahn Is it worth flipping this setting by default?

Ahrotahn:
With our changes the identity provider should return that the profile is not signed in.  Granted, we should probably add that section to the UI removal patch since it doesn't, or at least *shouldn't*, do anything.  

I tested with my local linux build as well as the windows builds for 122.0.6261.94-1 and 124.0.6367.207-1 from our binaries page and I'm not able to reproduce this.  

[...]

PR #3625, opened 2026-01-26
Disable Federated Credential Management (FedCM) by default

sqikerz:
solves #2869, as it is still an issue, although marked as closed.

This pull request makes a privacy-focused change to the ungoogled-chromium patch set by disabling Federated Credential Management (FedCM) by default. This feature is now turned off unless explicitly enabled, further reducing the browser's default surface for federated identity tracking.

Privacy and feature changes:

* Added a new patch, `disable-fedcm-by-default.patch`, which changes the default value of the `kFedCm` feature flag in `content_features.cc` from enabled to disabled. This disables FedCM by default for all users.
* Registered the new patch in the `patches/series` file to ensure it is applied as part of the ungoogled-chromium build process.

### DIFF
--- a/docs/default_settings.md
+++ b/docs/default_settings.md
@@ -17,5 +17,5 @@ Payment autofill | Disabled | chrome://settings/payments
 Preload pages | Disabled | chrome://settings/performance
 Search suggestions | Disabled | chrome://settings/syncSetup
 Show bookmark bar | Enabled | chrome://settings/appearance
-Third-party sign-in | Disabled | `--enable-features=FedCm-`
+Third-party sign-in | Disabled | `--enable-features=FedCm`
 WebRTC IP handling policy | Disable Non-Proxied UDP | `--webrtc-ip-handling-policy`


sqikerz:
I built (on windows) and tested, and it is working.

PF4Public:
🤔 I thought we disabled federated-thingy in one of the patches. @Ahrotahn do you remember, where?

sqikerz:
> 🤔 I thought we disabled federated-thingy in one of the patches. @Ahrotahn do you remember, where?

I believe it was here, but as I've seen (and few others reported) the sign-in prompt came back, at least this past week.

https://github.com/ungoogled-software/ungoogled-chromium/pull/2872

sqikerz:
Also just a thought, I believe it would be better UX for the FedCM toggle in settings is brought back, so users are able to re-enable it. The current patch in the codebase (prior to this PR) only removes the toggle, but the feature is still enabled by default.

I can add the change, it's a simple switch:
{03AEA5FB-1B49-40D3-BFE5-9160A520A9F8}

Ahrotahn:
This seems fine to me.  It would be better to remove that hunk in the remove_uneeded_ui patch instead of setting it to true, that way it will correctly reflect the state of the flag.  If you get the chance it would be great if you could also update [docs/default_settings.md](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/default_settings.md).

sqikerz:
> This seems fine to me. It would be better to remove that hunk in the remove_uneeded_ui patch instead of setting it to true, that way it will correctly reflect the state of the flag. If you get the chance it would be great if you could also update [docs/default_settings.md](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/default_settings.md).

Done, for location on the [docs/default_settings.md](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/docs/default_settings.md) I put NA, but the feature could be re-enabled by passing ``--enable-features=FedCm`` as a launch arg.

PF4Public:
> I put NA, but the feature could be re-enabled by passing `--enable-features=FedCm` as a launch arg

You could put the commandline parameter instead of NA then.

sqikerz:
Sorry for the double commit, on the app right now 😅

PF4Public:
No worries, we can squash commits

PF4Public (approved):
Thanks!

PR #3632, opened 2026-01-28
Add quilt(1) backup option to make_domsub_script.py

iskunk:
This allows files modified by the script to be backed up using quilt's convention for saving a copy of files prior to patching them. The backup is recorded as a virtual patch file, that can then be popped off the stack (using `quilt pop -f`) like any other patch.

----

Been sitting on this one for some time. This allows better integration with the Debian build, which already uses a `quilt` equivalent to apply patches to the tree. The `tar` backup method remains as the default option.


iskunk:
Thanks everyone!

PR #3634, opened 2026-01-31
Update to Chromium 145.0.7632.45

Ahrotahn:
This PR updates to Chromium 145.0.7632.45.

Normally I start work on major versions after the stable cut so there's fewer changes that need to be worked on, and normally the stable release is the following week.  It seems the window between the two has been extended to two weeks which gives us more time to work with!  

<br>

**Patch changes**:
* build-with-wasm-rollup.patch has been updated to use the rollup wasm instead of the pruned binaries for [[7368549]](https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/7368549) and [[7414812]](https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/7414812)

* disable-fonts-googleapis-references had a removal for [[7205386]](https://chromium-review.googlesource.com/c/chromium/src/+/7205386)

* disable-google-host-detection was updated for [[7275378]](https://chromium-review.googlesource.com/c/chromium/src/+/7275378) and [[7401580]](https://chromium-review.googlesource.com/c/chromium/src/+/7401580)

* The safebrowsing patch was updated for [[7358844]](https://chromium-review.googlesource.com/c/chromium/src/+/7358844), [[7083567]](https://chromium-review.googlesource.com/c/chromium/src/+/7083567), [[7168455]](https://chromium-review.googlesource.com/c/chromium/src/+/7168455), and [[7152278]](https://chromium-review.googlesource.com/c/chromium/src/+/7152278)
  There's also a good amount of removals now that most of @dumbmoron's changes have landed:  [[7277267]](https://chromium-review.googlesource.com/c/chromium/src/+/7277267), [[7277410]](https://chromium-review.googlesource.com/c/chromium/src/+/7277410), [[7278214]](https://chromium-review.googlesource.com/c/chromium/src/+/7278214), [[7278216]](https://chromium-review.googlesource.com/c/chromium/src/+/7278216), [[7278256]](https://chromium-review.googlesource.com/c/chromium/src/+/7278256), [[7278316]](https://chromium-review.googlesource.com/c/chromium/src/+/7278316), and [[7280949]](https://chromium-review.googlesource.com/c/chromium/src/+/7280949)

* The unused prefs patch was updated for [[7313802]](https://chromium-review.googlesource.com/c/chromium/src/+/7313802), [[7268724]](https://chromium-review.googlesource.com/c/chromium/src/+/7268724), [[7358156]](https://chromium-review.googlesource.com/c/chromium/src/+/7358156), [[7361845]](https://chromium-review.googlesource.com/c/chromium/src/+/7361845), [[7228206]](https://chromium-review.googlesource.com/c/chromium/src/+/7228206), and [[7506390]](https://chromium-review.googlesource.com/c/chromium/src/+/7506390)

* The reduced system info patch was updated for [[7179481]](https://chromium-review.googlesource.com/c/chromium/src/+/7179481) and [[7276122]](https://chromium-review.googlesource.com/c/chromium/src/+/7276122)

* disable-downloads-page-referrer-url.patch has been removed as the changes are now the default:  [(424764882)](https://issues.chromium.org/issues/424764882)

* enable-paste-and-go-new-tab-button.patch now includes a small revert of the changes in [[7237122]](https://chromium-review.googlesource.com/c/chromium/src/+/7237122)

<br>

**Script changes**:
* validate_patches.py has been updated to work with python 3.14

* The clone script has been updated to use gsutil 5.35, which also allowed the removal of the newer apitools commit checkout

* A new patch for the clone script has been added that allows gsutil to work with python 3.14

<br>

**Build environment notes**:
* If your system's version of `gn` is not new enough then it may not have functions available that are now being used in [[7226634]](https://chromium-review.googlesource.com/c/chromium/src/+/7226634) and [[7242247]](https://chromium-review.googlesource.com/c/chromium/src/+/7242247).  This patch reverts the relevant changes:

  ```diff
  --- a/build/rust/rust_bindgen.gni
  +++ b/build/rust/rust_bindgen.gni
  @@ -112,7 +112,6 @@
     # Calculate `_generated_bindings_rs`.
     _bindgen_outputs = get_target_outputs(":$_rust_bindgen_generator_target_name")
     _bindgen_rs_outputs = filter_include(_bindgen_outputs, [ "*.rs" ])
  -  assert(len(_bindgen_rs_outputs) == 1)
     _generated_bindings_rs = _bindgen_rs_outputs[0]
   
     rust_static_library(target_name) {
  --- a/build/rust/chromium_prelude/import_attribute.rs
  +++ b/build/rust/chromium_prelude/import_attribute.rs
  @@ -72,8 +72,8 @@
               }
           }
   
  -        let gn_dir = format!("//{}", path.join("/"));
  -        let mangled_crate_name = mangle_crate_name(&gn_dir, gn_name);
  +        let mangled_crate_name =
  +            escape_non_identifier_chars(&format!("{}:{gn_name}", path.join("/")))?;
   
           Ok(GnTarget {
               mangled_crate_name: Ident::new(&mangled_crate_name, span),
  @@ -120,42 +120,52 @@
       }
   }
   
  -/// Mangles a GN target path into a crate name that is globally unique
  -/// ("globally" means: within a GN build graph).
  +/// Escapes non-identifier characters in `symbol`.
   ///
  -/// NOTE: When updating GnTarget=>CrateName mangling algorithm, it needs to
  -/// be updated and kept in sync in 3 places: this function,
  -/// `//build/rust/rust_target.gni`, `//build/rust/rust_static_library.gni`.
  +/// Importantly, this is
  +/// [an injective function](https://en.wikipedia.org/wiki/Injective_function)
  +/// which means that different inputs are never mapped to the same output.
   ///
  -/// # Example
  +/// This is based on a similar function in
  +/// https://github.com/google/crubit/blob/22ab04aef9f7cc56d8600c310c7fe20999ffc41b/common/code_gen_utils.rs#L59-L71
  +/// The main differences are:
   ///
  -/// ```
  -/// // Calculate mangled name for the
  -/// // `//build/rust/chromium_prelude:import_test_lib` GN target:
  -/// let mangled = mangle_crate_name("//build/rust/chromium_prelude", "import_test_lib");
  -/// assert_eq!(mangled, "import_test_lib_8086ab2e");
  -/// ```
  -fn mangle_crate_name(dir: &str, target_name: &str) -> String {
  -    assert!(!target_name.is_empty()); // Caller is expected to verify.
  -    let dir_hash = gn_string_hash(dir);
  -    format!("{target_name}_{dir_hash}")
  -}
  +/// * Only a limited set of special characters is supported, because this makes
  +///   it easier to replicate the escaping algorithm in `.gni` files, using just
  +///   `string_replace` calls.
  +/// * No dependency on `unicode_ident` crate means that instead of
  +///   `is_xid_continue` a more restricted call to `char::is_ascii_alphanumeric`
  +///   is used.
  +/// * No support for escaping leading digits.
  +/// * The escapes are slightly different (e.g. `/` frequently appears in GN
  +///   paths and therefore here we map it to a nice `_s` rather than to `_x002f`)
  +fn escape_non_identifier_chars(symbol: &str) -> Result<String, String> {
  +    assert!(!symbol.is_empty()); // Caller is expected to verify.
  +    if symbol.chars().next().unwrap().is_ascii_digit() {
  +        return Err("Leading digits are not supported".to_string());
  +    }
   
  -/// `fn gn_string_hash` replicates `string_hash` from GN [1] which computes
  -/// the first 8 hexadecimal digits of an SHA256 hash of a string.
  -///
  -/// [1] https://gn.googlesource.com/gn/+/main/docs/reference.md#func_string_hash
  -fn gn_string_hash(s: &str) -> String {
  -    let sha256 = {
  -        let mut hasher = hmac_sha256::Hash::new();
  -        hasher.update(s.as_bytes());
  -        hasher.finalize()
  -    };
  -    let mut result = String::with_capacity(8);
  -    for byte in sha256.into_iter().take(4) {
  -        // `unwrap` is okay because writing into a `String` cannot fail.
  -        use std::fmt::Write;
  -        write!(&mut result, "{byte:02x}").unwrap();
  +    // Escaping every character can at most double the size of the string.
  +    let mut result = String::with_capacity(symbol.len() * 2);
  +    for c in symbol.chars() {
  +        // NOTE: TargetName=>CrateName mangling algorithm should be updated
  +        // simultaneously in 3 places: here, //build/rust/rust_target.gni,
  +        // //build/rust/rust_static_library.gni.
  +        match c {
  +            '_' => result.push_str("_u"),
  +            '/' => result.push_str("_s"),
  +            ':' => result.push_str("_c"),
  +            '-' => result.push_str("_d"),
  +            c if c.is_ascii_alphanumeric() => result.push(c),
  +            _ => return Err(format!("Unsupported character in GN path component: `{c}`")),
  +        }
       }
  -    result
  +    // The string replacements below are introduced to decrease the number
  +    // of characters in the crate name. This is particularly important while
  +    // building Chromium with lld_emit_indexes_and_imports=True, since it
  +    // will potentially need to generate huge filenames, and the build will
  +    // fail in some filesystems.
  +    result = result.replace("third_uparty_srust_s", "_tpr");
  +
  +    Ok(result)
   }
  --- a/build/rust/gni_impl/rust_target.gni
  +++ b/build/rust/gni_impl/rust_target.gni
  @@ -48,8 +48,31 @@
     if (defined(invoker.crate_name)) {
       _crate_name = invoker.crate_name
     } else {
  -    _dir_hash = string_hash(get_label_info(":${_target_name}", "dir"))
  -    _crate_name = "${_target_name}_${_dir_hash}"
  +    # Not using `get_label_info(..., "label_no_toolchain")` to consistently
  +    # use `//foo/bar:baz` instead of the alternative shorter `//foo/bar` form.
  +    _dir = get_label_info(":${_target_name}", "dir")
  +    _dir = string_replace(_dir, "//", "")
  +    _crate_name = "${_dir}:${_target_name}"
  +
  +    # The `string_replace` calls below replicate the escaping algorithm
  +    # from the `escape_non_identifier_chars` function in
  +    # //build/rust/chromium_prelude/import_attribute.rs.  Note that the
  +    # ordering of `match` branches within the Rust function doesn't matter,
  +    # but the ordering of `string_replace` calls *does* matter - the escape
  +    # character `_` needs to be handled first to meet the injectivity
  +    # requirement (otherwise we would get `/` => `_s` => `_us` and the same
  +    # result for `_s` => `_us`).
  +    _crate_name = string_replace(_crate_name, "_", "_u")
  +    _crate_name = string_replace(_crate_name, "/", "_s")
  +    _crate_name = string_replace(_crate_name, ":", "_c")
  +    _crate_name = string_replace(_crate_name, "-", "_d")
  +
  +    # The string replacements below are introduced to decrease the number
  +    # of characters in the crate name. This is particularly important while
  +    # building Chromium with lld_emit_indexes_and_imports=True, since it
  +    # will potentially need to generate huge filenames, and the build will
  +    # fail in some filesystems.
  +    _crate_name = string_replace(_crate_name, "third_uparty_srust_s", "_tpr")
     }
     _generate_crate_root =
         defined(invoker.generate_crate_root) && invoker.generate_crate_root
  --- a/build/rust/rust_static_library.gni
  +++ b/build/rust/rust_static_library.gni
  @@ -187,8 +187,32 @@
         if (defined(invoker.crate_name)) {
           _crate_name = invoker.crate_name
         } else {
  -        _dir_hash = string_hash(get_label_info(":${_target_name}", "dir"))
  -        _crate_name = "${_target_name}_${_dir_hash}"
  +        # Not using `get_label_info(..., "label_no_toolchain")` to consistently
  +        # use `//foo/bar:baz` instead of the alternative shorter `//foo/bar` form.
  +        _dir = get_label_info(":${_target_name}", "dir")
  +        _dir = string_replace(_dir, "//", "")
  +        _crate_name = "${_dir}:${_target_name}"
  +
  +        # The `string_replace` calls below replicate the escaping algorithm
  +        # from the `escape_non_identifier_chars` function in
  +        # //build/rust/chromium_prelude/import_attribute.rs.  Note that the
  +        # ordering of `match` branches within the Rust function doesn't matter,
  +        # but the ordering of `string_replace` calls *does* matter - the escape
  +        # character `_` needs to be handled first to meet the injectivity
  +        # requirement (otherwise we would get `/` => `_s` => `_us` and the same
  +        # result for `_s` => `_us`).
  +        _crate_name = string_replace(_crate_name, "_", "_u")
  +        _crate_name = string_replace(_crate_name, "/", "_s")
  +        _crate_name = string_replace(_crate_name, ":", "_c")
  +        _crate_name = string_replace(_crate_name, "-", "_d")
  +
  +        # The string replacements below are introduced to decrease the number
  +        # of characters in the crate name. This is particularly important while
  +        # building Chromium with lld_emit_indexes_and_imports=True, since it
  +        # will potentially need to generate huge filenames, and the build will
  +        # fail in some filesystems.
  +        _crate_name =
  +            string_replace(_crate_name, "third_uparty_srust_s", "_tpr")
         }
   
         _rlib_path = "${_output_dir}/lib${_crate_name}.rlib"
  ```
  </details>

* The desktop file for linux has been updated to remove the trailing asperands in [[7253206]](https://chromium-review.googlesource.com/c/chromium/src/+/7253206) which might require adjustments when being replaced

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/144.0.7559.109..145.0.7632.45?pretty=fuller&n=10000)

[Blog post](https://chromereleases.googleblog.com/2026/02/stable-channel-update-for-desktop_10.html)


iskunk:
> * A native build of rollup is being used instead of the wasm version we had an exception for in [[7368549]](https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/7368549) and [[7414812]](https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/7414812)
>   I haven't found an easy way to work around this, so for now you'll need to download the binary separately ...

Could you swap in the WASM version of rollup again? They switched away from it only because of [mysterious crashes in the Windows builds](https://crbug.com/461602362). As I understand, this was more or less [how Alpine Linux resolved the issue](https://codeberg.org/selfisekai/copium/src/branch/m145/cr145-rollup-wasm.patch).

Ahrotahn:
That's certainly an option so long as our windows builds aren't also affected.  
The only downside with that approach is it adds npm as a new dependency (I hadn't installed it since my rebuild).  

There is a copy of the rollup wasm in `third_party/node/node_modules/@rollup/`, and I'm not sure why they decided to add their own copy in `third_party/devtools-frontend/src/node_modules/@rollup/` but only that one was removed.  It should be possible to make a patch to have it use the still-existing copy instead.

Ahrotahn:
I've pushed a new change that removes `devtools_bundle=false` and added a new patch to handle the rollup and esbuild changes:  [build-with-wasm-rollup.patch](https://github.com/ungoogled-software/ungoogled-chromium/blob/98a66e4128f9e3e59b28430c9224d258803564f5/patches/core/ungoogled-chromium/build-with-wasm-rollup.patch)

This uses the existing wasm version of rollup so there aren't any extra downloads or programs required.  It works without issue on my end, but it'll be good to have confirmation from the windows team that it doesn't cause any issues when they get the time to try it out.

Nifury:
Building on Windows gives this error:
```
  File "F:\ungoogled-chromium-windows\build\src\third_party\devtools-frontend\src\scripts\build\build_inspector_overlay.py", line 91, in main
    rollup(input_path, output_path, filename, max_size,
  File "F:\ungoogled-chromium-windows\build\src\third_party\devtools-frontend\src\scripts\build\build_inspector_overlay.py", line 66, in rollup
    raise Exception("rollup failed: " + error)
Exception: rollup failed: [!] Error: Cannot load plugin "file:///F:/ungoogled-chromium-windows/build/src/third_party/devtools-frontend/src/node_modules/@rollup/plugin-terser": Directory import 'F:\ungoogled-chromium-windows\build\src\third_party\devtools-frontend\src\node_modules\@rollup\plugin-terser' is not supported resolving ES modules imported from F:\ungoogled-chromium-windows\build\src\third_party\node\node_modules\@rollup\wasm-node\dist\shared\loadConfigFile.js
Did you mean to import "file:///F:/ungoogled-chromium-windows/build/src/third_party/devtools-frontend/src/node_modules/@rollup/plugin-terser/dist/cjs/index.js"?.
```

Just need to path `third_party/devtools-frontend/src/scripts/build/build_inspector_overlay.py`

```diff
-        ['--plugin', rollup_plugin, '--plugin', path.join(devtools_paths.node_modules_path(), '@rollup', 'plugin-terser')],
+        ['--plugin', rollup_plugin, '--plugin', path.join(devtools_paths.node_modules_path(), '@rollup', 'plugin-terser', 'dist', 'cjs', 'index.js')],
```

Ahrotahn:
Thanks @Nifury, I've updated the patch with the change.

Nifury:
Thanks @Ahrotahn. 

I also ran into the following build error:

```
error: let chains are only allowed in Rust 2024 or later
  --> ../../components/user_data_importer/utility/parsing_ffi\zip_archive.rs:89:16
   |
89 |             if let Ok(file) = self.archive.by_index(i)
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: let chains are only allowed in Rust 2024 or later
  --> ../../components/user_data_importer/utility/parsing_ffi\zip_archive.rs:90:20
   |
90 |                 && let Some(outpath) = file.enclosed_name()
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: let chains are only allowed in Rust 2024 or later
  --> ../../components/user_data_importer/utility/parsing_ffi\zip_archive.rs:91:20
   |
91 |                 && let Some(r) = f(file, &outpath)
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

As a workaround, I patched `components/user_data_importer/utility/BUILD.gn` to explicitly set the Rust edition to 2024. 


```diff
 rust_static_library("zip_ffi_glue") {
+  edition = "2024"
   allow_unsafe = true
   crate_root = "parsing_ffi/lib.rs"
   sources = [
```

Everything now builds and runs fine on my end. That said, I'd like to know if there's a more appropriate or preferred way to handle this. 
image

Ahrotahn:
That seems to be more fallout from them setting `rustc_nightly_capability`.  The default is set to 2024 but it gets changed to 2021 [here](https://source.chromium.org/chromium/chromium/src/+/refs/tags/145.0.7632.26:build/rust/gni_impl/rust_target.gni;l=137-139) when the flag is false.  I guess it just wasn't a problem until they updated zip. 

The change you have would work fine for the individual package, but a better long-term solution would be to either remove that highlighted section or set the flag to true.  

If all platforms are setting the flag then maybe that change should be added to core.

clickot:
All quiet on the Portable Linux front :-)
Screenshot From 2026-02-03 13-51-20

PF4Public:
@Ahrotahn Please add the revert of changes from https://chromium.googlesource.com/devtools/devtools-frontend/+/1a823d03be536fd67bc6d002147d34ef2af21f58%5E%21/#F4 for `scripts/build/generate_css_js_files.js` into `build-with-wasm-rollup.patch`. Otherwise `typecheck` is broken without `esbuild`.

Ahrotahn:
Sure.  Do you still have a copy of the error output?  It would be good to figure out the core issue since a revert of the whole change may not be necessary.

PF4Public:
> Do you still have a copy of the error output?

~Unfortunately no, neither in `screen`"s scrollback, nor within copied text.~
I had to restart the build. Here is the error output:

> FAILED: [code=1] gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/devtools_entrypoint-bundle-tsconfig-tsconfig.json gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/code_highlighter.prebundle.js gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/code_highlighter.prebundle.js.map gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/code_highlighter.prebundle.d.ts 
> python3.13 ../../third_party/devtools-frontend/src/scripts/build/typescript/ts_library.py --tsconfig_output_location gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/devtools_entrypoint-bundle-tsconfig-tsconfig.json --deps ./code_highlighter-tsconfig.json ./css_files-tsconfig.json --front_end_directory gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter --sources gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/code_highlighter.prebundle.ts
> 
> TypeScript compilation failed. Used tsconfig gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/devtools_entrypoint-bundle-tsconfig-tsconfig.json
> 
> gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/devtools_entrypoint-bundle-tsconfig-tsconfig.json(47,5): error TS6053: File '/var/tmp/portage/www-client/ungoogled-chromium-145.0.7632.45_p1/work/chromium-145.0.7632.45/out/Release/gen/third_party/devtools-frontend/src/front_end/ui/components/code_highlighter/css_files-tsconfig.json' not found.
> 

> It would be good to figure out the core issue since a revert of the whole change may not be necessary.

It would be great if you could find a better solution. The problem is that with the changes from the commit mentioned above `third_party/devtools-frontend/src/scripts/build/generate_css_js_files.js` doesn"t generate the necessary css files and more specifically: `css_files-tsconfig.json`, essentially it does nothing. There is no error message — literally nothing. Therefore this particular step does not fail. What fails is the next one, which needs `css_files-tsconfig.json`. If you want to reproduce it, you'll need either `official_build` or `devtools_skip_typecheck` in order to enable type-checking. The latter has a convoluted and confusing logic (true/false).

Come to think of it, could it be that my `node` is old?

sqikerz:
> Come to think of it, could it be that my `node` is old?

I believe so. I am able to get further by using node.js 24.12.0, but it is still currently building. I'll edit my reply with any updates.

I forgot to set rustc_nightly_capability and now I'm recompiling 😭
{90794134-3C9F-43E9-AD12-1AC2E87D2BB2}

Ahrotahn:
The issue for the migration [(411053158)](https://issues.chromium.org/issues/411053158) mentions that node v24 is required, so if you're using an older version that is the most likely culprit.  I'd be surprised if there aren't other subtle problems that would be encountered building with an older version due to the amount of changes made.  

I'm taking a look at how setting `devtools_skip_typecheck` affects things, but in either case I think it might be more appropriate to have such a change be a platform patch unless if affects multiple builds.

PF4Public:
I'll try building with node-24. If that is the issue, no additional patch changes are necessary.

sqikerz:
By the way, the correct fix is definitely setting rustc_nightly_capability to true (or rather build_with_chromium). It was bug that was fixed upstream right after this target release:

https://github.com/chromium/chromium/commit/272bb673e5a2665321b7a62cbf75f58e60130c5d

This also is only an issue in UGC win & mac, because rust_sysroot_absolute is overrided:
```gni
# Use the Rust toolchain built in-tree. When false, we use the prebuilt Rust
# stdlibs that come with the specified custom toolchain.
use_chromium_rust_toolchain = rust_sysroot_absolute == ""

# `rustc_nightly_capability = false` will make the build avoid the use of
# Rust nightly features. There are no bots that test this and there is no
# guarantee it will work, but we accept patches for this configuration.
rustc_nightly_capability = use_chromium_rust_toolchain
```

```patch
--- a/build/config/rust.gni
+++ b/build/config/rust.gni
@@ -57,7 +57,7 @@ declare_args() {
   # build may depend on using a *nightly* version of Rust toolchain (see
   # `tools/rust/unstable_rust_feature_usage.md`) although other projects (e.g. V8)
   # may be compatible with the stable version of Rust toolchain.
-  rust_sysroot_absolute = ""
+  rust_sysroot_absolute = "//third_party/rust-toolchain"
 
   # Directory under which to find `bin/bindgen` (a `bin` directory containing
   # the bindgen exectuable).
```

PF4Public:
I can confirm that node upgrade did solve the issue with typecheck. Builds and runs.

sqikerz:
Win 64 builds and runs perfectly without any modifications to core

{C961F305-CBF9-40E3-8420-FCC31096971E}

dilinger:
FYI, for working around gn's missing string_hash, it's pretty easy to provide your own shell equivalent:

https://salsa.debian.org/chromium-team/chromium/-/blob/master/debian/patches/trixie/gn-string-hash.patch?ref_type=heads

Thought you might prefer that to a much more invasive reverting of string_hash() usage.

luke-jr:
Does this project not consider it a bug that building "from source" requires executing a binary blob (rollup)?

Ahrotahn:
It would be great to not require any blobs at all, and we strip out the majority of them, but I'm not sure we'd consider it a bug.  We already have a handful of exclusions that are required to build that you can see [here](https://github.com/ungoogled-software/ungoogled-chromium/blob/master/devutils/update_lists.py#L46-L106).  Granted, most of those are binary resources rather than executables.  

When we first added the exclusion for rollup I had checked that the hash of the binary matched what you would get from their repository.  That way we could at least verify that it is the same as the upstream project.  I just checked the sha256 for this release and it matches version 4.21.0:
```
4862964f64de5a2c46e4942daf2704cd916007635bc519980ea05d08adeb9de0  third_party/node/node_modules/@rollup/wasm-node/dist/wasm-node/bindings_wasm_bg.wasm
```

It would be preferable to not need to require this at all.  Do you know of a better alternative?

dilinger:
> 
> It would be preferable to not need to require this at all. Do you know of a better alternative?

We (Debian) are trying to figure this out now. The DFSG doesn't allow us to ship the binary unless we can build from source. So I'm currently trying to figure out the easiest path forward - bundling the (built) esbuild package, building rollup4 during the chromium build (we currently only have 3.x in the archive), or packaging up rolldown under the assumption that everyone will eventually switch to that.

luke-jr:
> It would be preferable to not need to require this at all. Do you know of a better alternative?

AFAIK these are open source projects, so can't they just get compiled?

rany2:
Tested 145.0.7632.45 on Flatpak.

iskunk:
> AFAIK these are open source projects, so can't they just get compiled?

The problem is, some of these are obnoxiously difficult to build---at least if you're not fluent in the ecosystems involved. Rollup, for example, is a hybrid of Node.js and Rust. The trick is to get a source tree that can build the tool from source, using our existing versions of both environments, without any network access. Ideally, with the source tree and built files being reproducible from a small specification.

dilinger:
btw, I did manage to get it building by copying /usr/bin/esbuild (0.27.1; I haven't yet tried with earlier versions) to third_party/devtools-frontend/src/third_party/esbuild/esbuild. Then you need to either modify 6 or 7 version strings in third_party/devtools-frontend/src/node_modules/esbuild/lib/main.js to match 0.27.1. Or, if you don't want to deal with patching main.js, you can also  `cp -ra /usr/lib/x86_64-linux-gnu/nodejs/esbuild third_party/devtools-frontend/src/node_modules/` (on a debian amd64 system).

PR #3653, opened 2026-02-12
Update to Chromium 145.0.7632.75

Ahrotahn:
This PR updates to Chromium 145.0.7632.75

Notes:

*  @iskunk's changes in [[7538780]](https://chromium-review.googlesource.com/c/chromium/tools/build/+/7538780) landed which should make the tarballs more resilient.  This causes our domain substitution script to add `testing/location_tags.json` to the list, so I've added that to the contingent paths since we don't run hooks in the clone script.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/145.0.7632.45..145.0.7632.75?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2026/02/stable-channel-update-for-desktop_13.html)


clickot:
builds on portable linux

PF4Public (approved):

networkException:
How do we want to handle https://chromereleases.googleblog.com/2026/02/stable-channel-update-for-desktop_13.html?

Ahrotahn:
I think these releases are close enough together that the PR could be updated to jump from .45 to .75 if you aren't concerned with having a tag for .67.  There shouldn't be any major changes.  

I'll start work on the new version in a bit once the tarball is available.

networkException:
Alright, I have the clone script running in the background but its been going for 40 Minutes

PR #3657, opened 2026-02-16
add installation source for Artix

salfter:
This is just a quick addition to the top-level README to include Artix's omniverse repo.  (AUR would also work, but omniverse is Artix-specific.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ Other platforms are discussed and tracked in this repository's Issue Tracker. Le
 ungoogled-chromium is available in the following **software repositories**:
 
 * Arch: Available in the AUR, [see instructions in ungoogled-chromium-archlinux](https://github.com/ungoogled-software/ungoogled-chromium-archlinux)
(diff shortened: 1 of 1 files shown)


PF4Public:
Is there a repository with the sourcecode of the package?

salfter:
> Is there a repository with the sourcecode of the package?

A binary package is published at https://omniverse.artixlinux.org/x86_64/.  I'm guessing there's a PKGBUILD that generates it, but I've not been able to locate it.  I'd assume that the PKGBUILD pulls source from here, same as the PKGBUILDs in AUR.

(I didn't create it; I'm just pointing out that a package exists.)

PF4Public:
@Ahrotahn @networkException Would it be acceptable without pointing to source repository?

networkException:
Not a fan honesty

Ahrotahn:
I wasn't able to find the source anywhere, so I'm inclined to agree.  It might just be copies of the arch builds which would already be covered.  Thanks anyways @salfter

PR #3674, opened 2026-02-28
Update to Chromium 146.0.7680.71

Ahrotahn:
This PR updates to Chromium 146.0.7680.71.

Notes:
* Many patches will have small updates for the Value::Dict and List changes in [[7515505]](https://chromium-review.googlesource.com/c/chromium/src/+/7515505) and the base:Contains removal [[7490440]](https://chromium-review.googlesource.com/c/chromium/src/+/7490440)

* The iridium trk patch was updated for [[7534169]](https://chromium-review.googlesource.com/c/chromium/src/+/7534169), [[7276994]](https://chromium-review.googlesource.com/c/chromium/src/+/7276994), [[7486048]](https://chromium-review.googlesource.com/c/chromium/src/+/7486048), and [[7487368]](https://chromium-review.googlesource.com/c/chromium/src/+/7487368)

* The pruned binaries patch had updates for [[7547284]](https://chromium-review.googlesource.com/c/chromium/src/+/7547284), [[7281253]](https://chromium-review.googlesource.com/c/chromium/src/+/7281253), [[7538447]](https://chromium-review.googlesource.com/c/chromium/src/+/7538447), and [[7454131]](https://chromium-review.googlesource.com/c/chromium/src/+/7454131)

* The safebrowsing patch had changes for [[7548924]](https://chromium-review.googlesource.com/c/chromium/src/+/7548924), [[7277325]](https://chromium-review.googlesource.com/c/chromium/src/+/7277325), and [[7510189]](https://chromium-review.googlesource.com/c/chromium/src/+/7510189)

* The unused prefs patch was updated for [[7502380]](https://chromium-review.googlesource.com/c/chromium/src/+/7502380), [[7534884]](https://chromium-review.googlesource.com/c/chromium/src/+/7534884), [[7453067]](https://chromium-review.googlesource.com/c/chromium/src/+/7453067), and [[7456887]](https://chromium-review.googlesource.com/c/chromium/src/+/7456887)

* The close confirmation patch eas updated to work with the changes in [[7533297]](https://chromium-review.googlesource.com/c/chromium/src/+/7533297)

* The tab search button patch was updated for [[7522889]](https://chromium-review.googlesource.com/c/chromium/src/+/7522889), [[7538447]](https://chromium-review.googlesource.com/c/chromium/src/+/7538447)

* The hide extension menu patch was updated for [[7487609]](https://chromium-review.googlesource.com/c/chromium/src/+/7487609)

[...]

PR #3676, opened 2026-03-01
Move Glic/ScreenAI nullification into a separate patch

iskunk:
With the 145.0.7632.116 release, we saw the debut of Glic (Gemini Live in Chrome), and thus patch updates to disable it. Both here and on the Debian side.

I noticed that u-c's Glic edits are rolled into the larger `fix-building-with-prunned-binaries.patch`, while Debian is using [a more targeted patch](https://salsa.debian.org/chromium-team/chromium/-/blob/dc6089a218ec39b51f115d99e711fa8916bf503d/debian/patches/disable/glic.patch) (itself mostly cribbed from u-c's).

Given that Debian needs the Glic edits, but not the other ones in the `fix-building` patch, and that it is preferable to maintain a single instance of the same changes, I'd like to break out those edits into a separate patch. I've also added to it some additional edits by @dilinger, who found a few GN `assert()` calls that also needed zapping to get the build in order.

(Note: I'm not sure if the `OnSearchifyStarted()` edits belong in the new patch; is that related to Glic?)


Ahrotahn:
After they added more glic changes in .116 I figured we would need to split this eventually.  

The searchify stuff is technically part of screen_ai instead of glic.  Does that section you included cause problems for Debian?  Maybe we could make the split for all of their AI changes instead of just glic if the other searchify removals we have don't cause issues for you.

iskunk:
I've asked @dilinger to comment, since he put together the patch, and the searchify bit presumably had to be zapped to get the build working.

A separate screen_ai patch seems reasonable, or do you think a general disable-ai patch is better in the long run?

dilinger:
Oh, I thought the searchify stuff was part of the glic removals; that's why I included it. We're also removing screenai as well, though with just a simple
```
--- a/chrome/browser/component_updater/screen_ai_component_installer.cc
+++ b/chrome/browser/component_updater/screen_ai_component_installer.cc
@@ -143,8 +143,7 @@ void ScreenAIComponentInstallerPolicy::D
 void ManageScreenAIComponentRegistration(ComponentUpdateService* cus,
                                          PrefService* local_state) {
   if (screen_ai::ScreenAIInstallState::ShouldInstall(local_state)) {
-    RegisterScreenAIComponent(cus);
-    return;
+    // don't actually register the service - we don't want the binary blob
   }

   // Clean up.
```

I don't have a strong opinion about separate patches versus one disable-ai patch. Part of it depends on how interwoven upstream's AI stuff will be; if it all ends up depending on each other, then maybe a single disable-ai patch makes sense?

Ahrotahn:
I haven't seen any movement that would imply the two would be consolidating any time soon, so we could go either way.  I guess it depends on what's easier for y'all.  If we only split out glic then you would need to have that one searchify removal included in your platform patches.  If we instead combine all AI removals then we would need to be sure that those extra searchify removals don't cause new problems for your build.

dilinger:
>  If we instead combine all AI removals then we would need to be sure that those extra searchify removals don't cause new problems for your build.

Thinking about it more, I think it makes sense to combine all the AI-removal stuff together.

iskunk:
> Thinking about it more, I think it makes sense to combine all the AI-removal stuff together.

All right, I'll rejigger the patches and update this PR.

iskunk:
All right, screen_ai and a couple things apparently tied to it are now in the new patch, which is renamed to `disable-ai.patch`. I've confirmed that Chromium builds with the new patch minus `prunned-binaries` in Debian, and that both together yield the same as what `prunned-binaries` did before (modulo Andres's edits).

There are still some AI/ML-related edits in `prunned-binaries` that may not may not be desirable to have in the separate patch. The line gets a little blurry between "disable AI thing" and "fix the build because this blob is verboten" :-)  @dilinger, could you let me know if you're happy with the balance between these two patches?

PF4Public:
> is verboten

gotcha! :D

dilinger:
> All right, screen_ai and a couple things apparently tied to it are now in the new patch, which is renamed to `disable-ai.patch`. I've confirmed that Chromium builds with the new patch minus `prunned-binaries` in Debian, and that both together yield the same as what `prunned-binaries` did before (modulo Andres's edits).
> 
> There are still some AI/ML-related edits in `prunned-binaries` that may not may not be desirable to have in the separate patch. The line gets a little blurry between "disable AI thing" and "fix the build because this blob is verboten" :-) @dilinger, could you let me know if you're happy with the balance between these two patches?

I think it generally looks good, though I had two questions:

1) The hunk that removes some privacy-sandbox stuff, shouldn't that be in disable-privacy-sandbox.patch? That seems related to advertising nonsense instead of AI.

2) What is the chrome/browser/accessibility/ removals stuff? Is that related to ReadAnything?  I'm unfamiliar with that, and whether that's AI or just a binary blob or what.

iskunk:
> I think it generally looks good, though I had two questions:
> 
> 1. The hunk that removes some privacy-sandbox stuff, shouldn't that be in disable-privacy-sandbox.patch? That seems related to advertising nonsense instead of AI.

Ah, good catch. Must've accidentally extended my cut-and-paste a bit too far :-]

I see a couple other similar edits in `prunned-binaries`, so I'll move those to `disable-privacy-sandbox.patch` and check that nothing breaks.

> 2. What is the chrome/browser/accessibility/ removals stuff? Is that related to ReadAnything?  I'm unfamiliar with that, and whether that's AI or just a binary blob or what.

I was wondering that too; anyone know what this "tree-fixing" business is about? Seems related to screen_ai.

FWIW, here's a list of other AI/ML-looking things I came across in `prunned-binaries`. Presumably these all require blobs (directly or indirectly), but is there anything among them undesirable enough to be covered in the new patch?

* `permissions_ai_ui_selector`

* `prediction_model_handler_provider`

* `PreloadingModelKeyedService`

* `autofill_ml_internals`

* `read_anything`

* `field_classification_model_handler`

* `optimization_guide::TFLiteOpResolver`

* `use_on_device_model_service`

iskunk:
All right, here's a reshuffling of the reshuffle.

This needs `disable-mei-preload.patch` in order to work, due to nearby edits. @dilinger, I noticed we don't delete `chrome/browser/resources/media/mei_preload/preloaded_data.pb`. Is that worth hanging onto?

The (Debian) patch series section would then look like this:
```
# Ungoogled Chromium
ungoogled/disable-ai.patch
ungoogled/disable-mei-preload.patch
ungoogled/disable-privacy-sandbox.patch
```

dilinger:
> This needs `disable-mei-preload.patch` in order to work, due to nearby edits. @dilinger, I noticed we don't delete `chrome/browser/resources/media/mei_preload/preloaded_data.pb`. Is that worth hanging onto?

We should probably nuke it, along with other protobuf output blobs. If it doesn't break too much, we could delete *.pb in d/copyright.


> 
> The (Debian) patch series section would then look like this:
> 
> ```
> # Ungoogled Chromium
> ungoogled/disable-ai.patch
> ungoogled/disable-mei-preload.patch
> ungoogled/disable-privacy-sandbox.patch
> ```

Looks good to me. If the patches are required to be ordered, feel free to add numbers (1001-disable-mei-preload.patch, 1002-disable-ai.patch, and either give disable-privacy-sandbox.patch a random number of leave it without a number).

iskunk:
> > This needs `disable-mei-preload.patch` in order to work, due to nearby edits. @dilinger, I noticed we don't delete `chrome/browser/resources/media/mei_preload/preloaded_data.pb`. Is that worth hanging onto?
> 
> We should probably nuke it, along with other protobuf output blobs. If it doesn't break too much, we could delete *.pb in d/copyright.

I can give that a try.

> Looks good to me. If the patches are required to be ordered, feel free to add numbers (1001-disable-mei-preload.patch, 1002-disable-ai.patch, and either give disable-privacy-sandbox.patch a random number of leave it without a number).

IMO, numbering the patches gets messy if we need to insert new patches between two existing ones, or before the first one, etc. Our reference here is just the u-c patch series anyway, as we're applying a subset of that.

@Ahrotahn et al., does this work for you? It went a bit beyond the original scope, but I hope this helps bolster the topicality of the patches.

dilinger:
For 146, note https://github.com/ungoogled-software/ungoogled-chromium/pull/3674#issuecomment-4002802041

Ahrotahn:
Yeah there's new changes in 146 that will need to eventually be moved as well.  

Those privacy sandbox changes aren't required to disable the feature, they were only included because of the pruned files.  I'm fine with them being moved since it keeps the related changes consolidated.  

I think it might be good to move the few remaining ML changes over as well, but maybe that would then make the patch more of build-without-models thing rather being AI-focused.  I'm fine with how they are now though if that works for you.  I'll get a build going using the new changes as I get time later today.

Ahrotahn (dismissed):

iskunk:
Should these changes go in as-is? I was going to update the PR for 146, but was having a rough time getting through a test build.

Ahrotahn:
Since I need to update the PR for 146 anyways I figured it would be fine if we get another approval before then.  Now it'll be better to wait until afterwards.

iskunk:
> Since I need to update the PR for 146 anyways I figured it would be fine if we get another approval before then. Now it'll be better to wait until afterwards.

Yeah, as long as that 146 PR is active, the patches are a moving target. I'll rebase and update here once that one's landed.

iskunk:
All right, this is now updated on top of 146.0.7680.75.

Same idea as before, only now this includes some additional edits from @dilinger. If you compare the trees after the patch series is applied, before and after this PR, you'll see 'em. I've included a copy of the diff here for reference:
<details>

```diff
diff -ru chromium-146.0.7680.75-before/chrome/browser/contextual_cueing/contextual_cueing_page_data.cc chromium-146.0.7680.75-after/chrome/browser/contextual_cueing/contextual_cueing_page_data.cc
--- chromium-146.0.7680.75-before/chrome/browser/contextual_cueing/contextual_cueing_page_data.cc	2026-03-14 01:20:55.970964600 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/contextual_cueing/contextual_cueing_page_data.cc	2026-03-14 01:27:02.284691451 -0400
@@ -105,15 +105,6 @@
     auto decision = DidMatchCueingConditions(config);
     if (decision == kAllowed) {
       if (kUseDynamicCues.Get() && config.has_dynamic_cue_label()) {
-        std::move(cueing_decision_callback_)
-            .Run(base::ok(CueingResult{
-                config.dynamic_cue_label(), config.default_text(),
-                /*is_dynamic=*/true, config.auto_open_eligible(),
-                config.has_auto_send_params()
-                    ? std::make_optional(CueingResult::AutoSendParams{
-                          /*auto_send_eligible=*/
-                          config.auto_send_params().auto_send_eligible()})
-                    : std::nullopt}));
         return;
       } else if (config.has_cue_label()) {
         return;
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/BUILD.gn	2026-03-14 01:27:02.285691499 -0400
@@ -8,7 +8,6 @@
 import("//components/guest_view/buildflags/buildflags.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 # TODO(b/430371855): This target should eventually become Glic's public target,
 # to be imported from external BUILD files. The "impl" source set below would
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/browser_ui/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/browser_ui/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-14 01:27:02.286691548 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glow_controller") {
   sources = [
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/fre/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/fre/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/fre/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/fre/BUILD.gn	2026-03-14 01:27:02.286691548 -0400
@@ -5,7 +5,6 @@
 import("//chrome/common/features.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 source_set("fre") {
   sources = [
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/media/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/media/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/media/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/media/BUILD.gn	2026-03-14 01:27:02.286691548 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("media") {
   sources = [
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/resources/glic_resources.cc chromium-146.0.7680.75-after/chrome/browser/glic/resources/glic_resources.cc
--- chromium-146.0.7680.75-before/chrome/browser/glic/resources/glic_resources.cc	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/resources/glic_resources.cc2026-03-14 01:27:02.287691596 -0400
@@ -10,7 +10,7 @@
 namespace glic {
 
 int GetResourceID(int id) {
-  if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2)) {
+  /*if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2))*/ {
     return id;
   }
 
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/shared/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/shared/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/shared/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/shared/BUILD.gn	2026-03-14 01:27:02.287691596 -0400
@@ -1,6 +1,5 @@
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("shared") {
   sources = [
diff -ru chromium-146.0.7680.75-before/chrome/browser/glic/test_support/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/glic/test_support/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/glic/test_support/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/glic/test_support/BUILD.gn	2026-03-14 01:27:02.288691645 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("test_support") {
   testonly = true
diff -ru chromium-146.0.7680.75-before/chrome/browser/resources/glic/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/resources/glic/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/resources/glic/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/resources/glic/BUILD.gn	2026-03-14 01:27:02.294691935 -0400
@@ -7,7 +7,6 @@
 import("//ui/webui/resources/tools/build_webui.gni")
 import("//ui/webui/resources/tools/optimize_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic"
diff -ru chromium-146.0.7680.75-before/chrome/browser/resources/glic/fre/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/resources/glic/fre/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-14 01:27:02.294691935 -0400
@@ -4,7 +4,6 @@
 
 import("//ui/webui/resources/tools/build_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic_fre"
diff -ru chromium-146.0.7680.75-before/chrome/browser/skills/skills_ui_tab_controller.cc chromium-146.0.7680.75-after/chrome/browser/skills/skills_ui_tab_controller.cc
--- chromium-146.0.7680.75-before/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-14 01:27:02.295691983 -0400
@@ -28,8 +28,6 @@
 DEFINE_USER_DATA(skills::SkillsUiTabController);
 
 namespace {
-using glic::mojom::SkillSource;
-
 constexpr base::TimeDelta kNotifyTimeoutSeconds = base::Seconds(60);
 constexpr base::TimeDelta kGlicPanelPollIntervalMilliseconds =
     base::Milliseconds(60);
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/views/glic/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/ui/views/glic/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/ui/views/glic/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/views/glic/BUILD.gn	2026-03-14 01:27:02.299692177 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/views/tabs/glic/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/ui/views/tabs/glic/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-14 01:27:02.300692225 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = [ "glic_actor_constants.h" ]
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-14 01:20:55.987965423 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-14 01:27:02.301692274 -0400
@@ -163,7 +163,6 @@
     "//chrome/browser/ui/omnibox",
     "//chrome/browser/ui/profiles",
     "//chrome/browser/ui/safety_hub",
-    "//chrome/browser/ui/tabs:glic",
     "//chrome/browser/ui/tabs:tab_enums",
     "//chrome/browser/ui/tabs:tab_strip",
     "//chrome/browser/ui/tabs:tab_strip_model_observer",
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn
--- chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn2026-03-11 18:12:25.000000000 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn	2026-03-14 01:27:02.302692322 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/toolbar_view.cc chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/toolbar_view.cc
--- chromium-146.0.7680.75-before/chrome/browser/ui/views/toolbar/toolbar_view.cc	2026-03-14 01:21:17.881024884 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/views/toolbar/toolbar_view.cc2026-03-14 01:27:03.996774303 -0400
@@ -679,10 +679,8 @@
   glic_button_->SetGlicPanelIsOpen(open);
 }
 #else
-void ToolbarView::SetGlicShowState(bool show) {
-}
-void ToolbarView::SetGlicPanelIsOpen(bool open) {
-}
+void ToolbarView::SetGlicShowState(bool show) { }
+void ToolbarView::SetGlicPanelIsOpen(bool open) { }
 #endif  // ENABLE_GLIC
 
 void ToolbarView::AnimationEnded(const gfx::Animation* animation) {
diff -ru chromium-146.0.7680.75-before/chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc chromium-146.0.7680.75-after/chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc
--- chromium-146.0.7680.75-before/chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc	2026-03-14 01:20:55.991965616 -0400
+++ chromium-146.0.7680.75-after/chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc	2026-03-14 01:27:02.306692516 -0400
@@ -141,6 +141,14 @@
     case MojoTabAlertState::kActorWaitingOnUser:
       *out = NativeTabAlertState::kActorWaitingOnUser;
       return true;
+#if BUILDFLAG(ENABLE_GLIC)
+    case MojoTabAlertState::kGlicAccessing:
+      *out = NativeTabAlertState::kGlicAccessing;
+      return true;
+    case MojoTabAlertState::kGlicSharing:
+      *out = NativeTabAlertState::kGlicSharing;
+      return true;
+#endif
   }
   NOTREACHED();
 }

```
</details>

As before, I've confirmed that both the Debian and (converted) u-c builds work with this.

I did notice the following warnings in the u-c build; any potential issues here?
```
chrome/browser/ui/tabs/tab_strip_api/converters/tab_converters.cc:177:11: warning: enumeration values 'kGlicAccessing' and 'kGlicSharing' not handled in switch [-Wswitch]
chrome/browser/ui/tabs/tab_strip_api/converters/tab_converters.cc:211:1: warning: non-void function does not return a value in all control paths [-Wreturn-type]
chrome/browser/ui/webui/browser_command/browser_command_handler.cc:98:11: warning: enumeration value 'kOpenSafeBrowsingEnhancedProtectionSettings' not handled in switch [-Wswitch]
chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc:95:11: warning: enumeration values 'kGlicAccessing' and 'kGlicSharing' not handled in switch [-Wswitch]
```

dilinger:
Some of my changes were a result of deleting things and then having to add them back in when the build failed. `chrome/browser/ui/views/toolbar/toolbar_view.cc` is an example of that; my change can be dropped there. The change to `chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc` might be another example of that, I'm not sure why I would've re-added stuff that is #ifdef'd out with the glic build flag.

The `assert(enable_glic)` are definitely needed, as well as a few other things. When I prepare .80 (for debian) later today, I can go through and verify which patch hunks aren't actually needed.

iskunk:
For `chrome/browser/ui/views/toolbar/toolbar_view.cc`, u-c already had a similar edit in `prunned-binaries`; I just went with your formatting when I moved the hunk over :-)

The `chrome/browser/ui/webui/tabs/tabs_mojom_traits.cc` change did have me scratching my head. I don't think we're too worried about things working when `ENABLE_GLIC` is true, so I'll remove that one, and rebase to .80.

Let me know what else can be removed, and I'll test it out and update here.

iskunk:
Oh! I see what happened in `tabs_mojom_traits.cc`. The u-c change deleted all those lines, and yours just added the `BUILDFLAG` bits. The hunks conflicted, and yours went in.

That brings up a different issue I noticed: Some of the edits to these files have looong sections of contiguous line removals, like the last hunk in `chrome/browser/navigation_predictor/navigation_predictor.cc` (starting [here](https://github.com/ungoogled-software/ungoogled-chromium/pull/3676/changes#diff-9a6fee148af1c24f43031d0a74816f0443c2a1e8ee894ffe98a3582e537c711bR540)). Edits like those are going to be fragile over time, since any change to a line of the function will cause the hunk to break. I was wondering if there's some kind of modus operandi at work here, since while on the one hand you have the fragility, on the other hand, you might want to require manual review when a line changes (since it could have implications for other changes that need to be made).

My normal preference is the `BUILDFLAG` (or just `#if 0`) approach, and when the start and end are within the same hunk, there's little practical difference. But I'm curious to know if the long-term experience here has shown it to not be adequate for bigger edits on a codebase as active as Chromium's.

PF4Public:
If I'm not mistaken, there is no single-correct-answer to that. On one hand disabling the code using pragmas you make the change resilient against upstream changes, on the other hand you make the context of the patch harder to infer, which might complicate adapting it to new changes.

dilinger:
Oh, sorry @iskunk , I misunderstood what you were saying.. I see that I'm not re-adding the buildflag stuff (in tabs_mojom_traits.cc22), I'm just not bothering to remove it because it is already protected with #ifdefs. However, upstream seems to be in the process of removing those ifdefs, so it's probably a good idea to include deletion of that in disable-ai.patch. That way when the patch is updated to 147 or 148, the conflict will be obvious.

FYI, I successfully built without the contextual_cueing_page_data.cc hunk, so you can probably drop that piece. Note that I"m only talking about the portion that I removed; the portion that was already removed in the u-c patch should probably stay in the disable-ai patch

EDIT: sorry, I thought I clicked the "comment"  button on this one hours ago before I left the house, but apparently I didn't!

dilinger:
https://chromium.googlesource.com/chromium/src/+/879529ed148860941944db3c6e04dbf0133008de%5E%21/   removes those BUILDFLAG ifdefs. We could simply revert that, but for now it's probably easier to delete the whole piece of code inside of the existing BUILDFLAG ifdefs.

For places where there's no glic-related ifdefs and we're adding our own (`#if 0` or whatever), that's personally my preference for long-term maintenance.

iskunk:
> If I'm not mistaken, there is no single-correct-answer to that. On one hand disabling the code using pragmas you make the change resilient against upstream changes, on the other hand you make the context of the patch harder to infer, which might complicate adapting it to new changes.

If it's a matter of making the patch easier to understand, there's also the way of adding comments to it---either at the top, between the files (which the diff format supports, not sure if that also goes for between individual hunks), or adding notes to the patched files. That doesn't seem too favored here, however, probably since a lot of the patches come out of the scramble of new major releases.

Sometimes, I think about the possibility of having a better tool (or tools) to manages these changes, since `quilt(1)` is a bit long in the tooth. I'm a big fan of [Meld](https://meldmerge.org/), and something in that vein would be a big improvement. (Is there anything out there along these lines, that perhaps I'm not aware of?)

> However, upstream seems to be in the process of removing those ifdefs, so it's probably a good idea to include deletion of that in disable-ai.patch

Huhh, I see. Yeah, Google would say, "Why would we want to disable Glic, when it's always going to be on?" :-]

I'll replace that patch hunk to delete all the lines, and the one you mentioned in `contextual_cueing_page_data.cc`, and get a test build going.

iskunk:
Okay, updated as discussed, rebased to 146.0.7680.80, and test builds check out. Here is the new before/after diff with the full patch series applied:
<details>

```diff
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/BUILD.gn	2026-03-15 09:20:23.214085961 +0000
@@ -8,7 +8,6 @@
 import("//components/guest_view/buildflags/buildflags.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 # TODO(b/430371855): This target should eventually become Glic's public target,
 # to be imported from external BUILD files. The "impl" source set below would
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/browser_ui/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/browser_ui/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-15 09:20:23.214085961 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glow_controller") {
   sources = [
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/fre/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/fre/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/fre/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/fre/BUILD.gn	2026-03-15 09:20:23.214085961 +0000
@@ -5,7 +5,6 @@
 import("//chrome/common/features.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 source_set("fre") {
   sources = [
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/media/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/media/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/media/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/media/BUILD.gn	2026-03-15 09:20:23.214085961 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("media") {
   sources = [
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/resources/glic_resources.cc chromium-146.0.7680.80-after/chrome/browser/glic/resources/glic_resources.cc
--- chromium-146.0.7680.80-before/chrome/browser/glic/resources/glic_resources.cc	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/resources/glic_resources.cc2026-03-15 09:20:23.214085961 +0000
@@ -10,7 +10,7 @@
 namespace glic {
 
 int GetResourceID(int id) {
-  if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2)) {
+  /*if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2))*/ {
     return id;
   }
 
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/shared/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/shared/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/shared/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/shared/BUILD.gn	2026-03-15 09:20:23.214085961 +0000
@@ -1,6 +1,5 @@
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("shared") {
   sources = [
diff -ru chromium-146.0.7680.80-before/chrome/browser/glic/test_support/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/glic/test_support/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/glic/test_support/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/glic/test_support/BUILD.gn	2026-03-15 09:20:23.215086006 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("test_support") {
   testonly = true
diff -ru chromium-146.0.7680.80-before/chrome/browser/resources/glic/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/resources/glic/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/resources/glic/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/resources/glic/BUILD.gn	2026-03-15 09:20:23.217086096 +0000
@@ -7,7 +7,6 @@
 import("//ui/webui/resources/tools/build_webui.gni")
 import("//ui/webui/resources/tools/optimize_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic"
diff -ru chromium-146.0.7680.80-before/chrome/browser/resources/glic/fre/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/resources/glic/fre/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-15 09:20:23.217086096 +0000
@@ -4,7 +4,6 @@
 
 import("//ui/webui/resources/tools/build_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic_fre"
diff -ru chromium-146.0.7680.80-before/chrome/browser/skills/skills_ui_tab_controller.cc chromium-146.0.7680.80-after/chrome/browser/skills/skills_ui_tab_controller.cc
--- chromium-146.0.7680.80-before/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-15 09:20:23.217086096 +0000
@@ -28,8 +28,6 @@
 DEFINE_USER_DATA(skills::SkillsUiTabController);
 
 namespace {
-using glic::mojom::SkillSource;
-
 constexpr base::TimeDelta kNotifyTimeoutSeconds = base::Seconds(60);
 constexpr base::TimeDelta kGlicPanelPollIntervalMilliseconds =
     base::Milliseconds(60);
diff -ru chromium-146.0.7680.80-before/chrome/browser/ui/views/glic/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/ui/views/glic/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/ui/views/glic/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/ui/views/glic/BUILD.gn	2026-03-15 09:20:23.218086142 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.80-before/chrome/browser/ui/views/tabs/glic/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/ui/views/tabs/glic/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-15 09:20:23.218086142 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = [ "glic_actor_constants.h" ]
diff -ru chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-15 09:20:06.085310723 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-15 09:20:23.219086187 +0000
@@ -163,7 +163,6 @@
     "//chrome/browser/ui/omnibox",
     "//chrome/browser/ui/profiles",
     "//chrome/browser/ui/safety_hub",
-    "//chrome/browser/ui/tabs:glic",
     "//chrome/browser/ui/tabs:tab_enums",
     "//chrome/browser/ui/tabs:tab_strip",
     "//chrome/browser/ui/tabs:tab_strip_model_observer",
diff -ru chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn
--- chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn2026-03-13 16:54:03.000000000 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn	2026-03-15 09:20:23.219086187 +0000
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/toolbar_view.cc chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/toolbar_view.cc
--- chromium-146.0.7680.80-before/chrome/browser/ui/views/toolbar/toolbar_view.cc	2026-03-15 09:20:06.768341637 +0000
+++ chromium-146.0.7680.80-after/chrome/browser/ui/views/toolbar/toolbar_view.cc2026-03-15 09:20:24.034123075 +0000
@@ -679,10 +679,8 @@
   glic_button_->SetGlicPanelIsOpen(open);
 }
 #else
-void ToolbarView::SetGlicShowState(bool show) {
-}
-void ToolbarView::SetGlicPanelIsOpen(bool open) {
-}
+void ToolbarView::SetGlicShowState(bool show) { }
+void ToolbarView::SetGlicPanelIsOpen(bool open) { }
 #endif  // ENABLE_GLIC
 
 void ToolbarView::AnimationEnded(const gfx::Animation* animation) {
```
</details>

Let me know if anything else needs addressing. Hoping we can get this in before Google finds _another_ CVE...

dot-gov:
I tried building with the new `disable-ai.patch‎` applied but without the `fix-building-with-prunned-binaries.patch‎`, and got this error:
```
In file included from ../../chrome/browser/skills/skills_update_observer.cc:5:
../..\chrome/browser/skills/skills_update_observer.h(13,10): fatal error: 'chrome/browser/glic/host/glic.mojom.h' file not found
   13 | #include "chrome/browser/glic/host/glic.mojom.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Adding this hunk from `fix-building-with-prunned-binaries.patch‎` fixed the issue:
```diff
--- a/chrome/browser/skills/BUILD.gn
+++ b/chrome/browser/skills/BUILD.gn
@@ -50,11 +50,7 @@ source_set("public") {
 source_set("ui") {
   sources = [
     "skills_dialog_launcher.cc",
-    "skills_ui_tab_controller.cc",
-    "skills_ui_tab_controller.h",
     "skills_ui_window_controller.cc",
-    "skills_update_observer.cc",
-    "skills_update_observer.h",
   ]

   public_deps = [
```

I'm using a custom patchset, which might explain why it works for you, at the same time is there a reason why that the above hunk isn't in `disable-ai.patch‎`? It seems like there's skills related removals in that as well, and it's unclear why some skills related removals are in `disable-ai.patch‎` vs `fix-building-with-prunned-binaries.patch‎`.

iskunk:
Good catch, @dot-gov. I've moved that hunk into `disable-ai.patch`. Debian test build still checks out.

> I'm using a custom patchset, which might explain why it works for you, at the same time is there a reason why that the above hunk isn't in `disable-ai.patch‎`? It seems like there's skills related removals in that as well, and it's unclear why some skills related removals are in `disable-ai.patch‎` vs `fix-building-with-prunned-binaries.patch‎`.

It's not always clear what the connection is between the features we're intentionally removing (glic, screen_ai) and some of these adjacent ones that break without them. Already, `disable-ai.patch` has pulled in several additional items (e.g. tree-fixing) to keep the build working. The situation with skills appears to be another case of this, though I wasn't privy to the breakage that spurred the original edit---that was one of @dilinger's. The Debian builds never ran into the breakage you saw, at any rate.

(It's funny how after pulling out everything AI-related, there's not a whole lot left in the `prunned-binaries` patch...)

iskunk:
Are there any further concerns with this PR?

Note that the u-c build for Debian needs the additional `assert()` removals. When I drop Debian's `disable/glic.patch` in favor of u-c's `prunned-binaries` (on current master), I get
```
*********************************************************************
WARNING: Support for linking against a C++ standard library other 
  than the one in-tree (buildtools/third_party/libc++) is deprecated
  and support for this will end. We plan to remove this option in 
  M138.
*********************************************************************
ERROR at //chrome/browser/glic/BUILD.gn:11:1: Assertion failed.
assert(enable_glic)
^-----
See //chrome/browser/BUILD.gn:8572:5: which caused the file to be included.
    "//chrome/browser/glic:mojo_bindings",
    ^------------------------------------
make[1]: *** [debian/rules:244: override_dh_auto_build-arch] Error 1
make[1]: Leaving directory '/home/build/work/ungoogled-chromium-146.0.7680.80'
make: *** [debian/rules:203: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2
```
(And if I drop `prunned-binaries` for `disable/glic.patch`, other u-c patches break.)

The original motivation for this change is to get the two projects playing nice with each other again. It's not possible as long as there are two independently-maintained patches targeting the same broad swath of functionality.

Ahrotahn:
I'm a bit behind, but I just looked over the changes and everything seems ok so far.  I'll get a build going with the changes sometime today.  If it doesn't cause problems on your end it might be good to have the last few autofill and prediction/permissions AI sections moved over as well.  

> I was wondering if there's some kind of modus operandi at work here

Yes, I've found gates often cause maintenance problems over time.  One of the worst examples I can remember was when a new function was added that happened to also match the section for a closing gate which caused an important override to be included in the removal.  It built and ran but there were issues during runtime that I spent days debugging to figure out the cause.  Now when I'm updating the patches I'll change those to full removals if the surrounding code needs to be updated anyways.  The full removals also make reviewing a patch easier since you can see what's actually being done without having to also have the source open.

iskunk:
> I'm a bit behind, but I just looked over the changes and everything seems ok so far. I'll get a build going with the changes sometime today.

Thanks, my main concern is that once the next release drops, that your work builds on this, rather than this getting left behind.

> If it doesn't cause problems on your end it might be good to have the last few autofill and prediction/permissions AI sections moved over as well.

I moved those over and tested the Debian build of .80, and in fact I got one compile failure:
```
FAILED: obj/components/autofill/core/browser/browser/autofill_manager.o
[...]
../../components/autofill/core/browser/foundations/autofill_manager.cc:836:21: error: no member named 'autofill_predictions' in 'autofill::AutofillManager::AsyncContext'
  836 |             context.autofill_predictions = std::move(model_predictions);
      |             ~~~~~~~ ^
../../components/autofill/core/browser/foundations/autofill_manager.cc:839:21: error: no member named 'password_manager_predictions' in 'autofill::AutofillManager::AsyncContext'
  839 |             context.password_manager_predictions = std::move(model_predictions);
      |             ~~~~~~~ ^
```
Fixed it with this new edit:
```diff
--- a/components/autofill/core/browser/foundations/autofill_manager.cc
+++ b/components/autofill/core/browser/foundations/autofill_manager.cc
@@ -833,10 +833,8 @@ void AutofillManager::RunMlModels(
                                    }));
         switch (source) {
           case HeuristicSource::kAutofillMachineLearning:
-            context.autofill_predictions = std::move(model_predictions);
             break;
           case HeuristicSource::kPasswordManagerMachineLearning:
-            context.password_manager_predictions = std::move(model_predictions);
             break;
           case HeuristicSource::kRegexes:
             NOTREACHED();
```
which I added to `disable-ai.patch`, and have pushed here. The overall before/after diff likewise gains the above.

> Yes, I've found gates often cause maintenance problems over time. One of the worst examples I can remember was when a new function was added that happened to also match the section for a closing gate which caused an important override to be included in the removal. It built and ran but there were issues during runtime that I spent days debugging to figure out the cause. Now when I'm updating the patches I'll change those to full removals if the surrounding code needs to be updated anyways.

I was thinking something like that was possible in theory, but wasn't sure it could actually happen in practice without breaking the build (as random-ish edits typically do). That's a good cautionary tale.

> The full removals also make reviewing a patch easier since you can see what's actually being done without having to also have the source open.

The question then is if e.g. upstream changes one line in a function that's fully removed, how long does it take to figure out that's what's going on and update the removal. At least you're familiar with the source/patches, and can identify (some) cases of this on sight; for someone who's only passingly familiar with either, the tools don't provide a whole lot of help, what with those big `*.rej` files...

dilinger:
Was this meant to be merged?  I see it got approved, but not merged..

Ahrotahn:
Two approvals are required to merge.  
Unfortunately my earlier approval is now stale since there were new changes added in .153 that this would need to be updated for.

iskunk:
I've updated this PR to .153. The combined before/after diff remains as follows:
<details>

```diff
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/BUILD.gn	2026-03-20 02:21:20.441570910 -0400
@@ -8,7 +8,6 @@
 import("//components/guest_view/buildflags/buildflags.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 # TODO(b/430371855): This target should eventually become Glic's public target,
 # to be imported from external BUILD files. The "impl" source set below would
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/browser_ui/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/browser_ui/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/browser_ui/BUILD.gn	2026-03-20 02:21:20.441570910 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glow_controller") {
   sources = [
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/fre/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/fre/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/fre/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/fre/BUILD.gn	2026-03-20 02:21:20.442570958 -0400
@@ -5,7 +5,6 @@
 import("//chrome/common/features.gni")
 import("//mojo/public/tools/bindings/mojom.gni")
 
-assert(enable_glic)
 
 source_set("fre") {
   sources = [
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/media/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/media/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/media/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/media/BUILD.gn	2026-03-20 02:21:20.442570958 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("media") {
   sources = [
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/resources/glic_resources.cc chromium-146.0.7680.153-after/chrome/browser/glic/resources/glic_resources.cc
--- chromium-146.0.7680.153-before/chrome/browser/glic/resources/glic_resources.cc	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/resources/glic_resources.cc	2026-03-20 02:21:20.443571006 -0400
@@ -10,7 +10,7 @@
 namespace glic {
 
 int GetResourceID(int id) {
-  if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2)) {
+  /*if (!base::FeatureList::IsEnabled(features::kGlicAssetsV2))*/ {
     return id;
   }
 
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/shared/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/shared/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/shared/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/shared/BUILD.gn	2026-03-20 02:21:20.443571006 -0400
@@ -1,6 +1,5 @@
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("shared") {
   sources = [
diff -ru chromium-146.0.7680.153-before/chrome/browser/glic/test_support/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/glic/test_support/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/glic/test_support/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/glic/test_support/BUILD.gn	2026-03-20 02:21:20.444571054 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("test_support") {
   testonly = true
diff -ru chromium-146.0.7680.153-before/chrome/browser/resources/glic/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/resources/glic/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/resources/glic/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/resources/glic/BUILD.gn	2026-03-20 02:21:20.451571392 -0400
@@ -7,7 +7,6 @@
 import("//ui/webui/resources/tools/build_webui.gni")
 import("//ui/webui/resources/tools/optimize_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic"
diff -ru chromium-146.0.7680.153-before/chrome/browser/resources/glic/fre/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/resources/glic/fre/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/resources/glic/fre/BUILD.gn	2026-03-20 02:21:20.452571440 -0400
@@ -4,7 +4,6 @@
 
 import("//ui/webui/resources/tools/build_webui.gni")
 
-assert(enable_glic)
 
 build_webui("build") {
   grd_prefix = "glic_fre"
diff -ru chromium-146.0.7680.153-before/chrome/browser/skills/skills_ui_tab_controller.cc chromium-146.0.7680.153-after/chrome/browser/skills/skills_ui_tab_controller.cc
--- chromium-146.0.7680.153-before/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/skills/skills_ui_tab_controller.cc	2026-03-20 02:21:20.453571488 -0400
@@ -28,8 +28,6 @@
 DEFINE_USER_DATA(skills::SkillsUiTabController);
 
 namespace {
-using glic::mojom::SkillSource;
-
 constexpr base::TimeDelta kNotifyTimeoutSeconds = base::Seconds(60);
 constexpr base::TimeDelta kGlicPanelPollIntervalMilliseconds =
     base::Milliseconds(60);
diff -ru chromium-146.0.7680.153-before/chrome/browser/ui/views/glic/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/ui/views/glic/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/ui/views/glic/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/ui/views/glic/BUILD.gn	2026-03-20 02:21:20.457571681 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.153-before/chrome/browser/ui/views/tabs/glic/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/ui/views/tabs/glic/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/ui/views/tabs/glic/BUILD.gn	2026-03-20 02:21:20.458571729 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = [ "glic_actor_constants.h" ]
diff -ru chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-20 02:28:06.758164907 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/BUILD.gn	2026-03-20 02:21:20.459571778 -0400
@@ -163,7 +163,6 @@
     "//chrome/browser/ui/omnibox",
     "//chrome/browser/ui/profiles",
     "//chrome/browser/ui/safety_hub",
-    "//chrome/browser/ui/tabs:glic",
     "//chrome/browser/ui/tabs:tab_enums",
     "//chrome/browser/ui/tabs:tab_strip",
     "//chrome/browser/ui/tabs:tab_strip_model_observer",
diff -ru chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn
--- chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/glic/BUILD.gn	2026-03-17 15:56:22.000000000 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/glic/BUILD.gn	2026-03-20 02:21:20.459571778 -0400
@@ -4,7 +4,6 @@
 
 import("//chrome/common/features.gni")
 
-assert(enable_glic)
 
 source_set("glic") {
   sources = []
diff -ru chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/toolbar_view.cc chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/toolbar_view.cc
--- chromium-146.0.7680.153-before/chrome/browser/ui/views/toolbar/toolbar_view.cc	2026-03-20 02:28:16.669642912 -0400
+++ chromium-146.0.7680.153-after/chrome/browser/ui/views/toolbar/toolbar_view.cc	2026-03-20 02:21:22.177654618 -0400
@@ -679,10 +679,8 @@
   glic_button_->SetGlicPanelIsOpen(open);
 }
 #else
-void ToolbarView::SetGlicShowState(bool show) {
-}
-void ToolbarView::SetGlicPanelIsOpen(bool open) {
-}
+void ToolbarView::SetGlicShowState(bool show) { }
+void ToolbarView::SetGlicPanelIsOpen(bool open) { }
 #endif  // ENABLE_GLIC
 
 void ToolbarView::AnimationEnded(const gfx::Animation* animation) {
diff -ru chromium-146.0.7680.153-before/components/autofill/core/browser/foundations/autofill_manager.cc chromium-146.0.7680.153-after/components/autofill/core/browser/foundations/autofill_manager.cc
--- chromium-146.0.7680.153-before/components/autofill/core/browser/foundations/autofill_manager.cc	2026-03-20 02:28:06.770165485 -0400
+++ chromium-146.0.7680.153-after/components/autofill/core/browser/foundations/autofill_manager.cc	2026-03-20 02:21:20.472572404 -0400
@@ -833,10 +833,8 @@
                                    }));
         switch (source) {
           case HeuristicSource::kAutofillMachineLearning:
-            context.autofill_predictions = std::move(model_predictions);
             break;
           case HeuristicSource::kPasswordManagerMachineLearning:
-            context.password_manager_predictions = std::move(model_predictions);
             break;
           case HeuristicSource::kRegexes:
             NOTREACHED();
```
</details>

@Ahrotahn, @networkException, please (re-)review and let me know if there is any issue preventing a merge.

iskunk:
Thank you @Ahrotahn, @PF4Public :-)

I'll get `disable-ai.patch` and its necessary accompaniments into the Debian build, and update the u-c Debian conversion to work with the new patch arrangement.

PR #3697, opened 2026-03-15
Bundle hyphenation patterns

### LINKED ISSUE #3135: 'hyphens' CSS property isn't working
### OS/Platform

Windows

### Installed

https://ungoogled-software.github.io/ungoogled-chromium-binaries/

### Version

Version 131.0.6778.204 (Official Build, ungoogled-chromium) (64-bit)

### Description

Hyphenation rules aren't being rendered

### How to Reproduce?

1. Apply hyphenation in CSS (e.g. hyphens:auto, or check out [a test page](https://clagnut.com/sandbox/hyphenation.html)). 
2. (Observe no changes)

### Actual behaviour

The property has no effect on the rendered text. In ungoogled chromium:
image

### Expected behaviour

Hyphen rules should be applied. In fresh/latest Chrome install (131.0.6778.205) :
image

### Relevant log output

_No response_

### Additional context

I haven't gone through toggling each flag yet, but here's the command I have under chrome://settings:
`"\chrome.exe" --restart --restart --flag-switches-begin --disable-search-engine-collection --disable-sharing-hub --enable-experimental-web-platform-features --enable-gpu-rasterization --extension-mime-request-handling=always-prompt-for-install --force-dark-mode --scroll-tabs=always --enable-features=FluentOverlayScrollbar,OverlayScrollbar,TabGroupsSaveUIUpdate,WebRtcHideLocalIpsWithMdns,WindowsScrollingPersonality --disable-features=ChromeLabs,GlobalMediaControlsCastStartStop --flag-switches-end` 

I'm almost certain hyphenation was working at some point in the past year--assuming my settings aren't the culprit, I think the issue would be somewhere in the blink>renderer>core>layout>inline code

implicitfield:
Fixes #3135 and makes the `headless_shell` target buildable without further changing any gn flags.

This excludes `*.hyb` files from pruning and adds a patch to generate the `hypen-data` directory on platforms that need it.

<details>
<summary>Old description</summary>

`third_party/hyphenation-patterns/mk_hyb_file.py` has been taken from [here](https://android.googlesource.com/platform/frameworks/minikin/+/refs/heads/master/tools/mk_hyb_file.py), which is also where the in-tree `build-hyb.sh` script fetches it from.

This patch will probably cause conflicts if/when this component gets updated/tweaked, but that has only really happened twice since the end of 2020.

The `.hyb` filess are fully reproducible at the moment, but IMO it's nicer to generate these from source anyway (but I'm open to changing any aspect of this PR as well).
</details>


PF4Public:
Awesome! Thanks.

Are extra steps needed? Like copying the files into installation dir, or are they already contained with other resources as a result?

implicitfield:
The `bundle_hyphen_data` dependency takes care of installing these (and registering a component), so no extra steps are needed :)

wukko:
> `third_party/hyphenation-patterns/mk_hyb_file.py` has been taken from [here](https://android.googlesource.com/platform/frameworks/minikin/+/refs/heads/master/tools/mk_hyb_file.py), which is also where the in-tree `build-hyb.sh` script fetches it from.

if i understand correctly, you're patching in the `mk_hyb_file.py` file to avoid downloading extra remote dependencies? otherwise, breaking domain substitution in `third_party/hyphenation-patterns/build-hyb.sh` and running it could do the same thing without having to redo `BUILD.gn`.

> The .hyb filess are fully reproducible at the moment, 

> This patch will probably cause conflicts if/when this component gets updated/tweaked, 

since `.hyb` files are fully reproducible, and this would cause extra maintenance tax, i feel it makes more sense to just exclude `third_party/hyphenation-patterns/hyb/` from pruning.

implicitfield:
> if i understand correctly, you're patching in the mk_hyb_file.py file to avoid downloading extra remote dependencies? otherwise, breaking domain substitution in third_party/hyphenation-patterns/build-hyb.sh and running it could do the same thing without having to redo BUILD.gn.

Yeah, just rerunning that without patching anything would work too if you didn't apply domain substitution. It's just that remote dependencies are generally frowned upon and that there isn't really a way to properly integrate that script into the build process (it won't work on windows, won't have proper dependency tracking without redoing BUILD.gn anyway, etc.)

> since .hyb files are fully reproducible, and this would cause extra maintenance tax, i feel it makes more sense to just exclude third_party/hyphenation-patterns/hyb/ from pruning.

You could do that too, though these really have historically been updated extremely rarely so the maintenance tax seems relatively low. I expected that files excluded from pruning were mostly things that have a standardized format, but that doesn't really apply seeing that there's already a few `.bin` files in there.

---

@PF4Public: It occurred to me that you'll actually need to install the `hyphen-data` directory from the build output directory alongside the `chrome` executable if you're using a custom install script, but that should be everything.

networkException:
I'll look into nixpkgs changes in a bit :)

PF4Public:
> It's just that remote dependencies are generally frowned upon

I'm with you on this one. It won't build on Gentoo as well without compromising the sandbox.



> It occurred to me that you'll actually need to install the `hyphen-data` directory from the build output directory alongside the `chrome` executable if you're using a custom install script, but that should be everything.

Yep, I was suspecting something like that.

Ahrotahn:
> since .hyb files are fully reproducible, and this would cause extra maintenance tax, i feel it makes more sense to just exclude third_party/hyphenation-patterns/hyb/ from pruning.

I agree with this.  Files that are either reproducible or match the hash of a third-party open-source project's build output are good candidates for exclusions when they are available in both the tarball and clone.  All that would need to be done is add '*.hyb' to the `PRUNING_EXCLUDE_PATTERNS` section in update_lists.py and update the pruning list.  Then we wouldn't need to keep track of any updates the script may need in the future.

implicitfield:
Alright, do we still want to patch in the `bundle_hyphen_data` target as a dependency for the `chrome` target, or should we let platform builds scripts decide if they want to build that target?

Ahrotahn:
I think we should keep it.  That way the only thing platforms would need to do is include the `hyphen-data` directory in their package if they want that capability.

implicitfield:
I've now reworked this (and updated the PR description accordingly) to no longer prune and instead bundle the in-tree `.hyb` files.

Ahrotahn (dismissed):
Thanks @implicitfield!

PF4Public (dismissed):

Ahrotahn:
@implicitfield when you get the chance would you please rebase this over the latest changes?

implicitfield:
> when you get the chance would you please rebase this over the latest changes?

I've now rebased this.

Ahrotahn (approved):

PF4Public (approved):

PR #3729, opened 2026-04-09
Fix search bar on new tab page for downstream disable-ai.patch users

dilinger:
When debian switched to using u-c's disable-ai.patch, we found that the offline new tab page was broken. Instead, users were sent to an online search engine, which is not desirable. It turns out that the part of this patch that removed ReadAnythingAppController was also removing unrelated (to AI) SearchBoxExtension code. This puts the code back.

More details can be seen @ https://bugs.debian.org/1132651

Note that u-c users don't see this due to the new tab page being overridden by extra/inox-patchset/0008-restore-classic-ntp.patch. However, at some point if that other patch were removed, u-c would've hit the same bug.

### DIFF
--- a/patches/core/ungoogled-chromium/disable-ai.patch
+++ b/patches/core/ungoogled-chromium/disable-ai.patch
@@ -1069,15 +1069,10 @@
  #include "chrome/renderer/searchbox/searchbox_extension.h"
  #endif  // !BUILDFLAG(IS_ANDROID)
  
-@@ -336,20 +335,6 @@ void ChromeRenderFrameObserver::DidCommi
- }
+@@ -341,14 +340,6 @@ void ChromeRenderFrameObserver::DidClear
+     SearchBoxExtension::Install(render_frame()->GetWebFrame());
+   }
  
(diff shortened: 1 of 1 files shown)


PF4Public (approved):
Thanks!

Ahrotahn (approved):
Thanks @dilinger!

Ivelisse09:
> When debian switched to using u-c's disable-ai.patch, we found that the offline new tab page was broken. Instead, users were sent to an online search engine, which is not desirable. It turns out that the part of this patch that removed ReadAnythingAppController was also removing unrelated (to AI) SearchBoxExtension code. This puts the code back.
> 
> More details can be seen @ https://bugs.debian.org/1132651
> 
> Note that u-c users don't see this due to the new tab page being overridden by extra/inox-patchset/0008-restore-classic-ntp.patch. However, at some point if that other patch were removed, u-c would've hit the same bug.
> 
> _(Please ensure you have read SUPPORT.md and docs/contributing.md before submitting the Pull Request)_

I want you to tesch me everiting because my life Is you'r computersnd yiu know i know who you peole áre

PR #3738, opened 2026-04-15
Update to Chromium 147.0.7727.101

Ahrotahn:
This PR updates to Chromium 147.0.7727.101
There were no notable changes needed on our end.  

<br>

[Change log](https://chromium.googlesource.com/chromium/src/+log/147.0.7727.55..147.0.7727.101?pretty=fuller&n=10000)
[Blog post](https://chromereleases.googleblog.com/2026/04/stable-channel-update-for-desktop_15.html)

PR #3744, opened 2026-04-17
Pin depot_tools to commit 5e98835 in utils/clone.py

hambingen:
Chromium 147.0.7727.101 broke `utils/clone.py` due to a mismatch in the `DEPS` file ([more info](https://github.com/ungoogled-software/ungoogled-chromium/pull/3738)). However some projects like [ungoogled-chromium-macos](https://github.com/ungoogled-software/ungoogled-chromium-macos/blob/a949291cb2fc05f7c30a2606b21b7e5b04383a90/retrieve_and_unpack_resource.sh#L46) depend on this functionality.

Fix the script via an ugly hack that pins depot_tools to a specific [working commit](https://chromium.googlesource.com/chromium/tools/depot_tools.git/+/5e9883513113f4c6b74e33d80c066075d0fd3960).

### DIFF
--- a/utils/clone.py
+++ b/utils/clone.py
@@ -88,8 +88,8 @@ def clone(args): # pylint: disable=too-many-branches, too-many-locals, too-many-
 
     get_logger().info('Cloning depot_tools')
     # To-Do: Undo this conditional once upstream rolls depot_tools
-    # More info: https://github.com/ungoogled-software/ungoogled-chromium/pull/3738#issuecomment-4256065678
-    if (chromium_version == '147.0.7727.101'):
+    # https://github.com/ungoogled-software/ungoogled-chromium/pull/3738#issuecomment-4256065678
+    if chromium_version == '147.0.7727.101':
         dt_commit = '5e9883513113f4c6b74e33d80c066075d0fd3960'
     else:
         dt_commit = re.search(r"depot_tools\.git'\s*\+\s*'@'\s*\+\s*'([^']+)',",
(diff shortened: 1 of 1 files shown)


Ahrotahn:
I seem to recall that there was a platform that required the use of the commit listed in the DEPS file, this might cause a problem for their build.  I guess we would need to check if using this commit works for everyone.  

The DEPS file will be updated for 148 but I don't know how things will be handled for the refreshes that will happen between now and the 5th.  Hopefully they will have it updated before then but if not then we would then need to update the version for this change each time.  

Maybe a better way to handle this would be to add a new argument to the script that allows passing a commit to be used as an override?

Qubik65536:
I believe I am in need of this (or some fix) to release `147.0.7727.116`, which also fixes the glic issue that is preventing a lot of people to update all together.

I would simply remove the problematic section from DEPS as described in https://github.com/ungoogled-software/ungoogled-chromium/pull/3738 (until further fix is in place) if that sounds good?

Qubik65536:
I ended up using tarball for 147.0.7727.116 :/

Ahrotahn:
@teeminus Have you tried building with deopt_tools at HEAD somewhat recently?  I went through some old PRs and it seems like the only reason we check out at the DEPS commit was for a windows build issue.  Maybe things have changed since then and we could remove the commit checkout altogether.

Nifury:
@Ahrotahn Yes it builds fine with depot_tools at HEAD.

Ahrotahn:
Thanks for testing that @Nifury, and sorry we're so slow with this @hambingen & @Qubik65536.  

@hambingen would you be willing to update the PR to uproot the dtcommit section?  Otherwise we can close this and I can remove it in the next update.

PF4Public:
What is the status of this PR? Are we closing it or should it be reworked?

Ahrotahn:
I meant to close it after 149, sorry about that!

PR #3775, opened 2026-05-09
Add patch for disabling JIT

dot-gov:
This PR adds a flag to disable JIT using the existing "JavaScript optimization" setting. If the flag is enabled, turning off JavaScript optimization also turns off JIT. Normally turning off JIT disables webasm as well, which breaks some sites, but Microsoft's drumbrake allows it to still work even with JIT turned off. That's not enabled by default, so that also requires adding the following flags to flags.gn:
```
v8_enable_drumbrake = true
v8_drumbrake_bounds_checks = true
```
~I wanted to keep the changes small, so I left the flags commented out by default. The code always adds the `--wasm_jitless` flag regardless of whether drumbrake is enabled, but it doesn't seem to cause ill effects even if drumbrake is disabled, so I didn't adding a guard for it.~

Other Chromium based browsers have this feature, though with slightly nicer UI:

https://github.com/brave/brave-core/pull/30703/files

https://github.com/GrapheneOS/Vanadium/

To test, go to https://ffmpegwasm.netlify.app/playground/
1. If it's stuck on loading, then webasm is disabled. You can also check for a console message for "ReferenceError: WebAssembly is not defined"
2. To test whether drumbrake is enabled, try encoding a video file. Click "load sample files", then paste the following into the arguments box: `["-i", "video.webm", "-c:v", "libx264", "video.mp4"]`. With JIT enabled the file encodes in around 7s, but with JIT disabled and with drumbrake, it takes over 2 minutes.

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -25,6 +25,7 @@ If a switch requires a value, you must specify it with an `=` sign; e.g. flag `-
   `--max-connections-per-host` | (from Bromite) Configure the maximum allowed connections per host. Valid values are `6` and `15`
   `--omnibox-autocomplete-filtering` | Restrict omnibox autocomplete results to a combination of search suggestions (if enabled), bookmarks, and internal chrome pages. Accepts `search`, `search-bookmarks`, `search-chrome`, and `search-bookmarks-chrome`.
   `--popups-to-tabs` | Makes popups open in new tabs.
+  `--disabling-javascript-optimization-disables-jit` | Disables JIT too if the "JavaScript optimization" setting is turned off
 
 - ### Available only on desktop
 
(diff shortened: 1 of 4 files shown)


dot-gov:
It seems like `validate_config` doesn't play nice with the commented lines I added, so let me know whether we should add drumbrake related flags (uncommented) to the file, or remove them altogether.

Ahrotahn:
Works as expected in my build.  

It would be better to uncomment the gn flags and add them to the list like the others.  The extra build time and binary size was negligible for me.  

Also don't forget to update docs/flags.md with the new flag.

dot-gov:
Updated PR with your requested changes

Ahrotahn (approved):
Thanks @dot-gov!

PF4Public (approved):
Thanks!

Issue #2933, opened 2024-06-26
Move away from Cirrus CI


networkException:
We did so for platform repos already, lets move to GitHub Actions here as well.

PR #3776, opened 2026-05-11
Port CirrusCI jobs to github actions

dot-gov:
In another PR I noticed the cirrusCI checks had a warning that [it was shutting down on June 1](https://cirruslabs.org/), so I quickly ported over the current CirrusCI jobs to github actions. There's [a past issue](https://github.com/ungoogled-software/ungoogled-chromium/issues/2933) for this, but it looks like it hasn't been touched in forever.

Implementation notes:
1. I ended up keeping the current CirrusCI files, and tried to reuse them where possible. Once CirrusCI is shut down we can remove them. This also means both will run side by side, which should be fine and helps weed out any edge cases
2. Using custom containers is a pain with github actions, so I resorted to having it install dependencies on every job run, which seems to be the recommended approach
3. I ported over the caching logic to use github caches, so it doesn't redownload the tarballs for each run


Ahrotahn:
Good catch!  I generally only follow through to cirrus when an action gets hung up, and I wouldn't have even noticed the extra banner among the others.  

It might be better to use
```yaml
with:
  python-version: 'pypy3.10' 
```
for the python setup instead of using the `.python-version` file.  It would clear up the extra warnings in the logs.  

I'm surprised that there was enough space available for the chromium source.  That was an ongoing issue in the past as the size slowly increased.  Apparently the ubuntu-latest image has ~25GB of free space available by default and I think we'll probably run into space issues the next time a clone is required.  
There is a nice blog post [here](https://carlosbecker.com/posts/github-actions-disk-space/) that details a simple way to free up a lot more space.  I think we should either incorporate a similar step or the linked action to be sure we have enough room available.  

Also I think it would be a good idea to make a separate cache key for the pip step like we currently do for cirrus.  Any time savings for that are welcome since the setup is run for each step.

dot-gov:
> It might be better to use
> 
> ```yaml
> with:
>   python-version: 'pypy3.10' 
> ```
> 
> for the python setup instead of using the `.python-version` file. It would clear up the extra warnings in the logs.

I added a `python-version-file` parameter to silence the warning. I want to keep the .python-version version file because it provides a recommended python version for other tooling that support it, eg. uv or pipx. Also, any reason to use pypy in particular? It's not actively developed so we'll probably have to revert back to cpython at some point, and none of the script looks heavy enough that would benefit from the better performance.

> Apparently the ubuntu-latest image has ~25GB of free space available by default

This is probably fine because they seemed to have increased the disk size at some point. I added a `df -h` command at the end of a job (ie. after the source has been extracted) and got:

```
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       145G   66G   79G  46% /
tmpfs           7.9G   84K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
efivarfs        128M   32K  128M   1% /sys/firmware/efi/efivars
/dev/sda16      881M   64M  756M   8% /boot
/dev/sda15      105M  6.2M   99M   6% /boot/efi
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```

I also did a fresh shallow clone (including deps) and the disk usage is ~30 GB, so we should be good here.

> I think we'll probably run into space issues the next time a clone is required.

I think the clone logic was broken for a while because our scripts are running with python 3.10, but the scripts in the depot_tools repo require at least 3.11 because of a StrEnum import. Upgrading the python version also opens up a can of worms because a bunch of dependencies have to be upgraded, and one of them is the formatter, which requires the code to be reformatted. I can include those changes here too if you want, but I didn't want this PR to include a bunch of unrelated changes.

> Also I think it would be a good idea to make a separate cache key for the pip step like we currently do for cirrus. Any time savings for that are welcome since the setup is run for each step.

Actually I don't think caching was enabled for pip step at all, so I pushed a commit that adds it. The total python setup step takes like 10s, though so any savings are minimal.

Ahrotahn:
`python-version-file` works too!  

I wasn't aware they increased the storage space, so that's good news.  
I also wasn't aware about the python update, I guess it's been a while since we've run a job that needed to clone.  

Most of our scripts are long overdue for being updated.  I've done some small changes to work with newer versions of python but the knock-on effects you mentioned are one of the reasons I haven't done more.  I'd be interested in your changes, though it would be easier to work with those updates in a separate PR.  

Historically we kept to older versions of tooling in order to maintain compatibility with the latest stable Debian, but there's no other reason it can't be updated to use newer, maintained, tooling.

Ahrotahn (approved):
Thanks @dot-gov this looks good to me!

PF4Public (approved):
Thanks!

Ahrotahn:
After merging this I noticed that there were two new action failures from dependabot:  [1](https://github.com/ungoogled-software/ungoogled-chromium/actions/runs/26362288383), [2](https://github.com/ungoogled-software/ungoogled-chromium/actions/runs/26362288463)
I didn't realize that we had dependabot enabled on the repo.  I guess the new actions trigger that now.  

I feel like it would just be more noise than help and would be best to turn it off.  There's also a Copilot Autofix option I'd like to disable as well if y'all are ok with that.  
@networkException have you found dependabot to be useful on the arch repo?  Maybe it's worth fixing?

networkException:
I've been ignoring dependabot on the arch repo for the most time because updating npm packages for CI wasn't really a priority

Issue #3698, opened 2026-03-17
Custom Flags Don't Work With Vertical Tabs


avenged110:
### OS/Platform

macOS

### Installed

Downloaded from the [macOS build's Releases page](https://github.com/ungoogled-software/ungoogled-chromium-macos/releases).

### Version(s)

146.0.7680.80-1.1 – 147.0.7727.116-1.1 (arm64)

### Have you tested that this is not an upstream issue or an issue with your configuration?

This does not affect the upstream _Chromium_ project because the flags are specific to the _Ungoogled Chromium_ project. Also, the issue is present regardless of configuration, fresh or otherwise.

### Description

_Ungoogled Chromium_-specific flags for disabling tab hover cards, tab close buttons, and the tab search button don't work when tabs are vertical. There could be other flags similarly affected which I didn't happen to notice.

### How to Reproduce?

1. In chrome://flags, enable "#hide-tab-close-buttons," "#vertical-tabs," and "#tab-hover-cards" and relaunch.
2. Navigate to chrome://settings/appearance and set "Tab strip position" to vertical.
3. Open a tab.
4. Notice that the tab close button is present and that a hover card still appears when hovering over-top of the tab.

### Actual behavior

The tab close buttons and tab hover cards are not disabled like the flags claim when vertical ("side") tabs are enabled. This does not affect "top" (traditional) tabs as the flags work as-expected and as they always have in that mode.

### Expected behavior

The _Ungoogled Chromium_-specific flags for disabling these UI elements should disable them regardless of the tab strip's position.

tomasz1986:
I agree that it would be very nice to have these flags work with vertical tabs. For myself, especially with no way to hide tab close buttons, if you shrink the vertical tab column to its minimum size, the current tab only shows the close button and nothing else, so there is no favicon visible, and also, there is no way to know if the tab is loading or refreshing (technically there is, by looking at the refresh icon in the address bar, but it's quite cumbersome 🙁), because even then only the close button is shown with no spinning circle around it.

avenged110:
I'm not particularly familiar with the workings of Github, but I noticed someone mentioned the 'hover cards' and 'close button' patches were updated during the initial update to version 147.x, but just wanted to note that these flags are still broken in practice when using vertical tabs under the as-of-yet latest macOS build of version 147.0.7727.116-1.1.

Qubik65536:
[...]

PR #3782, opened 2026-05-15
Check hide tab close buttons on vertical tabs

khamer:
this is a fix for part of #3698 - fixing vertical tabs to respect the hide tab close buttons flag. I've tested the patch application against Chromium 148.0.7778.167 and I've gotten a downstream fork (Helium) to compile using this patch as well and reviewed the code. I didn't get all the way through a local compile of ungoogled-chromium itself as I believe I'm just stuck on getting the toolchain setup properly, but nevertheless, this is a small patch and so the best of my ability I've tested it.

The motivation for this fix is the difficulty in click (especially right-clicking) on tabs while in vertical mode and tabs are shrunk down to the smallest size. I consider this a straight forward bug fix - just making the existing flag correctly affect vertical tabs the way it affects horizontal ones.

Just noting, I did use AI to find and develop this fix, but I have reviewed all code involved and believe I'm matching existing code conventions and the fix is implemented consistent to how this flag is checked elsewhere.


Ahrotahn:
Works as expected in my build.  The only nit I have is that the change should be added to the existing patch instead of being in a new one.

khamer:
Merged into the existing patch.

PF4Public (approved):
Thanks!

Ahrotahn (approved):
Thanks @khamer

PR #3810, opened 2026-06-05
extra/disable-jit-flag: build drumbrake only if supported

### LINKED ISSUE #3792: v8_enable_drumbrake=true in flags.gn causes GN assertion failure when building for x86
### OS/Platform

Windows

### Installed

Compiled from source

### Version

148.0.7778.178 (build failure)

### Description

v8_enable_drumbrake=true in flags.gn causes GN assertion failure when building for x86

### How to Reproduce?

1. Clone ungoogled-chromium-windows
2. Build for x86 (target_cpu="x86")
3. Run build.py
4. [GN generation fails at v8/BUILD.gn:590](https://github.com/ungoogled-software/ungoogled-chromium-windows/actions/runs/26300651359/job/77424631509#step:4:3339)

### Actual behaviour

GN exits with assertion error: "DrumBrake is only available on x64, arm64 on Windows, Linux and MacOS."

### Expected behaviour

GN generation completes successfully; x86 build proceeds normally.

### Relevant log output

```sh
C:\ungoogled-chromium-windows\build\src>"out\Default\gn.exe" "gen" "out\Default" "--fail-on-unused-args"
ERROR at //v8/BUILD.gn:590:3: Assertion failed.
  assert(
  ^-----
DrumBrake is only available on x64, arm64 on Windows, Linux and MacOS.
See //v8/BUILD.gn:591:7: 
      is_drumbrake_supported,
      ^---------------------
This is where it was set.
See //BUILD.gn:223:9: which caused the file to be included.
        "//v8:gn_all",
        ^------------

C:\ungoogled-chromium-windows\build\src>exit
  File "C:\ungoogled-chromium-windows\build.py", line 330, in <module>
    main()
  File "C:\ungoogled-chromium-windows\build.py", line 300, in main
    _run_build_process('out\\Default\\gn.exe', 'gen', 'out\\Default', '--fail-on-unused-args')
  File "C:\ungoogled-chromium-windows\build.py", line 62, in _run_build_process
    subprocess.run(('cmd.exe', '/k'),
  File "C:\hostedtoolcache\windows\Python\3.12.10\x64\Lib\subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('cmd.exe', '/k')' returned non-zero exit status 1.
```

### Additional context

In V8's build system, `is_drumbrake_supported` requires all of the following ([source](https://source.chromium.org/chromium/_/chromium/v8/v8/+/ad6e4525c418a92147c8247ef9d144ce4c242a38:gni/v8.gni;l=348)):

```
is_drumbrake_supported =
    v8_enable_webassembly && v8_enable_pointer_compression &&
    (v8_current_cpu == "x64" || v8_current_cpu == "arm64" || v8_current_cpu == "riscv64") &&
    (target_os == "win" || target_os == "linux" || target_os == "mac" || target_os == "ios")
```

x86 builds fail both conditions: `v8_current_cpu` is not in the supported list, and `v8_enable_pointer_compression` is disabled by default on x86.

The assertion is triggered whenever `v8_enable_drumbrake && v8_enable_webassembly && !is_drumbrake_supported` ([source](https://source.chromium.org/chromium/_/chromium/v8/v8/+/ad6e4525c418a92147c8247ef9d144ce4c242a38:BUILD.gn;l=589)):

```
if (v8_enable_drumbrake && v8_enable_webassembly) {
  assert(
      is_drumbrake_supported,
      "DrumBrake is only available on x64, arm64 on Windows, Linux and MacOS.")
}
```

dumbmoron:
Moves the `v8_enable_drumbrake` arg below `is_drumbrake_supported`, so that we can use it as the deciding factor on whether it should be built.

Resolves #3792

### DIFF
--- a/flags.gn
+++ b/flags.gn
@@ -16,4 +16,3 @@ treat_warnings_as_errors=false
 use_official_google_api_keys=false
 use_unofficial_version_number=false
 v8_drumbrake_bounds_checks=true
-v8_enable_drumbrake=true
--- a/patches/extra/ungoogled-chromium/add-flag-for-disabling-jit.patch
+++ b/patches/extra/ungoogled-chromium/add-flag-for-disabling-jit.patch
@@ -9,6 +9,30 @@
 +     "If enabled, disables JIT too if the \"JavaScript optimization\" setting is turned off. ungoogled-chromium flag.",
 +     kOsAll, SINGLE_VALUE_TYPE("disabling-javascript-optimization-disables-jit")},
  #endif  // CHROME_BROWSER_UNGOOGLED_FLAG_ENTRIES_H_
+--- a/v8/gni/v8.gni
++++ b/v8/gni/v8.gni
+@@ -102,9 +102,6 @@ declare_args() {
+   # Sets -DV8_LITE_MODE.
+   v8_enable_lite_mode = false
+ 
+-  # WebAssembly interpreter (DrumBrake)  build flag.
+-  v8_enable_drumbrake = false
+-
+   # Enable Wasm interpreter tracing.
+   v8_enable_drumbrake_tracing = false
+ 
+@@ -351,6 +348,11 @@ is_drumbrake_supported =
+     (target_os == "win" || target_os == "linux" || target_os == "mac" ||
+      target_os == "ios")
+ 
++declare_args() {
++  # WebAssembly interpreter (DrumBrake)  build flag.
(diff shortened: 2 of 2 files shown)


PF4Public (approved):
Thanks!

Ahrotahn (approved):
Thanks @dumbmoron

PR #3842, opened 2026-06-26
Fix building with ASAN+UBSAN

implicitfield:
These partially defined classes cause errors such as `undefined symbol: typeinfo for local_discovery::ServiceDiscoverySharedClient` when linking in an `is_asan=true` + `is_ubsan=true` build. I believe that's due to the implementations for these not getting compiled at all, not just because of the forward-declarations.

The constructor that I deleted for `DnsSdRegistry` seems to only be used in tests so this builds fine without it.


PF4Public (approved):

Ahrotahn (approved):
Thanks @implicitfield!

PR #3865, opened 2026-07-12
optimize disable_ai_patch,remove unnecessary code.

### LINKED ISSUE #3861: I don't think we need this code.(platform win/mac)
### OS/Platform

Win/Mac OS

### Installed

Compiled from source

### Version

main

### Description

I don't think we need this code.

### How to Reproduce?
```diff
--- a/chrome/browser/actor/ui/actor_ui_tab_controller.cc
+++ b/chrome/browser/actor/ui/actor_ui_tab_controller.cc
@@ -50,7 +50,6 @@ ActorUiTabController::ActorUiTabControll
       scoped_unowned_user_data_(tab.GetUnownedUserDataHost(), *this) {
   CHECK(base::FeatureList::IsEnabled(features::kGlicActorUi));
   CHECK(base::FeatureList::IsEnabled(features::kGlicActor));
-  CHECK(actor_keyed_service_);
 }
```
I don't think we need this code.

### Actual behaviour

I don't think we need this code.

### Expected behaviour

I don't think we need this code.

### Relevant log output

```sh

```

### Additional context

_No response_

colayu:
This value is now always guaranteed to be present, so it can never be null. We can remove this check. Otherwise, keeping it here is a bit inconsistent, since the other usages of this value in this file don't perform any null checks. It makes the code look a little odd.

### DIFF
--- a/patches/core/ungoogled-chromium/disable-ai.patch
+++ b/patches/core/ungoogled-chromium/disable-ai.patch
@@ -135,17 +135,7 @@
 -
    return nullptr;
  }
- 
---- a/chrome/browser/actor/ui/actor_ui_tab_controller.cc
-+++ b/chrome/browser/actor/ui/actor_ui_tab_controller.cc
-@@ -50,7 +50,6 @@ ActorUiTabController::ActorUiTabControll
-       scoped_unowned_user_data_(tab.GetUnownedUserDataHost(), *this) {
-   CHECK(base::FeatureList::IsEnabled(features::kGlicActorUi));
-   CHECK(base::FeatureList::IsEnabled(features::kGlicActor));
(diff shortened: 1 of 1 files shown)


Ahrotahn (approved):
This change can be removed since all relevant calls to ActorUiTabController have been removed.  

The value for actor_keyed_service_ is **_not_** guaranteed to be present and can absolutely be null.  It is initialized as nullptr in the header and is set from outside during construction, so the lack of null checks is poor form.  

Also the space character shouldn't have been removed, but it seems to not cause any problems and will be re-added by quilt in the next refresh.

PF4Public (approved):
Thanks @colayu @Ahrotahn !

PR #3871, opened 2026-07-15
devutils/update_lists: use fork multiprocessing only on posix

wukko:
fixes breakage on windows introduced by #3866.

fork is indeed available on all POSIX systems, but windows isn't one, so we use the default spawn method on all non-POSIX systems (nt/windows) instead.

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -404,5 +404,6 @@ def main(args_list=None):
 
 
 if __name__ == "__main__":
-    multiprocessing.set_start_method('fork')
+    if os.name == 'posix':
+        multiprocessing.set_start_method('fork')
     main()


networkException (approved):

Ahrotahn (approved):
Thanks @wukko!

iskunk:
Ah, I missed that the "fork" method is not available on Windows.

Note that this means the same original failure mode (where `--domain-exclude-prefix` silently becomes a no-op) would occur there. I'd suggest sticking in e.g. `assert os.name == 'posix', 'This option only works on POSIX systems'` to save someone else the trouble of debugging it.

Making this work consistently across platforms would be a bit more involved. I tried looking for a way of passing along the parent process's global state, but could not find the right incantation.

PR #3888, opened 2026-07-28
Disable GCM: fail fast PushManager.subscribe() without a token

### LINKED ISSUE #3770: PushManager#subscribe() never resolves nor rejects
### OS/Platform

macOS

### Installed

OS/Platform's package manager

### Version

Version 147.0.7727.116 (Official Build, ungoogled-chromium) (arm64)

### Description

PushManager#subscribe() never resolves nor rejects

### How to Reproduce?

On a secure origin:
1. Grant notification permission to the site
2. Register a service worker (can be minimal)
3. Attempt to subscribe to a push service

For example:
```js
await Notification.requestPermission(); // then allow notifications
const serviceWorkerRegistration = await navigator.serviceWorker.register("./service-worker.js");
await serviceWorkerRegistration.pushManager.subscribe({applicationServerKey: "", userVisibleOnly: true}); // with a valid value for applicationServerKey
```

### Actual behaviour

The promise returned by PushManager#subscribe never resolves, causing unexpected behaviour in applications that depend on the result.

### Expected behaviour

The promise at least rejects, to allow applications to gracefully handle the lack of Web Push support without workarounds.

### Relevant log output

```sh

```

### Additional context

This problem has been referenced as a comment on a previous issue: https://github.com/ungoogled-software/ungoogled-chromium/issues/1482#issuecomment-1865167127

ShortTimeNoSee:
disable-gcm makes `GCMClientImpl::Start()` a no-op but `GCMDriverDesktop::EnsureStarted()` still returns `SUCCESS` and posts start work that never completes, which leaves `GetToken`/`PushManager.subscribe()` hanging

This extends `patches/core/ungoogled-chromium/disable-gcm.patch` so `EnsureStarted()` returns `GCM_DISABLED`. Matches Chromium's disabled-GCM mapping and makes subscribe reject with `AbortError` ("Registration failed - push service error") as opposed to hanging. Local `Notification` is fine. This of course doesn't add an alternative push backend.

Same approach as my Helium PR at https://github.com/imputnet/helium/pull/2221. Helium maintainers asked to land it here.

Fixes #3770


devutils/tests/test_disable_gcm_fail_fast_push.py
Ahrotahn, line 1: Please remove this.  We already have tooling to test if patches apply cleanly.

patches/core/ungoogled-chromium/disable-gcm.patch
```diff
@@ -154,3 +154,36 @@
...
  
  std::string GCMClientImpl::GetStateString() const {
+--- a/components/gcm_driver/gcm_driver_desktop.cc
++++ b/components/gcm_driver/gcm_driver_desktop.cc
+@@ -1189,28 +1189,10 @@ void GCMDriverDesktop::SetAccountTokens(
+ }
+ 
+ GCMClient::Result GCMDriverDesktop::EnsureStarted(
+-    GCMClient::StartMode start_mode) {
++    GCMClient::StartMode /*start_mode*/) {
```
Ahrotahn: This is unnecessary.

Ahrotahn (changes requested):

patches/core/ungoogled-chromium/disable-gcm.patch
ShortTimeNoSee: It was just to silence unused param warnings after the body stopped using it. I'll remove it 👍

Ahrotahn (dismissed):

PF4Public (dismissed):
Thanks @ShortTimeNoSee, @Ahrotahn

Ahrotahn:
When you get the chance would you mind rebasing this over the latest changes?

Ahrotahn (approved):

PF4Public (approved):

PR #3914, opened 2026-08-15
Preserve Google WebAuthn AppID validation

### LINKED ISSUE #3913: Domain substitution breaks Google WebAuthn AppID validation
Domain substitution currently rewrites content/browser/webauth/webauth_request_security_checker_impl.cc and .h. Those files contain the upstream Chromium Google AppID compatibility check:

- caller_origin.DomainIs("google.com")
- https://www.gstatic.com/securitykey/origins.json
- https://www.gstatic.com/securitykey/a/google.com/origins.json

After substitution, an explicit appidExclude from accounts.google.com is rejected with SecurityError: This is an invalid domain. The rejection occurs before an attached chrome.webAuthenticationProxy extension receives onCreateRequest, so an authenticator extension cannot handle it.

This is observable in current ungoogled-based builds: a google.com create request with googleLegacyAppidSupport=true reaches the proxy, while the equivalent current Google Account flow with explicit appidExclude using the upstream gstatic URL fails before the proxy.

These strings do not initiate a Google service request. They are local validation constants for a compatibility rule shared by Chromium and Firefox. Excluding the security checker implementation and header from domain substitution preserves that validation without restoring background Google integration.

I have a small change ready that removes those two source files from domain_substitution.list.

Ehekatl:
Fixes #3913.

Domain substitution rewrites the Google-domain check and both gstatic AppID constants in Chromiums WebAuthn security checker. That makes the built-in Google compatibility rule impossible to match, so current Google passkey creation fails with SecurityError before an attached WebAuthn proxy can receive the request.

Exclude only the checker implementation and header from substitution. Their Google strings are local validation constants, not service endpoints used for background requests.

Validation:
- reproduced explicit appidExclude rejection on an accounts.google.com page in an ungoogled-based Chromium 146 build
- verified the equivalent googleLegacyAppidSupport request reaches chrome.webAuthenticationProxy
- git diff --check
- directly executed utils.tests.test_domain_substitution.test_update_timestamp with PYTHONPATH=.:utils

A full Chromium build is intentionally left to project CI.

### DIFF
--- a/devutils/update_lists.py
+++ b/devutils/update_lists.py
@@ -109,6 +109,11 @@ PRUNING_EXCLUDE_PATTERNS = [
 # Paths to exclude by prefixes of the POSIX representation for domain substitution
 DOMAIN_EXCLUDE_PREFIXES = [
     'components/test/',
+    # Preserve local validation and test vectors for legacy Google WebAuthn AppIDs.
+    # The checker compares these AppID URLs locally and does not fetch them.
+    'content/browser/webauth/authenticator_impl_unittest.cc',
+    'content/browser/webauth/webauth_request_security_checker_impl.cc',
+    'content/browser/webauth/webauth_request_security_checker_impl.h',
     'net/http/transport_security_state_static.json',
     'net/http/transport_security_state_static_pins.json',
     # Exclusions for Visual Studio Project generation with GN (PR #445)
--- a/domain_substitution.list
+++ b/domain_substitution.list
@@ -4515,11 +4515,8 @@ content/browser/web_contents/web_contents_view_aura_unittest.cc
 content/browser/web_contents/web_contents_view_mac_unittest.mm
 content/browser/web_contents/web_drag_dest_mac_unittest.mm
 content/browser/webauth/authenticator_common_impl.cc
-content/browser/webauth/authenticator_impl_unittest.cc
 content/browser/webauth/authenticator_mojom_traits_unittest.cc
 content/browser/webauth/authenticator_test_base.h
-content/browser/webauth/webauth_request_security_checker_impl.cc
-content/browser/webauth/webauth_request_security_checker_impl.h
 content/browser/webauth/webauth_request_security_checker_impl_unittest.cc
 content/browser/webid/delegation/jwt_signer_unittest.cc
 content/browser/webid/idp_network_request_manager_unittest.cc


dumbmoron:
this list is autogenerated by a script, you probably need to update that and then regenerate it

PF4Public:
> Exclude only the checker implementation and header from substitution. Their Google strings are local validation constants, not service endpoints used for background requests.

I assume this change targets the following code:

```c
  // As a compatibility hack, sites within google.com are allowed to assert two
  // special-case AppIDs. Firefox also does this:
  // https://groups.google.com/forum/#!msg/mozilla.dev.platform/Uiu3fwnA2xw/201ynAiPAQAJ
  const GURL gstatic_appid(kGstaticAppId);
  const GURL gstatic_corp_appid(kGstaticCorpAppId);
  DCHECK(gstatic_appid.is_valid() && gstatic_corp_appid.is_valid());
  if (caller_origin.DomainIs("google.com") && !appid_url.has_ref() &&
      (appid_url.EqualsIgnoringRef(gstatic_appid) ||
       appid_url.EqualsIgnoringRef(gstatic_corp_appid))) {
    *out_app_id = appid;
    return blink::mojom::AuthenticatorStatus::SUCCESS;
  }
```

Are we really willing to allow this exclusion?

> Validation:
> reproduced explicit appidExclude rejection on an accounts.google.com page in an ungoogled-based Chromium 146 build
> verified the equivalent googleLegacyAppidSupport request reaches chrome.webAuthenticationProxy
> git diff --check
> directly executed utils.tests.test_domain_substitution.test_update_timestamp with PYTHONPATH=.:utils
> A full Chromium build is intentionally left to project CI.

So how is it supposed to be verified?

Ehekatl:
@PF4Public @dumbmoron  Yes, that is the code. I think the exception is safe because Chromium only compares those AppID URLs locally; it never fetches them.

I moved the exclusions to DOMAIN_EXCLUDE_PREFIXES and regenerated the list against Chromium 151. I also included the existing AppID test file so its inputs still match the checker. After applying domain substitution, the checker and its tests keep the real values while nearby WebAuthn files are still substituted. Lint and utility tests pass; I have not done a full Chromium build.

PF4Public:
Sorry, but I'm still not convinced that opening an exception for google aligns with the project goals.

wukko:
imo it makes no sense for ***ungoogled*** chromium to make an exception for google's broken auth. google should update their webauthn implementation (and generally follow standards, maybe?) instead of forcing web browsers to make an exception for them.

Ehekatl:
I agree it's google own problem, very annoying though, thanks for the review

PR #3922, opened 2026-08-20
Add flag to force reduced motion or no reduced motion

### LINKED ISSUE #3786: Add a new flag that allows users to override system prefers-reduced-motion setting
### Description

Add a new flag that allows users to override system prefers-reduced-motion setting

### Who's implementing?

### The problem

If Animations are disabled in Windows settings or Reduce Motion is enabled in macOS settings, there is no way to keep animations in Chromium.

### Possible solutions

Add a new flag that allows users to override system prefers-reduced-motion setting, like Firefox's "ui.prefersReducedMotion".

### Alternatives

_No response_

### Additional context

These command line flags exist:
--force-prefers-reduced-motion
--force-prefers-no-reduced-motion
But they will not work when right click - open with ...

seNqejnicru5:
Add flag "Force reduced motion" using existing switches, like Firefox's "ui.prefersReducedMotion". It has three options: Default, Force reduced motion, Force no reduced motion. Closes #3786

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -90,6 +90,7 @@ Chromium contains switches that do no have corresponding entries in `chrome://fl
   `--disable-webgl` | Disable all versions of WebGL.
   `--enable-low-end-device-mode` | Force low-end device mode when set.
   `--force-dark-mode` | Forces dark mode in UI for platforms that support it.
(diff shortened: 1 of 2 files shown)


Ahrotahn (approved):

PF4Public (approved):
Thanks!

PR #3947, opened 2026-09-08
devutils/check_downloads_ini: fix crash on default invocation

dodo-md:
## Summary
- `check_downloads_ini.py`'s `--downloads-ini` argparse option uses `type=Path`, but argparse does not apply `type` conversion to the `default` value — so running the script with no arguments passed a `str` instead of a `Path` into `DownloadInfo`, causing a crash.
- Fix by constructing `default_downloads_ini` as a `Path` directly instead of `str(...)`.

## Test plan
- [x] Ran `devutils/check_downloads_ini.py` with no arguments and confirmed it no longer crashes
- [x] Ran `devutils/check_downloads_ini.py -d <path>` and confirmed explicit paths still work


Ahrotahn (approved):

PF4Public (approved):
Thanks!

